PR #4, opened 2015-06-26
Tracing abstraction

spencergibb:
The abstraction is based heavily off of http://htrace.incubator.apache.org.  

The terminology is based on google [dapper](http://research.google.com/pubs/pub36356.html).

* __Trace__: The overarching unit of sleuth.  A tree of spans.  Has an id.
* __Span__: Nodes in a trace tree.  Edges represent parent/child relationship.  Spans without a parent are root spans.  Has a name, id, trace id, parent span ids, start and end times.
* __Annotations__: Data associated with a span
* __Sampler__: Determines when to trace

### API

Trace is in interface with DefaultTrace is an implementation.  It's job is to start Spans appropriately.

* `TraceScope startSpan(String name)`
* `TraceScope startSpan(String name, Span parent)`
* `TraceScope startSpan(String name, Sampler<T> s)`

TraceScope is a class that defines behavior for the life of a Span.

* `public TraceScope(ApplicationEventPublisher publisher, Span span, Span savedSpan)`
* `public Span detach()`
* `public void close()`

You'll notice the `ApplicationEventPublisher` in the constructor.  Currently two spring application events are published: SpanStartedEvent and SpanStoppedEvent, which are self explanatory.  If have implemented listeners that for setting the trace id and span id via Slf4j MDC so those values can be shown in logs.  This replicates the 4financeit correlationid that was printed in the logs.  This is also where publishing spans to various pipelines can be implemented: Zipkin, cloudfoundry, jdbc etc...  Because it is event driven, multiple entities can consume spans.

Sampler interface.  Allows customization of when to trace.  A default Sampler is an argument to `DefaultTrace`.

### Sampler

* `boolean next(T info)`

### Instrumentation

Instrumenting code is the next important step in tracing.  There is instrumentation for hystrix, spring scheduling, spring mvc controllers, servlet filter, `RestTemplate` and Zuul.  There are two major ways of instrumenting code, via AOP and via lifecycle hooks (spring mvc, RestTemplate).

@mstine @dsyer @joshlong

@marcingrzejszczak I'd like you to review and ask any questions you might have about how this is different from the implementation you committed.  I believe everything you had is still there.  What you called `Correlation-Id` is now analogous to `Trace-Id`, you just wouldn't care for the `Span-Id`'s if you didn't want to.  It's pluggable now.


marcingrzejszczak:
I'm on it - 70 files changes... dis gonna be good :P

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/TraceScope.java
```diff
@@ -0,0 +1,86 @@
+package org.springframework.cloud.sleuth;
+
+import java.io.Closeable;
+
+import lombok.SneakyThrows;
+import lombok.Value;
+import lombok.experimental.NonFinal;
```
marcingrzejszczak: I'm sorry if I'm out of line and I should only comment on the trace-id related stuff but is it "safe" to use experimental features of Lombok in a library of Spring-Cloud caliber?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebAutoConfiguration.java
```diff
@@ -0,0 +1,87 @@
...
+
+	@Bean
+	public WebMvcConfigurerAdapter webMvcConfigurerAdapter(
+			TraceHandlerInterceptor handlerInterceptor) {
+		return new TraceWebConfigurer(handlerInterceptor);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public FilterRegistrationBean correlationHeaderFilter() {
```
marcingrzejszczak: `traceHeaderFilter`?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/slf4j/Slf4jSpanStartedListener.java
```diff
@@ -0,0 +1,25 @@
...
+ * @author Spencer Gibb
+ */
+@Slf4j
+public class Slf4jSpanStartedListener implements ApplicationListener<SpanStartedEvent> {
+
+	@Override
+	public void onApplicationEvent(SpanStartedEvent event) {
+		Span span = event.getSpan();
+		//TODO: what log level?
+		log.info("Starting span: {}", span);
```
marcingrzejszczak: In our production cases INFO was really a good thing cause we saw where the initalization of the whole business process took place. 

On the other hand we had to introduce filtering if the URIs cause Swagger users flooded our logs with meaningless entries. But after setting the patterns it's really ok right now. So IMO it should be info

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/DefaultTraceTests.java
```diff
@@ -0,0 +1,120 @@
...
+import org.springframework.cloud.sleuth.event.SpanStoppedEvent;
+import org.springframework.cloud.sleuth.sampler.AlwaysSampler;
+import org.springframework.cloud.sleuth.sampler.IsTracingSampler;
+import org.springframework.context.ApplicationEvent;
+import org.springframework.context.ApplicationEventPublisher;
+
+/**
+ * @author Spencer Gibb
+ */
+public class DefaultTraceTests {
```
marcingrzejszczak: Why not Spock? :P

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/scheduling/TraceSchedulingAspect.java
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @author Tomasz Nurkewicz, 4financeIT
+ * @author Michal Chmielarz, 4financeIT
+ * @author Marcin Grzejszczak, 4financeIT
+ * @author Spencer Gibb
+ *
+ * @see Trace
+ */
+@Aspect
+public class TraceSchedulingAspect {
```
marcingrzejszczak: That's pretty much exactly the same what we did if I remember it correctly

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/circuitbreaker/TraceCommand.java
```diff
@@ -0,0 +1,59 @@
...
+import com.netflix.hystrix.HystrixCommandGroupKey;
+import com.netflix.hystrix.HystrixThreadPoolKey;
+import org.springframework.cloud.sleuth.Trace;
+import org.springframework.cloud.sleuth.TraceScope;
+
+/**
+ * Abstraction over {@code HystrixCommand} that wraps command execution with Trace setting
+ *
+ * @see HystrixCommand
+ * @see CorrelationIdUpdater
```
marcingrzejszczak: I guess this should be removed

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
```diff
@@ -0,0 +1,118 @@
...
+		this.skipPattern = DEFAULT_SKIP_PATTERN;
+	}
+
+	public TraceFilter(Trace trace, Pattern skipPattern) {
+		this.trace = trace;
+		this.skipPattern = skipPattern;
+	}
+
+	@Override
+	protected void doFilterInternal(HttpServletRequest request,
```
marcingrzejszczak: This is the same point of entry that we had - in comes the request and you pass / add "trace-id" headers, right?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceRestTemplateInterceptor.java
```diff
@@ -0,0 +1,57 @@
...
+ * Interceptor that verifies whether the trance and span id has been set on the
+ * request and sets them if one or both of them are missing.
+ *
+ * @see org.springframework.web.client.RestTemplate
+ * @see Trace
+ *
+ * @author Marcin Grzejszczak, 4financeIT
+ * @author Spencer Gibb
+ */
+public class TraceRestTemplateInterceptor implements ClientHttpRequestInterceptor {
```
marcingrzejszczak: This is the same for ensuring that when using `RestTemplate` the requests have the proper headers

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/slf4j/Slf4jSpanStartedListener.java
```diff
@@ -0,0 +1,25 @@
...
+		MDC.put(Trace.SPAN_ID_NAME, span.getSpanId());
```
marcingrzejszczak: And there we have MDC.put with the former 'Correlation-Id' and current 'Trace-id'.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
spencergibb: Yes

marcingrzejszczak:
Maybe I'm wrong but we have also this:

https://github.com/4finance/micro-infra-spring/blob/master/micro-infra-spring-base/src/main/groovy/com/ofg/infrastructure/web/correlationid/CorrelationIdAspect.java

Where we try to ensure that if someone is using Servlet 3.0 support with the Callable<> in Controllers that we pass the proper correlationid. I don't see it here - is it deliberate or did you solve it by some other means?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/slf4j/Slf4jSpanStartedListener.java
spencergibb: :+1:

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceRestTemplateInterceptor.java
spencergibb: :+1:

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/DefaultTraceTests.java
spencergibb: I'm lazy? ;-)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/slf4j/Slf4jSpanStartedListener.java
spencergibb: good to know

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/DefaultTraceTests.java
marcingrzejszczak: Oh come on :D You would have done it in Spock in no time with less code :P

spencergibb:
I wondered about Callable in Controllers, right now I trace _ALL_ controllers.  But you're concerned with the thread there right?

spencergibb:
I've replaced references to `correlationId` with references to tracing.

marcingrzejszczak:
Yeah exactly - we've also traced all controllers since we had a filter. However a new thread is spawned and you loose correlationID info. That's why we did the aspect.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/TraceScope.java
spencergibb: I'm not worried about it.  Lombok doesn't "ship" with spring cloud it's just compile time.

spencergibb:
OK, I'll test that and make sure `Callables` returned from controllers still work.

spencergibb:
That would also be an issue when Observable support https://github.com/spring-cloud/spring-cloud-netflix/pull/246 is in spring cloud netflix.

marcingrzejszczak:
Definitely...

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/DefaultTraceTests.java
spencergibb: I don't know spock yet, so 'no time' might be a stretch.  I want to learn it though.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/TraceScope.java
marcingrzejszczak: :+1:

spencergibb:
Htrace has some [wrappers](https://git-wip-us.apache.org/repos/asf?p=incubator-htrace.git;a=tree;f=htrace-core/src/main/java/org/apache/htrace/wrappers;h=c5e9dc01c82ac3222c40a371661a7f163017b206;hb=HEAD) that could be used there with developer education.

marcingrzejszczak:
Awesome! That would be great!

spencergibb:
@marcingrzejszczak added fix for returning `Callable` from a `Controller`.

marcingrzejszczak:
With Callables and Controllers it seems that everything from our side is there.

BTW @dsyer 2 questions:

1) what next could be merged from our side? 
2) From our perspective it would be cool if you released some 0.0.1 non snapshot version of both Zookeeper and Sleuth. That way we would plug it in into our stuff to see what's missing and how it's actually working. 

WDYT?

spencergibb:
@marcingrzejszczak I'll look at a milestone soon.

marcingrzejszczak:
BTW @spencergibb  - why did you remove all the spock tests? :P you could have reused them

PR #6, opened 2015-07-28
Added Span.Type so we can differentiate between client and server events

spencergibb:
@dsyer What do you think?  In brave, you would use a `ClientTracer` or `ServerTracer`.


spencergibb:
closed in favor of #7

Issue #14, opened 2015-08-06
Instrumentation for spring integration


spencergibb:
Email thread as first thoughts.

----
Date: Tue, 4 Aug 2015 16:26:46 -0600
Subject: Tracing messages and spring integration
From: Spencer Gibb
To: Artem Bilan, Gary Russell

Artem and Gary,

I'm looking to add support to [spring-cloud-sleuth](https://github.com/spring-cloud/spring-cloud-sleuth) for tracing (start/end timings, etc...) spring integration messages, especially those that cross thread boundaries.

Dave pointed me at the following:

https://jira.spring.io/browse/INT-2166
https://github.com/spring-projects/spring-integration/pull/1493

Especially [SecurityContextPropagationChannelInterceptor.java]
(https://github.com/artembilan/spring-integration/blob/4e3f36c8ab6767cd141e9ee29b5efce924b1700b/spring-integration-security/src/main/java/org/springframework/integration/security/channel/SecurityContextPropagationChannelInterceptor.java) because it is similar to how the tracing context is held (ie a ThreadLocal).

Is this a good model to follow?  If not, any other ideas?

----
Date: Tue, 4 Aug 2015 20:20:13 -0400
Subject: Re: Tracing messages and spring integration
From: Gary Russell
To: Spencer Gibb
Cc: Artem Bilan

Yes, we made it a subclass of for ThreadStatePropagationChannelInterceptor exactly that reason (propagating other thread-based state).

However, we don't currently have a solution for propagating across the network (which is a bit more tricky for security, we don't want to simply convey an Authentication - the sender might not be trusted).

If you need to propagate it across loosely coupled integration-based apps (e.g. via rabbitmq etc), it might be better to carry the trace information in a message header, similar to MessageHistory (when enabled in SI).

We added something similar to XD to convey XD module transits by adding to the header as the message transits each module in an XD stream.

Hope that helps.

Gary

----
Subject: Re: Tracing messages and spring integration
Date: Wed, 5 Aug 2015 09:22:58 +0100
To: Spencer Gibb

Yeah message headers sound like the best place to put the correlation ids. But the injection point (and how to automate it/ optimize it) might be similar. We just need a way to catch the incoming requests mainly.

dsyer:
The security context interceptor is a good building block: it keeps the thread context up to date when Spring Integration hands off between threads. Missing are:

[...]

PR #15, opened 2015-08-06
Instrument spring integration.

spencergibb:
Via a global channel interceptor that adds headers to messages.

fixes gh-14

For review @dsyer @garyrussell @artembilan


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/integration/TraceContextPropagationChannelInterceptor.java
```diff
@@ -0,0 +1,203 @@
...
+
+	private String getParentId(Span span) {
+		return span.getParents() != null && !span.getParents().isEmpty() ? span
+				.getParents().get(0) : null;
+	}
+
+	protected void populatePropagatedContext(Span span, Message<?> message,
+											 MessageChannel channel) {
+		if (span != null) {
+			Span currentContext = getCurrentSpan();
```
spencergibb: I'm still fuzzy on what I should be doing with the span here.  This is leftover from `SecurityContextPropagationChannelInterceptor`.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/integration/TraceContextPropagationChannelInterceptor.java
```diff
@@ -0,0 +1,203 @@
...
+	@Override
+	@SuppressWarnings("unchecked")
+	public final Message<?> postReceive(Message<?> message, MessageChannel channel) {
+		if (message instanceof MessageWithSpan) {
+			MessageWithSpan messageWithSpan = (MessageWithSpan) message;
+			Message<?> messageToHandle = messageWithSpan.message;
+			populatePropagatedContext(messageWithSpan.span, messageToHandle, channel);
+
+			publish(new ClientReceivedEvent(this, messageWithSpan.span));
+			return message;
```
spencergibb: I'm returning the `message` itself and not `messageWithSpan.message` because I want the combined headers in `message`.  Is this safe? Another alternative would be to return `messageWithSpan.message` wrapped with `message`'s headers.

dsyer: I don't know about the prior art much, but it seems more idiomatic to use `Message` for everything and rely on headers to convey correlation data. Maybe a custom implementation of `MessageHeaders` would be better? Or just a helper that knows about specific sleuthy header names.

artembilan:
@spencergibb , @dsyer , your PR is on our radar. 
Let us know how quickly you need a feedback from us.
The next week is a "release" one for us, so we are very busy right now to wrap up everything we want for them.

Thanks for your patient!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/integration/TraceContextPropagationChannelInterceptor.java
spencergibb: So put the span in a header as well?  The `MessageWith*` came from how they did the security context.  The difference here is that we want to modify headers, where the security context did not.

dsyer:
@spencergibb I rebased onto master with that last commit. I see what you are doing now, and it's quite narrowly restricted to thread context propagation. We need more components so we can actually manage the context (create spans, mark them as ended etc.) when messages enter and leave the system. That is probably going to be harder, so maybe we need to ask @artembilan and Gary for help. Or maybe it needs to be aware of spring-cloud-stream (cc @markfisher).

PR #21, opened 2015-08-12
Added traceid setting for feign

marcingrzejszczak:


spencergibb:
Added other headers (needed for zipkin), polished and merged via acf7417e219dd8be3c4aebea0c0686cdfa24b321

PR #28, opened 2015-08-18
Add MilliSpan.getAnnotations()

rwinch:
This fixes the previous commit.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/MilliSpan.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/MilliSpan.java
@@ -17,6 +17,7 @@
 package org.springframework.cloud.sleuth;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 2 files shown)


spencergibb:
Since this uses lombok, you just need to change the name of the variable from `kVAnnotations` to `annotations`.

rwinch:
@spencergibb Do you want to expose it as a setter too?

spencergibb:
it's final, so it doesn't.  We currently don't need to.

rwinch:
@spencergibb Thanks I didn't catch that. Will it also be immutable as documented on the interface?

spencergibb:
That's probably the biggest inconsistency. One of the two needs to change :-) If that's the case, your change make a lot more sense.  We should also change `getTimelineAnnotations`.

rwinch:
@spencergibb 

> One of the two needs to change :-)  

I'm fine with whichever you feel is best. I just wanted the project to build :)

> We should also change getTimelineAnnotations.

If it is ok with you, I'll leave this up to you since I'm not sure if you wanted to change the interface documentation or update the implementation.

Issue #27, opened 2015-08-18
Use ThreadLocal.remove() to prevent memory leaks


rwinch:
Using `ThreadLocal.set(null)` to clear a value out causes a memory leak when `ThreadPool`s are used. The problem is this:

* `ThreadLocal.set(null)` will lookup a `Map` associated with the current `Thread`.
* It sets the key of the map to `this` and a value of `null`

The `Thread` is then contained within a `Thread` pool that outlasts the application. So if the application is un-deployed, the reference to `this` remains within the `Thread`.

At first glance the problem may sound fairly minor. However, in many instances `this` is loaded by the child `ClassLoader` (i.e. the Web Application's `ClassLoader`). One example of how this happens is when using `NamedThreadLocal`.

The implication is that `this.getClass().getClassLoader()` will hold a reference to an un-deployed application's `ClassLoader` which refers to all the classes it loaded. This means there is a rather large leak within the application.

To avoid this, `ThreadLocal.remove()` should be used instead. The difference is that the `Map` entry is moved entirely vs mapping the key to a `null` value.

One example of the issue can be found in `TraceContextHolder.currentSpan`. There appear to be other places in Spring Cloud Sleuth that this should be fixed too.

nuotang:
登良

PR #30, opened 2015-08-18
Use ThreadLocal.remove() rather than .set(null)

spencergibb:
fixes gh-27

/cc @rwinch


rwinch:
Looks good to me :)

cincat:
@rwinch @spencergibb I think the memory leaks is not caused by ThreadLocal.set(null), because Thread just hold the weak reference of ThreadLocal variable, the strong relationship held is by the thread itself.

marcingrzejszczak:
You're referring to the code that is 6 years old. There's no such code anymore!

cincat:
It's okay, I just wanna to confirm that ThreadLocal.set(null) cannot cause memory leaks.

cincat:
And I am certain that the code is never changed since 6 years ago.

spencergibb:
@cincat there's no such class `TraceContextHolder` anymore.

PR #31, opened 2015-08-19
add /messaging annotations

spencergibb:
Created using https://github.com/scratches/voter-counter

Could ignore tracing headers.

@dsyer @adriancole @mstine @b

screenshotsleuthmessaging

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Trace.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Trace.java
@@ -16,6 +16,8 @@
 
 package org.springframework.cloud.sleuth;
 
+import java.util.Arrays;
+import java.util.List;
 import java.util.concurrent.Callable;
 
 /**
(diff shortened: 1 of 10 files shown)


dsyer:
Yeah, need to ignore the tracing headers.

spencergibb:
tracing headers ignored.

dsyer:
That's a start. We could say something about the payload (type, size if known, e.g. if it's a String or byte[])? ANy other ideas @garyrussell, @markfisher ?

adriancole:
I like the namespacing

Food for thought: we will eventually need to think about the size of this data. Going from binary data with integer keys (thrift or protobuf) to uncompressed json with lengthy key names is probably an order of magnitude more data. Add to this that the biggest data issue in observability at Twitter is most was never read. Should we be auto attaching things like accept or content header values? Would this help in latency analysis?

spencergibb:
maybe there can be some kind of opt-in or opt-out and CF could be opinionated on what that is.

spencergibb:
Closed via 0f76427ebf48b2f0838aa725c882c101e821f543

PR #53, opened 2015-11-07
Updates to latest zipkin/brave 

adriancole:
Zipkin now has Span.timestamp/duration, so no more need for
"aquire/release".

This also corrects the host/endpoint logged for timeline annotations, as
it should always be the local process.

See https://github.com/openzipkin/zipkin/blob/master/zipkin-thrift/src/main/thrift/com/twitter/zipkin/zipkinCore.thrift

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -48,8 +48,6 @@ logging:
 8. Hit `http://localhost:3380`, `http://localhost:3380/call`, `http://localhost:3380/async` for some interesting sample traces (the app callas back to itself).
 9. Goto `http://localhost:8082` for zipkin web (8080 if running locally from source, the host is the docker host, so if you are using boot2docker it will be different)
 
-WARNING: The docker images for zipkin are old and don't work very well (the UI in particular). Zipkin is in a state of flux, but it should settle down soon when there is an actual release. Best results actually come from building from source and running the jar files (the query and collector services need command line arguments, so check the zipkin README for updates).
-
 NOTE: You can see the zipkin spans without the UI (in logs) if you run the sample with `sample.zipkin.enabled=false`.
 
 There are a few samples with slightly different features:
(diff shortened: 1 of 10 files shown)


spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinSpanListener.java
```diff
@@ -59,6 +59,7 @@ public ZipkinSpanListener(SpanCollector spanCollector, EndpointLocator endpointL
 	@EventListener
 	@Order(0)
 	public void start(SpanAcquiredEvent event) {
+		// Starting a span in zipkin means adding: traceId, id, parentId(optional), and timestamp
 		event.getSpan().addTimelineAnnotation("acquire");
```
adriancole: basically, in a bit, we can remove the acquire/release annotations as they are obviated by the new fields Span.timestamp and duration

adriancole:
No more fake acquire/release annotations needed when not using core annotations.

screen shot 2015-11-19 at 11 15 32 am

README.adoc
```diff
@@ -48,8 +48,6 @@ logging:
 8. Hit `http://localhost:3380`, `http://localhost:3380/call`, `http://localhost:3380/async` for some interesting sample traces (the app callas back to itself).
 9. Goto `http://localhost:8082` for zipkin web (8080 if running locally from source, the host is the docker host, so if you are using boot2docker it will be different)
 
-WARNING: The docker images for zipkin are old and don't work very well (the UI in particular). Zipkin is in a state of flux, but it should settle down soon when there is an actual release. Best results actually come from building from source and running the jar files (the query and collector services need command line arguments, so check the zipkin README for updates).
```
adriancole: docker images are published as a part of release now. Not really much flux in this category anymore.

spencergibb:
:+1:

PR #55, opened 2015-11-14
An example for tracing an application using Apache Camel, but also for discussion about programatic usage

alexec:
This PR is to see if I've got programmatic usage of Spring Sleuth correct.


marcingrzejszczak:
HI! You can check out this: https://github.com/4finance/micro-infra-spring/blob/master/micro-infra-camel/src/main/java/com/ofg/infrastructure/camel/aspects/CorrelationIdOnCamelRouteAspect.java and this: https://github.com/4finance/micro-infra-spring/blob/master/micro-infra-camel/src/main/java/com/ofg/infrastructure/camel/CorrelationIdInterceptor.java which seems to be working :)

PR #58, opened 2015-11-24
Providing support to enable tracing in Executors viz. java.unit.concurrent package

gauravrmazra:
I have added classes to provide support for tracing in executors. I was evaluating to use spring-cloud-sleuth in one of our project for tracing. I saw it was not providing support for Executors. Now, I have written a glue code to provide support, this is what I wrote to support spring-cloud-sleuth in our application. I hope this will be helpful to this project as well.


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/concurrent/executor/support/TraceCallableSupport.java
```diff
@@ -0,0 +1,37 @@
...
+ * A helper class to decorate {@link Callable} to {@link TraceCallable}
+ * @author Gaurav Rai Mazra
+ *
+ */
+public class TraceCallableSupport {
+	private TraceCallableSupport() {
+		
+	}
+	
+	public static <V> Callable<V> decorate(Trace trace, Callable<V> command) {
```
spencergibb: Seems redundant to me. Just replace it with `new TraceCallable<>(trace, command);`

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/concurrent/executor/support/TraceRunnableSupport.java
```diff
@@ -0,0 +1,35 @@
...
+ * A helper class to decorate {@link Runnable} to {@link TraceRunnable}
+ * @author Gaurav Rai Mazra
+ *
+ */
+public class TraceRunnableSupport {
+	private TraceRunnableSupport() {
+		
+	}
+	
+	public static Runnable decorate(Trace trace, Runnable delegate) {
```
spencergibb: Seems redundant to me. Just replace it with new TraceRunnable(trace, delegate);

dsyer:
Maybe a test or two would smooth this one on its way?

gauravrmazra:
I will sure write test cases for this.

gauravrmazra:
I have written a test case for TraceableExecutorService and to support my claim in issue
https://github.com/spring-cloud/spring-cloud-sleuth/issues/60

Issue #61, opened 2015-11-30
Doubt for close() method of TraceScope class


gauravrmazra:
I was seeing close() method of TraceScope class. In this method, current thread's span is replaced with other span(savedSpan). See line number 108. Why is this done this way?

dsyer:
Why do you think it's a problem?

gauravrmazra:
In TraceCallable and TraceRunnable, first span is started and we get TraceScope reference and run the delegated Thread(Runnable or Callable) and when it is done. We call close() of that particular scope, which does as mentioned in above question. 
My concern is related to Executors. If one executor is shared to do some process and multiple requests are coming which is catered by some service using executors(It is just a use case). Then in worker threads of executors, the above code is creating problem and lead to Htrace error message when space is started. This is the reason I asking for this way. Isn't it if Worker thread is done its processing then it should clear its ThreadLocal context as well.

I will write some sample app with code which is able to reproduce this problem and with possible solution.

dsyer:
I'm not sure I see the problem yet, but a sample project will certainly help. BTW I'm in the middle of a big refactor of all the core APIs so the names and locations of things might change soon.

gauravrmazra:
I have created the sampleapp
https://github.com/gauravrmazra/spring-cloud-sleuth-executor-sampleapp

If multiple request are send from UI, it throws weird Htrace error (runtime exception)

The solution is simple. Just remove the current span from ThreadLocal context of the worker thread within TraceCallable and TraceRunnable

PR #64, opened 2015-12-01
Added tests for trace callable and runnable

marcingrzejszczak:
@spencergibb @dsyer - can you please take a look at this and tell me if also in your opinion those tests should pass? :)


spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/TraceCallableTest.java
```diff
@@ -0,0 +1,62 @@
...
+	TraceManager traceManager = Mockito.mock(TraceManager.class);
+
+	@Test
+	@Ignore("Will fail because trace is not removed after callable gets executed")
+	public void should_remove_span_from_thread_local_after_finishing_work() throws Exception {
+		givenCallableGetsSubmitted(thatSetsTraceInCurrentThreadLocalWithInitialTrace());
+
+		Trace secondTrace = whenCallableGetsSubmitted(thatRetrievesTraceFromThreadLocal());
+
+		then(secondTrace).isNull();
```
dsyer: I don't think this test should pass.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/TraceCallableTest.java
```diff
@@ -0,0 +1,62 @@
...
+	}
+
+	private Callable<Trace> thatSetsTraceInCurrentThreadLocalWithInitialTrace() {
+		return new Callable<Trace>() {
+			@Override
+			public Trace call() throws Exception {
+				TraceContextHolder.setCurrentTrace(Mockito.mock(Trace.class));
```
dsyer: Usage of `TraceContextHolder` is basically banned or at least heavily discouraged so this should not be in any real user code.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/TraceCallableTest.java
dsyer: This one should pass:

```
@RunWith(MockitoJUnitRunner.class)
public class TraceCallableTest {

	ExecutorService executor = Executors.newSingleThreadExecutor();
	TraceManager traceManager = new DefaultTraceManager(new AlwaysSampler(),
			new RandomUuidGenerator(), Mockito.mock(ApplicationEventPublisher.class));

	@Test
	public void should_remove_span_from_thread_local_after_finishing_work()
			throws Exception {
		Trace firstTrace = givenCallableGetsSubmitted(thatSetsTraceInCurrentThreadLocalWithInitialTrace());

		Trace secondTrace = whenCallableGetsSubmitted(thatRetrievesTraceFromThreadLocal());

		then(secondTrace.getSpan().getTraceId()).isNotEqualTo(firstTrace.getSpan().getTraceId());
	}

	private Callable<Trace> thatSetsTraceInCurrentThreadLocalWithInitialTrace() {
		return new Callable<Trace>() {
			@Override
			public Trace call() throws Exception {
				return TraceContextHolder.getCurrentTrace();
			}
		};
	}

	private Callable<Trace> thatRetrievesTraceFromThreadLocal() {
		return new Callable<Trace>() {
			@Override
			public Trace call() throws Exception {
				return TraceContextHolder.getCurrentTrace();
			}
		};
	}

	private Trace givenCallableGetsSubmitted(Callable<Trace> callable)
			throws InterruptedException, java.util.concurrent.ExecutionException {
		return whenCallableGetsSubmitted(callable);
	}

	private Trace whenCallableGetsSubmitted(Callable<Trace> callable)
			throws InterruptedException, java.util.concurrent.ExecutionException {
		return this.executor.submit(new TraceCallable<>(this.traceManager, callable))
				.get();
	}

}
```

dsyer:
I merged this by mistake. If you want we can still discuss the expected behaviour here.

marcingrzejszczak:
The problem emerges especially when using Hystrix commands. If we're using shared thread pools then there is an issue that a Callable / Runnable is used within a thread that has already processed a Callable / Runnable. That means that in a ThreadLocal there has already been a Span and a new Span is being processed by that Callable / Runnable. In that situation an exception is thrown because there should be nothing there set in the ThreadLocal. 

That's why I wanted to clear the entries inside the ThreadLocal after a Callable / Runnable processing took place.

Another issue is that the Hystrix command is not storing the current Trace from the current thread like it should (https://github.com/4finance/micro-infra-spring/blob/master/micro-infra-spring-base/src/main/java/com/ofg/infrastructure/hystrix/CorrelatedCommand.java#L18). But this is a topic for another PR actually...

dsyer:
Probably the same issue being discussed in #61? I'll change the test and see if it helps flush out the stale trace.

marcingrzejszczak:
Yup it's exactly the same thing.

When I was doing testing the app blew up in places with common threadpools
and tracedelegates.

PR #65, opened 2015-12-01
Added passing of trace data

marcingrzejszczak:
@dsyer, @spencergibb - I've modified the way trace data is passed in the filter. IMO we should always pass the values of tracing to the response. If the data is not there we should generate it.


dsyer:
I can see that maybe we should send something back to the caller, but I don't think we should generate trace and span ID if the `TraceManager` doesn't. I.e. I believe we do not need an independenct `IdGenerator` here, but rather we can simply pull the span and trace ID from the trace that is active when the response is sent.

dsyer:
This also raises the question of why `IdGenerator` is needed anywhere else at all. (And also why it isn't a normal Spring `IdGenerator`, but that's a separate issue.)

dsyer:
Hopefully I have captured your and my concerns in 1929647 (which I believe is a lot simpler).

marcingrzejszczak:
Yeah it seems so :)

PR #66, opened 2015-12-02
Added test for TraceCommand together with potential fix

marcingrzejszczak:
@dsyer , @spencergibb - WDYT?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,7 +8,7 @@ before_install:
   - gem install asciidoctor
 install:
 - ./mvnw install -P docs -q -U -DskipTests=true -Dmaven.test.redirectTestOutputToFile=true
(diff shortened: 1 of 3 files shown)


dsyer:
LGTM. Why doesn't it suffer from the same problem as the `TraceCallable` and `TraceRunnable` (#61) - we don't detach the parent trace from the thread at the end of the command?

marcingrzejszczak:
I've added a test and it's working thanks to the changes in the `close` method. But actually there would be the very same issue as in [#61]

marcingrzejszczak:
Done in https://github.com/spring-cloud/spring-cloud-sleuth/commit/c1eb04b68aa3cabfe34fae6699e68d8f29055ac3

PR #68, opened 2015-12-07
Support for STOMP messages over websocket

gauravrmazra:


marcingrzejszczak:
Hi!  Thanks for the PR :)

I have some questions though before we proceed. 

- Did you sign the Contributor License Agreement? 
- Could you please add some integration tests to show that the feature is working?

gauravrmazra:
Hey Marcin

Yes, I have signed the Contributor License agreement.

Sure, I will write some integration test and will submit in a day or two.

gauravrmazra:
Just an update, I will submit integration tests on weekend.

marcingrzejszczak:
Great! Thanks a lot :)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/integration/StompMessageBuilderHelper.java
```diff
@@ -0,0 +1,137 @@
...
+				span.addAnnotation(key, value);
+			}
+		}
+		
+		Object payload = message.getPayload();
+		if (payload != null) {
+			span.addAnnotation("/messaging/payload/type", payload.getClass().getCanonicalName());
+			
+			if (payload instanceof String) {
+				span.addAnnotation("/messaging/payload/size", String.valueOf(((String)payload).length()));
```
adriancole: this can be misleading as size is used both for bytes and string length (which depending on encoding could be quite different)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/integration/StompMessageBuilderHelper.java
```diff
@@ -0,0 +1,137 @@
...
+				span.addAnnotation("/messaging/payload/size", String.valueOf(((byte[])payload).length));
+			}
+		}
+	}
+	
+	static class StompMessageBuilder {
+		private Map<String, Object> headers = new TreeMap<String, Object>();
+		private Message<?> message;
+		
+		public StompMessageBuilder(final Message<?> message) {
```
adriancole: nit: public methods on a package private type

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/integration/StompMessageBuilderHelper.java
```diff
@@ -0,0 +1,137 @@
...
+import org.springframework.messaging.Message;
+import org.springframework.messaging.simp.SimpMessageHeaderAccessor;
+import org.springframework.messaging.simp.SimpMessageType;
+
+/**
+ * Helper class to create STOMP message
+ * @author Gaurav Rai Mazra
+ *
+ */
+public class StompMessageBuilderHelper {
```
adriancole: This type is a bit of a mouthful. If a builder needs a helper, maybe there's an opportunity to simplify?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/integration/TraceStompMessageChannelInterceptor.java
```diff
@@ -0,0 +1,107 @@
...
+		String name = "message/" + getChannelName(channel);
+		Trace trace;
+		if (hasText(spanId) && hasText(traceId)) {
+
+			MilliSpanBuilder span = MilliSpan.builder().traceId(traceId).spanId(spanId);
+			String parentId = getHeader(message, Trace.PARENT_ID_NAME);
+			String processId = getHeader(message, Trace.PROCESS_ID_NAME);
+			String spanName = getHeader(message, Trace.SPAN_NAME_NAME);
+			if (spanName != null) {
+				span.name(spanName);
```
adriancole: not sure the intent of the defensive programming here. Are you expecting a call to span.name(null) to throw? (I don't know how lombok behaves, so maybe it does)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/integration/StompMessageBuilderHelper.java
```diff
@@ -0,0 +1,137 @@
...
+					value = entry.getValue();
+					pushHeaders(messageHeaderAssessor, key, value);
+				}
+			}
+
+			return org.springframework.messaging.support.MessageBuilder.createMessage(this.message.getPayload(),
+					messageHeaderAssessor.getMessageHeaders());
+		}
+
+		private void pushHeaders(final SimpMessageHeaderAccessor assessor, final String key, final Object value) {
```
adriancole: typo SimpMessageHeaderAccessor

adriancole:
looking forward to the feature. There's a bit of cleanup warranted, as well unit and integration tests (especially due to the amount of branching). However, I'm really happy you were able to share the work in progress. Keep it up!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/integration/StompMessageBuilderHelper.java
gauravrmazra: Before that there is payload type is also shown. So, it might not create any confusion. I am not sure if encoding type in case of String need to be shown as annotation.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/integration/StompMessageBuilderHelper.java
gauravrmazra: I will be refactoring this static nested class and the helper class soon. When I started writing it, there was no Helper class. I was bit confused in putting addAnnotations method in builder, so created a helper class.

gauravrmazra:
Thanks for reviewing and providing you comments. I will be refactoring it and will be submitting integration tests on weekend.

gauravrmazra:
@dsyer @spencergibb 

The build is failed due to org.springframework.messaging packages not found.
Some stack trace is:

[ERROR] COMPILATION ERROR : 
[ERROR] /home/travis/build/spring-cloud/spring-cloud-sleuth/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/integration/TraceContextPropagationChannelInterceptor.java:[27,37] package org.springframework.messaging does not exist

Will you please look into it? Is it my problem or the latest changes in the build script?

dsyer:
Works for me. Maybe we have a bad jar in our cache in travis. Can you build locally?

gauravrmazra:
@dsyer  It was failing locally.

Dependency scope as test is removed for spring messaging

dsyer:
It's working on master. Must be something in your PR?

gauravrmazra:
@dsyer  Now, it passed all the checks. Scope for one dependency("spring-messaging") was test. I removed it.

gauravrmazra:
@adriancole  Is this PR need any improvement? it is still not merged from last 25 days

adriancole:
Always room for improvement :) suppose real question  is about blocking
comments from me. I'm happy that you  have tests, now, and the structures
are simpler.

If no-one else has a blocking comment,  and generally folks think this
should be in the main repo, we should merge this.

The 'should this feature exist' question is generally easy to answer when
more than just the author says they have stake in this merged (often
because they don't want to copy this into a fork).

Do you know of anyone else who'd use this?

Regardless, I have no blocking comments personally (even though I have
plenty of nits I could mention)

gauravrmazra:
I will update the comments. This feature will most likely will be required by anyone who want to use Stomp messages over JMS. This is used in my company's one of project at least one I am working on :)
I think this is a general feature similar to JMS messages.

marcingrzejszczak:
If to use JMS you're using Spring Integration (or Spring Stream) then we already have support for that in Sleuth.

gauravrmazra:
Sleuth doesn't provide support for STOMP messages. I was saying that if I
am using any JMS broker like ActiveMQ which support STOMP protocol, than
also tracing was not possible even when Spring Integration is used.

marcingrzejszczak:
Merged in https://github.com/spring-cloud/spring-cloud-sleuth/commit/502b6b1f850772046fb1706ae55db273e818ed24

PR #72, opened 2015-12-14
Cleans up unused brave refs and consolidates to zipkin docker 1.26.0

adriancole:


adriancole:
k should be good to go now.. will merge on green

Issue #39, opened 2015-09-08
Pass trace data via @HystrixCommand


spencergibb:

spencergibb:
See https://github.com/Netflix/Hystrix/issues/92#issuecomment-12848797

morfeo8marc:
I've read the issues, and also taken a look at your extension of HistrixCommand. What is the purpouse of this extension?

What should be use on combimatiom with @HistrixCommamd to allow tracing happen? Should a custom concurrent strategy be implememted?

spencergibb:
@morfeo8marc the purpose of the extension is to use is when you construct a `HystrixCommand` by hand. This issue is to support passing trace data via `@HystrixCommand`.

dnathanson:
Just ran into this issue.  :+1:

dnathanson:
When this is implemented, will it handle the case where Hystrix isolation policy is THREAD?  I tried using extending TraceCommand instead of annotating with @HystrixCommand and while it does create a span, the trace context is not passed to Hystrix's child thread so the parent/child relationship between spans is lost.

swehner:
The workaround of using a custom ConcurrencyStrategy works for this case:
```
hystrix.plugin.HystrixConcurrencyStrategy.implementation=SpanPreservingConcurrencyStrategy
```
This would be the implementation:
```
public class SpanPreservingConcurrencyStrategy extends HystrixConcurrencyStrategy {
	@Override
	public <T> Callable<T> wrapCallable(Callable<T> callable) {
		return new SpanPreservingCallable<T>(callable);
	}
	
	public static class SpanPreservingCallable<T> implements Callable<T> {
		private Span currentSpan;
		private Callable<T> callable;

		public SpanPreservingCallable(Callable<T> callable) {
			this.callable = callable;
			this.currentSpan = TraceContextHolder.getCurrentSpan();
		}
		
		@Override
		public T call() throws Exception {
			// set span of the calling thread
			TraceContextHolder.setCurrentSpan(currentSpan);
			try {
				return callable.call();
			} finally {
				// reset span
				TraceContextHolder.setCurrentSpan(null);
			}
		}
	}
}
```
I've tried to make this more extensible so it could be included in spring-cloud-netflix, but I've encountered several problems as this class is instantiated by Hystrix and not by Spring...

spencergibb:
Thanks!

adriancole:
Good point that hystrix needs to propagate data. It is probably also
[...]

PR #78, opened 2015-12-23
Issues #39 hystrix concurrency strategy

marcingrzejszczak:
CC @spencergibb , @dsyer , @adriancole 

I'll add Javadocs after we agree on the approach


spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/hystrix/JavanicaITest.java
```diff
@@ -0,0 +1,125 @@
...
+import org.springframework.context.annotation.EnableAspectJAutoProxy;
+import org.springframework.test.context.junit4.SpringJUnit4ClassRunner;
+
+import com.jayway.awaitility.Awaitility;
+import com.netflix.hystrix.contrib.javanica.annotation.HystrixCommand;
+
+@RunWith(SpringJUnit4ClassRunner.class)
+@SpringApplicationConfiguration(classes = {
+		JavanicaITest.JavanicaITestConfiguration.class })
+public class JavanicaITest {
```
adriancole: I like the idea of this, but I have to know too many things, or navigate multiple files to figure out what Javanica is and what its relationship to the setup ceremony is.

What if you inline a Javanica class and use it here (preferably as the first member type, at the top of this file)? Then review would be possible without navigating to look at files not touched.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/hystrix/JavanicaITest.java
```diff
@@ -0,0 +1,125 @@
...
+	private void whenHystrixCommandGetsExecutedViaJavanica() {
+		this.javanicaDelegation.doSthThatDelegatesToJavanica();
+	}
+
+	private void thenSpanPutInTheAsyncThreadIsSameAs(final Span span) {
+		Awaitility.await().until(new Runnable() {
+			@Override
+			public void run() {
+				then(span.getTraceId()).isNotNull()
+						.isEqualTo(javanicaClass.getTraceId());
```
adriancole: does the assertion language require you to do .isNotNull first? I'm used to assertj, which always checks null nicely first. It means less boilerplate and easier to read tests.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/hystrix/JavanicaITest.java
```diff
@@ -0,0 +1,125 @@
...
+				then(span.getName())
```
adriancole: mental note export: if we do this type of thing quite a lot, making a span assert would cut down on clutter and allow chaining. ex. `then(span).hasTraceId(foo).hasName(bar)`

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/hystrix/JavanicaITest.java
marcingrzejszczak: I can change the name of the test. `HystrixViaAnnotationsITest` or sth like that.

I don't get the ` inline a Javanica class ` part - can you provide an example? You want me to move the class from the bottom of the file upwards?

I prefer to have the test as such and the crucial fields on top and the technical rest in the bottom part. But of course I can change that.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/hystrix/JavanicaITest.java
```diff
@@ -0,0 +1,125 @@
...
+		public JavanicaDelegation(JavanicaClass javanicaClass) {
+			this.javanicaClass = javanicaClass;
+		}
+
+		public void doSthThatDelegatesToJavanica() {
+			this.javanicaClass.doSth();
+		}
+	}
+
+	public static class JavanicaClass {
```
adriancole: nit on classes like this.. if public, should comment why, else take off the visibility modifier.

ex. 
```java
public static class JavanicaClass { // public because X can only process public types
```

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/hystrix/JavanicaITest.java
```diff
@@ -0,0 +1,125 @@
...
+
+		AtomicReference<Span> span;
+
+		@HystrixCommand
+		public void doSth() {
```
adriancole: Assuming Sth means something?

Since JavanicaClass and the method doSth don't really mean anything, why not make this class say what it's doing?

```
 /** Catches the span associated with a hystrix command invocation. */
 static class JavanicaSpanCatcher {
    final BlockingQueue<Span> spans = new LinkedBlockingQueue<>();

    /** With sleuth integration, we expect javanica-annotated methods to have a current span associated */
    @HystrixCommand
    void invoke() {
       spans.add(TraceContextHolder.getCurrentSpan());
    }

    Span takeSpan() throws InterruptedException {
      return spans.take();
    }
  }
```

adriancole: by naming this what it does, navigation is less important as it is more clear what the fixture does.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/hystrix/JavanicaITest.java
```diff
@@ -0,0 +1,125 @@
...
+						.isEqualTo(javanicaClass.getTraceId());
+				then(span.getName())
+						.isNotEqualTo(javanicaClass.getSpanName());
+			}
+		});
+	}
+
+	@After
+	public void cleanTrace() {
+		TraceContextHolder.removeCurrentTrace();
```
adriancole: obvious to me, possibly not others that this is static state, hence needing to do this else taint other test runs. Would be nice if there was a way to operate without static state, and use that in tests.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/hystrix/JavanicaITest.java
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	private Span givenASpanInCurrentThread() {
+		Span span = this.traceManager.startSpan("existing").getSpan();
+		this.traceManager.continueSpan(span);
+		return span;
+	}
+
+	private void whenHystrixCommandGetsExecutedViaJavanica() {
+		this.javanicaDelegation.doSthThatDelegatesToJavanica();
```
adriancole: maybe mention why it is important that we only invoke via delegation in tests? Ex. I might make a test for direct and also delegated, unless we only expect wrapping. If so, maybe comment why we only expect delegated.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/hystrix/JavanicaITest.java
```diff
@@ -0,0 +1,125 @@
...
+	}
+
+	@After
+	public void cleanTrace() {
+		TraceContextHolder.removeCurrentTrace();
+	}
+
+	@DefaultTestAutoConfiguration
+	@EnableHystrix
+	@EnableAspectJAutoProxy(proxyTargetClass = true)
```
adriancole: curious: is this only needed because javanicaDelegation doesn't have a parameter of JavanicaClass? If so, maybe just use parameter-based wiring and avoid proxying as anything that enables proxies makes me wonder why it is needed. That sort of thing brings the test itself out of focus in favor of why annotations like this are needed, until the point where they are just copy-pasted and the root reason forgotten.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/hystrix/JavanicaITest.java
```diff
@@ -0,0 +1,125 @@
...
+		JavanicaITest.JavanicaITestConfiguration.class })
+public class JavanicaITest {
+
+	@Autowired JavanicaClass javanicaClass;
+	@Autowired JavanicaDelegation javanicaDelegation;
+	@Autowired TraceManager traceManager;
+
+	@Test
+	public void should_set_span_on_an_hystrix_command_annotated_method() {
+		final Span span = givenASpanInCurrentThread();
```
adriancole: mentioned offline, and understand this is a POV vs other POV thing, but let's consider this class has about 100 lines of stuff, but only one test. The test is indirected by splitting into BDD methods. This is an optimization of BDD style over actual usage. In my experience, tests that represent usage are great ways to teach people how to use the code and also where to plop edge cases into. This particular method would be imho less complex and more compelling inlined.

Intentionally took off the "this."s because they are horrendous to read imho.

```java
Span span = traceManager.startSpan("existing").getSpan();

// it is important to note why we need to invoke this..
traceManager.continueSpan(span);

// invoke a hystrix call whose only task is to catch its span
spanCatcher.invoke();

// make sure this is a child span?
assertThat(spanCatcher.takeSpan())
   .hasTraceId(span.getId())
   .hasSpanName("what we actually expect");
```

adriancole:
commented tons as marcin asked me to say what I'd prefer, not because I feel they are blocking.

LGTM

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/hystrix/JavanicaITest.java
marcingrzejszczak: +1

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/hystrix/JavanicaITest.java
marcingrzejszczak: very good idea

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/hystrix/JavanicaITest.java
marcingrzejszczak: Good point - I'll refactor the names. I'm not a fan of adding javadocs where it's not necessary.

Instead of writing

```
 /** Catches the span associated with a hystrix command invocation. */
 static class JavanicaSpanCatcher {
```

I can just name the class

```
 static class HystrixCommandInvocationSpanCatcher {
```

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/hystrix/JavanicaITest.java
marcingrzejszczak: It will be difficult since as you said the Trace logic operates on ThreadLocals and the tests share threads.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/hystrix/JavanicaITest.java
marcingrzejszczak: This delegation here was unnecessary - I've made a similar test for async where delegation made sense.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/hystrix/JavanicaITest.java
marcingrzejszczak: The annotation is not necessary - thanks for the hint. I actually thought that it would be that's why I added it.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/hystrix/JavanicaITest.java
marcingrzejszczak: I followed the convention with the `this` - I don't like them either :)

As for what you wrote - I hate comments in the code. IMO the majority of them gets outdated the moment you write them.

Why should I write

```
// make sure this is a child span?
```

if I can write

```
make_sure_this_is_a_child_span()
```

or even better

```
then(spanCatcher.takeSpan()).isAChildSpan()
```

But I guess this is a matter of personal preference :)

marcingrzejszczak:
Thanks for the review! I really liked it :)

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/hystrix/JavanicaITest.java
marcingrzejszczak: +1

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/hystrix/JavanicaITest.java
adriancole: >
> or even better
>
> then(spanCatcher.takeSpan()).isAChildSpan()
>
> my fav of the lot.

> But I guess this is a matter of personal preference :)
>
indeed

adriancole:
> Thanks for the review! I really liked it :)
>
mutually!

PR #80, opened 2015-12-29
Fixes missing zipkin service name and polishes span converters.

adriancole:
Fixes missing zipkin service name and polishes span converters.

Missing service name:

Zipkin service names were logged as null, which is invalid and led to
them showing up as "unknown" in the zipkin ui. This was due to a wiring
bug, and a special-case, which this change fixes.

The special-case was when a sleuth span had no annotations. Since zipkin
service names are attached to annotations, they are only queryable when 
annotations exist. When there are no annotations, we add the "lc" 
binary annotation, which makes that span attached to the correct service
in zipkin.

Polishing:

Zipkin timestamps were not always set as microseconds. This fixes that.

The de-facto label in zipkin for unknown service is "unknown". This
fixes the code, which formerly fell back to "application".

This also removes complexity in assigning timestamp and duration as we
no longer need to make pseudo-annotations "acquire" and "release".

Finally, this adds tests about above consistently to both scs-zipkin and
scs-zipkin-stream.

### DIFF
--- a/spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/ServerPropertiesHostLocator.java
+++ b/spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/ServerPropertiesHostLocator.java
@@ -16,7 +16,6 @@
 
 package org.springframework.cloud.sleuth.stream;
 
-import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.web.ServerProperties;
 import org.springframework.boot.context.embedded.EmbeddedServletContainerInitializedEvent;
 import org.springframework.cloud.sleuth.Span;
@@ -28,15 +27,14 @@ import org.springframework.context.event.EventListener;
  */
 public class ServerPropertiesHostLocator implements HostLocator {
 
-	@Value("${spring.application.name:application}")
-	private String appName;
-
-	private ServerProperties serverProperties;
-
+	private final ServerProperties serverProperties;
+	private final String appName;
 	private Integer port;
 
-	public ServerPropertiesHostLocator(ServerProperties serverProperties) {
+	public ServerPropertiesHostLocator(ServerProperties serverProperties,
+																		 String appName) {
 		this.serverProperties = serverProperties;
+		this.appName = appName;
 	}
 
 	@Override
@@ -80,7 +78,7 @@ public class ServerPropertiesHostLocator implements HostLocator {
 
 	private String getServiceName(Span span) {
 		String serviceName;
-		if (span.getProcessId() != null) {
+		if (span.getProcessId() != null) { // TODO: javadocs say this isn't nullable!
 			serviceName = span.getProcessId().toLowerCase();
 		}
 		else {
--- a/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ServerPropertiesEndpointLocator.java
+++ b/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ServerPropertiesEndpointLocator.java
@@ -16,7 +16,6 @@
 
 package org.springframework.cloud.sleuth.zipkin;
 
-import org.springframework.beans.factory.annotation.Value;
 import org.springframework.boot.autoconfigure.web.ServerProperties;
 import org.springframework.boot.context.embedded.EmbeddedServletContainerInitializedEvent;
 import org.springframework.cloud.util.InetUtils;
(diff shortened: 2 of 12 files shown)


adriancole:
cc @marcingrzejszczak who noticed the bug

spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/ServerPropertiesHostLocator.java
```diff
@@ -28,15 +27,14 @@
  */
 public class ServerPropertiesHostLocator implements HostLocator {
 
-	@Value("${spring.application.name:application}")
```
adriancole: this is new'ed up, so value injection never happened.

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ServerPropertiesEndpointLocator.java
```diff
@@ -30,15 +29,14 @@
  */
 public class ServerPropertiesEndpointLocator implements EndpointLocator {
 
-	@Value("${spring.application.name:application}")
```
adriancole: this is new'ed up, so value injection never happened.

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinSpanListener.java
```diff
@@ -121,12 +122,14 @@ public void release(SpanReleasedEvent event) {
...
 	 */
-	public com.twitter.zipkin.gen.Span convert(Span span) {
+	// Visible for testing
+	com.twitter.zipkin.gen.Span convert(Span span) {
 		com.twitter.zipkin.gen.Span zipkinSpan = new com.twitter.zipkin.gen.Span();
+
 		addZipkinAnnotations(zipkinSpan, span, this.localEndpoint);
-		List<BinaryAnnotation> binaryAnnotationList = createZipkinBinaryAnnotations(span,
-				this.localEndpoint);
-		zipkinSpan.setDuration(span.getEnd() - span.getBegin());
```
adriancole: wrong timeunit

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinSpanListener.java
```diff
@@ -139,35 +142,20 @@ public void release(SpanReleasedEvent event) {
...
-			}
-			else if (zipkinAnnotation.getValue().equals("release")) {
-				endTs = zipkinAnnotation.getTimestamp();
-			}
-			else {
-				zipkinSpan.addToAnnotations(zipkinAnnotation);
-			}
-		}
-		if (startTs != null) {
-			zipkinSpan.setTimestamp(startTs);
```
adriancole: wrong time unit here and below

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinSpanListener.java
```diff
@@ -176,9 +164,8 @@ else if (zipkinAnnotation.getValue().equals("release")) {
 	 *
 	 * @return list of Annotations that could be added to Zipkin Span.
 	 */
-	private List<BinaryAnnotation> createZipkinBinaryAnnotations(Span span,
-			Endpoint endpoint) {
-		List<BinaryAnnotation> l = new ArrayList<>();
```
adriancole: removed unnecessary list initializations here and elsewhere, as builders expose addX methods

spring-cloud-sleuth-zipkin/src/test/java/org/springframework/cloud/sleuth/zipkin/ZipkinSpanListenerTests.java
```diff
@@ -66,11 +69,54 @@
...
+		assertThat(result.annotations.get(0).host)
+				.isEqualTo(listener.localEndpoint);
+		assertThat(result.binary_annotations.get(0).host)
+				.isEqualTo(result.annotations.get(0).host);
+	}
+
+	/** zipkin's Endpoint.serviceName should never be null. */
+	@Test
+	public void localEndpointIncludesServiceName() {
+		assertThat(listener.localEndpoint.service_name)
```
adriancole: implicitly also checks null

spencergibb:
Nice, just cursory glance on mobile. Glad you're getting into sleuth code!

adriancole:
seems to improve things based on the e-e tests, so giving it a merge

Issue #85, opened 2015-12-31
Add some Samplers


dsyer:
We need at least a configurable fixed percentage-based sampler, and some docs on how to use it to control span data collection. It could be used  with `@RefreshScope` for instance to drive dynamic sampling based on traffic rates.

Also probably useful would be a way to force export of span data by adding a special header. That could even be the default sampler.

dsyer:
See also #56

adriancole:
Some context on zipkin and how it deals with the second part of this issue:

A sampled boolean propagation field is commonly used. This is often set
after a sampling policy applies, and propagates in HTTP as a header (to
avoid next hop on sampled=false mistakingly creating a new root span)

That's different than a debug flag, at least in zipkin, where it is a hint
made at the instrumentation point (ex the tracing api) to say make sure
this always passes. This lives in the span as opposed to propagation
context, so processed at collection time.

I think this topic came up in open tracing, too. I will update my comment
with links later.

adriancole:
On the first part (sampling algo), the dapper paper has a commonly
implemented approach, which is to hash the trace id as a part of the
collection-side sampling function. This allows us to ensure we drop an
entire trace vs parts of it (as partial traces are misleading and can be
worse than no traces).

http://research.google.com/pubs/archive/36356.pdf

IIRC, sleuth ids are UUIDs encoded as a string. We can either hash the
string or retain the 128bits behind it and use that as an input to get the
same effect. For example, htrace split the 128bits into two accessors
(upper and lower). This would allow a numeric function to be applied (note
afaik htrace don't have a collection-side sampler, which may explain why
they don't use the span id as an input to their sampling implementations)

https://github.com/apache/incubator-htrace/blob/master/htrace-core4/src/main/java/org/apache/htrace/core/SpanId.java

The zipkin-java project is closer to dapper where the trace ids are 64bit
numbers. I will add a compatible impl there, soon.

adriancole:
I took a look at how sampler was used (which is a little more complex than necessary due to #86).

Unless I'm mistaken, the sampling function applies regardless of whether a span is a parent or child. This implies that a sampler impl should take some care to avoid sparse traces. For example, we should either document or attempt to avoid losing a part of a trace because the sampler of the parent had a different answer than the child.

gauravrmazra:
htrace has CountSampler (every N calls) and ProbabilitySampler (some request based on threshold) for instance. Is it a good add-on?

[...]

PR #90, opened 2015-12-31
Issues #85 samplers

marcingrzejszczak:
Initial approach


adriancole:
Thanks for putting an initial impl together.

What tradeoffs are you looking for in this approach vs sampling methods used in other tracers impls?

I can guess, but interested in what we are trying to improve on, ex overcoming flaws in algorithms in existing ones like brave, finagle, zipkin, htrace, etc. 

With something as critical as sampling, usually I try to get someone distant from the code to ack it as well, as for example I'm not a statistician. I've used the distributed-tracing Google group, gitter, and 1-1 email to field questions like this before.

If this is just a straw man for discussion, that's an answer, too. If so, I'd say next step is to get an expert opinion.

marcingrzejszczak:
Like presented in the message - this is an ultra-initial approach to start a discussion like the one that we're having. Maybe I should have written *DO NOT MERGE* to ensure that this is not ready to merge at all.

The algorithm in this sampler is from the statistics point of view terrible - there is no proper distribution or anything like that. I've done this deliberately to begin with sth.

I had no knowledge of implementations of such algorithms in brave for instance. If you have any knowledge about such concrete implementation please paste a link or sth. 

Definitely will ping those groups - thanks for the hints! That's why I put this PR in the first place.

marcingrzejszczak:
We had a long Sunday chat with @adriancole and came to the following conclusions.

1) We have to take TraceID into consideration during sampling to be consistent.
2) We treat the traceID as a 128 bit number - like HTrace does - https://github.com/apache/incubator-htrace/blob/master/htrace-core4/src/main/java/org/apache/htrace/core/SpanId.java
3) We apply the Finagle algorithm to define whether we should sample or not - https://github.com/openzipkin/zipkin-java/pull/55/files#diff-31b830705fabb2e936c73337dfa5f219R83

adriancole:
on 3) TL;DR drop the salt from finagle's algo when using in sleuth

Finagle makes a sampling decision once, usually at the root of a trace. That decision is propagated by a sampling flag to all downstream spans. Because the result is remembered, there's no consistency problem.

The sampler in sleuth is invoked for every span in the trace, not just the root span. If we add a salt, we will end up with an inconsistent sample function. Tests should be able to prove this.

marcingrzejszczak:
@adriancole - pls check if now it looks better. Take a look especially at the bits manipulation (I have working with bits directly ;) ).

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSampler.java
```diff
@@ -0,0 +1,57 @@
...
+		}
+		String traceId = currentSpan.getTraceId();
+		UUID traceUuid = UUID.fromString(traceId);
+		BigInteger traceAsInt = UuidToBigIntegerConverter.uuidToBigInt(traceUuid);
+		BigInteger mod = traceAsInt.mod(MOD_DIVISOR);
+		return mod.compareTo(percentageAsBigInteger()) <= 0;
+	}
+
+	private BigInteger percentageAsBigInteger() {
+		double percents = samplerConfiguration.getPercentage() * 100;
```
adriancole: If you use the factory method, BigInteger.valueOf I suspect this would be less dancing

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSampler.java
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	@Override
+	public boolean next(Void info) {
+		Span currentSpan = traceAccessor.getCurrentSpan();
+		if (currentSpan == null) {
+			return false;
+		}
+		String traceId = currentSpan.getTraceId();
+		UUID traceUuid = UUID.fromString(traceId);
```
adriancole: non-blocking, but I'd probably fold these steps (as you may not always want to use the UUID type). Ex. TraceIdToBigIntegerConverter can mention that it is UUID formatted instead. This keeps the math together, but separate from id decoding

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSampler.java
adriancole: I don't think we need the precision of a double to represent a sample percent. I'd prefer float (even if this wasn't introduced here)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSampler.java
```diff
@@ -0,0 +1,57 @@
...
+		BigInteger traceAsInt = UuidToBigIntegerConverter.uuidToBigInt(traceUuid);
+		BigInteger mod = traceAsInt.mod(MOD_DIVISOR);
+		return mod.compareTo(percentageAsBigInteger()) <= 0;
+	}
+
+	private BigInteger percentageAsBigInteger() {
+		double percents = samplerConfiguration.getPercentage() * 100;
+		String percentsAsString = Integer.toString((int) percents);
+		return new BigInteger(percentsAsString);
```
adriancole: should probably mention that the precision is whole-numbers percents only. For example, in finagle, the trace rate is 0.1% which this wouldn't support.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/SamplerConfiguration.java
```diff
@@ -0,0 +1,21 @@
...
+
+@ConfigurationProperties("spring.sleuth.sampler")
+@Data
+public class SamplerConfiguration {
+
+	/**
+	 * Percentage of requests that should be sampled.
+	 * E.g.
+	 * <ul>
+	 * <li> 1.0 - 100% requests should be sampled </li>
```
adriancole: nice docs!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/SamplerConfiguration.java
```diff
@@ -0,0 +1,21 @@
...
+	 * <li> 0.8 - 80% of requests should be sampled </li>
+	 * <li> 0.0 - 0% requests should be sampled </li>
+	 * </ul?
+	 */
+	private double percentage = 0.1;
```
adriancole: nit-repeat 0.1f and switch to float as double is distracting

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/UuidToBigIntegerConverter.java
```diff
@@ -0,0 +1,23 @@
...
+class UuidToBigIntegerConverter {
+
+	/**
+	 * Takes two sets of bits (most and least significant ones), puts them together
+	 * and represents as a BigInteger
+	 *
+	 * @param uuid - UUID to convert
+	 * @return - BigInteger representation of the UUID
+	 */
+	static BigInteger uuidToBigInt(UUID uuid) {
```
adriancole: said earlier, I'd love this to be the decoder function from String -> BigInteger. The UUID type is a convenience for that conversion.

adriancole: reason is that you can much more easily optimize this part, if it is String -> BigInteger.. Also you can address malformed UUIDs

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSamplerTest.java
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	@Test
+	public void should_pass_given_percent_of_samples() throws Exception {
+		samplerConfiguration.setPercentage(0.3d);
+
+		int numberOfSampledElements = countNumberOfSampledElements();
+
+		//TODO: We have random data thus tests can fail due to this percentage. If that's the case we should change it to .isPositive()
+		then(numberOfSampledElements).isCloseTo(30, withPercentage(30));
```
adriancole: 30% is a hefty error rate... did you actually need that? FWIW I found 3% to be the sweet spot

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSamplerTest.java
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	private int countNumberOfSampledElements() {
+		int passedCounter = 0;
+		for (int i = 0; i < 100; i++) {
+			boolean passed = new PercentageBasedSampler(samplerConfiguration, traceReturningSpanWithUuid()).next(null);
```
adriancole: note to self or willing others: add lambda support for test tree!

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSamplerTest.java
```diff
@@ -0,0 +1,73 @@
...
+public class PercentageBasedSamplerTest {
+
+	SamplerConfiguration samplerConfiguration = new SamplerConfiguration();
+	TraceAccessor traceAccessor = traceReturningSpanWithUuid();
+
+	@Test
+	public void should_pass_all_samples_when_config_has_1_percentage() throws Exception {
+		samplerConfiguration.setPercentage(1d);
+
+		boolean passed = new PercentageBasedSampler(samplerConfiguration, traceAccessor).next(null);
```
adriancole: I bet you love next(null) right? :P

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSamplerTest.java
marcingrzejszczak: When I executed a very big number of samples I got a 75% error rate ;)

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSamplerTest.java
marcingrzejszczak: or write tests in Groovy ;)

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSamplerTest.java
marcingrzejszczak: Of course I do :) who doesn't ;)

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/sampler/UuidToBigIntegerConverterTest.java
```diff
@@ -0,0 +1,52 @@
...
+		UUID uuid = new UUID(testData.mostSignificantBits, testData.leastSignificantBits);
+
+		BigInteger bigInteger = UuidToBigIntegerConverter.uuidToBigInt(uuid);
+
+		then(bigInteger).isEqualByComparingTo(testData.expectedInteger);
+	}
+
+	public Object[] parametersForShould_convert_uuid_to_big_integer() {
+		return $(
+				sample(Long.MAX_VALUE, Long.MAX_VALUE, TWO.pow(126).subtract(ONE)),
```
adriancole: style + tiredness, so feel free to say "I'll pass"..

because I'm tired I can't remember why 2^126-1... to relate the left side to the right side it might be clearer to use a byte buffer.  ex. this is longer, but doesn't require the reader to think through the math..

```java
new BigInteger(
  ByteBuffer.allocate(128).putLong(Long.MAX_VALUE).putLong(Long.MAX_VALUE).array()
)
```
Above syntax might be wrong.. even easier with Okio buffer as you don't need to pre-allocate

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSampler.java
marcingrzejszczak: good idea - I'll move rename the class and move all the UUID logic there.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/UuidToBigIntegerConverter.java
marcingrzejszczak: +1

marcingrzejszczak:
@adriancole - more changes arrived... back to you then :)

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/sampler/UuidToBigIntegerConverterTest.java
marcingrzejszczak: As the Long's MAX_VALUE javadoc says

```
/**
     * A constant holding the maximum value a {@code long} can
     * have, 2<sup>63</sup>-1.
     */
    @Native public static final long MAX_VALUE = 0x7fffffffffffffffL;
```

So we get

1...63x1...64x1

the first 1 is a sign bit.
the 127 bits are the remaining bits thus the greatest value possible is
2^127-1.

Per analogy if we have 3 bits then the greatest is 111 which is 8.

We have 128 bits where there is also a bit sign. So 127 bits. And in the
Long's Max value we have the additional -1. Alltogether 2^127 - 1.

I also tried your approach but first it should be:

new BigInteger(
  ByteBuffer.allocate(16).putLong(Long.MAX_VALUE).putLong(Long.MAX_VALUE).array()
)

Not 128 cause it's in bytes not bits, but still it's not working as it
should ;)

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/sampler/UuidToBigIntegerConverterTest.java
```diff
@@ -0,0 +1,52 @@
...
+import static java.math.BigInteger.ONE;
+import static java.math.BigInteger.ZERO;
+import static junitparams.JUnitParamsRunner.$;
+import static org.assertj.core.api.BDDAssertions.then;
+import static org.springframework.cloud.sleuth.sampler.UuidToBigIntegerConverterTest.TestData.sample;
+
+@RunWith(JUnitParamsRunner.class)
+public class UuidToBigIntegerConverterTest {
+
+	private static final BigInteger TWO = new BigInteger("2");
```
adriancole: food for thought..

junit is always new instance for each test. If you run with the normal `@RunWith(Parameterized.class)`, you get what might end up as less clutter.

```
@Parameterized.Parameter(0)
public long mostSignificantBits;
@Parameterized.Parameter(1)
public long leastSignificantBits;
...
```

IOTW, you don't need to create a TestData type

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/sampler/UuidToBigIntegerConverterTest.java
```diff
@@ -0,0 +1,52 @@
...
+	@Parameters
+	public void should_convert_uuid_to_big_integer(TestData testData) throws Exception {
+		UUID uuid = new UUID(testData.mostSignificantBits, testData.leastSignificantBits);
+
+		BigInteger bigInteger = UuidToBigIntegerConverter.uuidToBigInt(uuid);
+
+		then(bigInteger).isEqualByComparingTo(testData.expectedInteger);
+	}
+
+	public Object[] parametersForShould_convert_uuid_to_big_integer() {
```
adriancole: also style, but to me `$(` is weirder than just initializing the array.

```java
@Parameters
  public static Object[][] createData() {
    return new Object[][]{
        {Long.MAX_VALUE, Long.MAX_VALUE, expectedThing}.
        {0L, 0L, ZERO},
        ...
```

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/sampler/UuidToBigIntegerConverterTest.java
adriancole: Heh the length of that explanation implies we need a simpler input. One
shouldn't need to figure so many things out to know if the inputs are right
or not.

Food for thought:

If the goal is to prove we can cover all 128 bits, how about initialize the
longs as 0xFFFF.... Then on the RHS, use the array ctor of BigInteger with
an array filled by Arrays.fill((byte)0xFF) or more explicitly an array
literal made of 0xFFs

Even if you alternate one can visually compare left with right without any
math. This means less likely one will just assume the inputs and outputs
are correct out of laziness. This could help prevent bugs from sticking
around.

Remember, at the end of the day, the bigint is just a convenient holder of
128 random bits for the purpose of comparing them as a number.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSampler.java
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	@Override
+	public boolean next(Void info) {
+		Span currentSpan = traceAccessor.getCurrentSpan();
+		if (currentSpan == null) {
+			return false;
+		}
+		BigInteger traceAsInt = TraceIdToBigIntegerConverter.traceIdToBigInt(currentSpan.getTraceId());
+		BigInteger mod = traceAsInt.mod(MOD_DIVISOR);
```
adriancole: much nicer!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/TraceIdToBigIntegerConverter.java
```diff
@@ -0,0 +1,24 @@
...
+class TraceIdToBigIntegerConverter {
+
+	/**
+	 * Takes two sets of bits (most and least significant ones), puts them together
+	 * and represents as a BigInteger
+	 *
+	 * @param traceIdToConvert
+	 * @return - BigInteger representation of the TraceId
+	 */
+	static BigInteger traceIdToBigInt(String traceIdToConvert) {
```
adriancole: I've an idea about how to eliminate this btw.., instead of using BigInteger, just compare 2 longs (or 4 ints)

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSamplerTest.java
adriancole: sounds like something to revisit, because if the error rate is this high, we are missing something.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSamplerTest.java
adriancole: :cry:

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/TraceIdToBigIntegerConverter.java
marcingrzejszczak: Maybe let's leave it for the time being? If it becomes a performance problem issue then we'll fix it?

adriancole: agreed!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/DefaultStringToUuidConverter.java
```diff
@@ -0,0 +1,44 @@
...
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Default implementation that converts String into UUID
+ * On parse exceptions an exception is thrown and logged.
+ *
+ * Successful and failed parse attempts are registered in counters.
```
adriancole: remove this line

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/DefaultStringToUuidConverter.java
```diff
@@ -0,0 +1,44 @@
+package org.springframework.cloud.sleuth.sampler;
+
+import lombok.extern.slf4j.Slf4j;
+
+import java.util.UUID;
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Default implementation that converts String into UUID
+ * On parse exceptions an exception is thrown and logged.
```
adriancole: update this to parse exceptions coerce to null.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/DefaultStringToUuidConverter.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * Successful and failed parse attempts are registered in counters.
+ *
+ * @author Marcin Grzejszczak
+ */
+@Slf4j
+public class DefaultStringToUuidConverter implements StringToUuidConverter {
+
+	@Override
+	public UUID convert(String source) {
```
adriancole: If we don't have a Nullable annotation in the classpath, I'd add..

```java
/** Returns a UUID parsed from the input, or null if failed for any reason. */
```

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSampler.java
```diff
@@ -0,0 +1,47 @@
...
+
+/**
+ * Sampler that based on the given percentage rate will allow sampling.
+ *
+ * A couple of assumptions have to take place in order for the algorithm to work properly:
+ *
+ * <ul>
+ *     <li>We're taking the TraceID into consideration for sampling to be consistent</li>
+ *     <li>We treat the TraceID as a UUID - a 128 bit number - like HTrace does - https://github.com/apache/incubator-htrace/blob/master/htrace-core4/src/main/java/org/apache/htrace/core/SpanId.java</li>
+ *     <li>We apply the Finagle algorithm to define whether we should sample or not (we're doing a % 100 not greater) - https://github.com/openzipkin/zipkin-java/pull/55/files#diff-31b830705fabb2e936c73337dfa5f219R83</li>
```
adriancole: Change reference to zipkin algorithm https://github.com/openzipkin/zipkin-java/blob/master/zipkin-java-core/src/main/java/io/zipkin/TraceIdSampler.java#L78

(as this isn't the finagle one anymore)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSampler.java
```diff
@@ -0,0 +1,47 @@
...
+import org.springframework.cloud.sleuth.TraceAccessor;
+
+/**
+ * Sampler that based on the given percentage rate will allow sampling.
+ *
+ * A couple of assumptions have to take place in order for the algorithm to work properly:
+ *
+ * <ul>
+ *     <li>We're taking the TraceID into consideration for sampling to be consistent</li>
+ *     <li>We treat the TraceID as a UUID - a 128 bit number - like HTrace does - https://github.com/apache/incubator-htrace/blob/master/htrace-core4/src/main/java/org/apache/htrace/core/SpanId.java</li>
```
adriancole: drop HTrace ref as we don't treat it like they do

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/UuidTraceIdToThresholdComparable.java
```diff
@@ -0,0 +1,54 @@
...
+final class UuidTraceIdToThresholdComparable implements Comparable<String> {
+
+	private static final int EIGHT_BYTES = 64;
+	private static final int GREATER_THAN_THRESHOLD = 1;
+	/**
+	 * We're shifting the first 64 bits by 8 bytes (64 bits) to the left cause these are
+	 * the most significant bits. Then we're adding -1 (1111...1111). That's how we're getting
+	 * 1111....1111 (128 times '1')
+	 */
+	private static final BigInteger MAX_128 =
```
adriancole: this actually doesn't end up as you mention above..

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/DefaultStringToUuidConverter.java
marcingrzejszczak: We don't have it :/

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/UuidTraceIdToThresholdComparable.java
adriancole: I'd replace the whole thing with this, as I dumped the bytes and found the above isn't accurate.

Ex. here's a test, which I'm pretty sure will fail

```java
    // See that it is 127 bits
    assertThat(UUIDSampler.MAX_128.toString(2))
        .hasSize(127);
```

Here's the code
```java
    static final BigInteger MAX_128 = max_128signed();

    /**
     * The Long.MAX_VALUE in binary 0 followed by 63 1s.
     *
     * We simulate a 128bit long, by doing the same, except following by 127 1s
     */
    static BigInteger max_128signed() {
      byte[] max_128signed = new byte[16];
      Arrays.fill(max_128signed, (byte) -1); // initialize to 11111111
      max_128signed[0] = (byte) 127; // reset MSBs to 01111111
      return new BigInteger(max_128signed);
    }
```

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/UuidTraceIdToThresholdComparable.java
```diff
@@ -0,0 +1,54 @@
...
+	 * @param traceId - the UUID version of Trace Id
+	 */
+	@Override
+	public int compareTo(String traceId) {
+		UUID uuid = stringToUuidConverter.convert(traceId);
+		if (uuid == null) {
+			return GREATER_THAN_THRESHOLD;
+		}
+		BigInteger asInteger = BigInteger.valueOf(uuid.getMostSignificantBits())
+						.shiftLeft(EIGHT_BYTES)
```
adriancole: since we are talking about 128 bits, I'd use bits as the measurement, not bytes

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSamplerTest.java
adriancole: can you trim this to 3% now?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/UuidTraceIdToThresholdComparable.java
marcingrzejszczak: Actually the javadoc should be

```
/**
	 * We're shifting the first 64 bits by 8 bytes (64 bits) to the left cause these are
	 * the most significant bits. Then we're adding -1 (1111...1111). That's how we're getting
	 * 0111....1111 ('0' and then 127 times '1')
	 */
```

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSampler.java
```diff
@@ -0,0 +1,46 @@
...
+import org.springframework.cloud.sleuth.TraceAccessor;
+
+/**
+ * Sampler that based on the given percentage rate will allow sampling.
+ *
+ * A couple of assumptions have to take place in order for the algorithm to work properly:
+ *
+ * <ul>
+ *     <li>We're taking the TraceID into consideration for sampling to be consistent</li>
+ *     <li>We apply the Zipkin algorithm to define whether we should sample or not (we're doing a mod 100) - https://github.com/openzipkin/zipkin-java/blob/master/zipkin-java-core/src/main/java/io/zipkin/TraceIdSampler.java#L78</li>
```
adriancole: sorry to say.. doc's wrong.. not doing mod anymore.. "we're comparing against a threshold"

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/UuidTraceIdToThresholdComparable.java
```diff
@@ -0,0 +1,65 @@
...
+
+import java.math.BigInteger;
+import java.util.Arrays;
+import java.util.UUID;
+
+/**
+ * Given the absolute value of a random 128 bit trace id, we expect inputs to be balanced across
+ * 0-MAX. Threshold is the range of inputs between 0-MAX that we retain.
+ *
+ * Can be compared with a String UUID TraceId to see its relation to the provided threshold.
```
adriancole: I didn't quite grok this sentence.. maybe something like..

This decodes a trace id in UUID format into a 128 bit number, then compares it against a threshold.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSamplerTest.java
```diff
@@ -0,0 +1,94 @@
...
+
+		boolean passed = new PercentageBasedSampler(samplerConfiguration, traceReturningSpanWithInvalidUuid(), stringToUuidConverter).next(null);
+
+		then(passed).isFalse();
+	}
+
+	@Test
+	public void should_pass_given_percent_of_samples() throws Exception {
+		int numberOfIterations = 10000;
+		float percentage = 0.3f;
```
adriancole: yay

adriancole:
+100

adriancole:
This is really great work @marcingrzejszczak

marcingrzejszczak:
Merged here - https://github.com/spring-cloud/spring-cloud-sleuth/commit/640ad4c6f79103da81cd959c1850bc52553119e0

Thanks @adriancole for the reviews!

adriancole:
my pleasure

Issue #82, opened 2015-12-30
Pare down what TraceFilter is logging


adriancole:
The primary use case of distributed tracing is latency analysis, so we probably can refine this quite a bit. Moreover, observability systems need to do least harm. Logging introduces overhead and storage burden that may impact the performance or usability of the system. This is why most distributed tracing systems log the least data required.

I don't think `o.s.c.s.instrument.web.TraceFilter` logs least data required: we literally log all headers. This includes useless data and data we should be careful with, such as cookies. For example, the accept-charset header alone is often a few times larger than a competing system's complete trace.

I suggest we invert the policy. Instead of logging the most data, log the least needed for analysis. Make things opt-in. This will get us far more mileage out of our storage systems, and make our instrumentation code do more with less cycles and memory.

Here are some examples of data minimalism in practice, in other tracing libraries.

#### Finagle
Finagle is behind twitter. Due to the volume of requests, they are very conservative with how much they trace. They only add the http method (as span name), and "http.uri" (stripped of parameters). Layers below do log other data, such as "srv/finagle.version", and "finagle.retry", but for http, only 2 things.

#### Brave
Brave does roughly the same thing as finagle, adding "http.responsecode".

#### Naver Pinpoint
Pinpoint logs library-specific events such as connect/disconnected, and up to several attributes. For example, they will record 64-512 characters of query parameters. Most clients will only record the URI and if cookie storage is set, the cookie.

#### Groupon
Groupon mentioned they cherry-pick a few fields for classification purposes (on root spans)
* User agent 
* X-Forwarded-For info 
* Specialized data fields (browser identifying cookie, user identifying cookie) 
* Any A/B style experiment data they are bucketed into

adriancole:
started a discussion topic on core http tags: https://groups.google.com/forum/#!topic/distributed-tracing/dnzo02hO8KY

adriancole:
I think I've got all the answers needed on this. I'll summarize a recommendation tomorrow

PR #93, opened 2016-01-04
Pare down what TraceFilter is logging #82

gauravrmazra:


adriancole:
Adrian's 2p is to not introduce more types.

I appreciate your putting effort into this, Gaurav. Instrumentation's first goal is to do no harm. That implies that we want to avoid adding logic and new types as logic = overhead.. it is usually also complexity. I'm already concerned about complexity in this type, which you can get a rough gauge of by counting the max indent level.

By pare-down, I actually meant to be explicit about what are the critical keys we expect to log, and be quite conservative. I was imagining a strict vs variable set, as done in other tracing libraries.

A *custom* decorator or such could be added by those who need more headers or whatnot. If the custom things become common enough to be main code, we'd solve that, but no sooner.

For example, you'll notice addRequestAnnotations is already protected.. IOTW, those wanting to add more logic or concepts such as custom filters could do that in a subtype after we've pared down defaults.

gauravrmazra:
@adriancole  If you see, I limits the default request header tracing to four headers i.e. Accept, Origin, Host and User-Agent (strict which we logs) and variable header tracing which could be set by the Application User by setting spring.sleuth.instrument.web.paramPattern and comma seperated request param tracing (variable user-define i.e. what he wants to be put in annotations).

By using this, Approach to what to put in annotations or not is bit simplified. I know, User can't forcefully add annotations from request headers. But consider the case following case.

If strict header params set is introduced, but user want his custom headers to be part of span. How would that be possible? He may need to do in his logic. But with above approach he defines what request headers he want to be in span annotations which will be handled by default.

adriancole:
@gauravrmazra one challenge here is that you are assuming you know what custom behavior users will want. This is a slippery slope. ex. we start with a filter for headers, then query, then body etc.

There are some other concerns here, for example the accept header isn't very valuable in latency troubleshooting. I'd prefer to re-use the experience of other projects and be very picky. remember every property has a direct relationship with storage requirements.

gauravrmazra:
Default set of header that need be in annotation could be decided. I assumed (:)) that these four could be helpful for tracing. My example could lead to impression that the design is derived on assuming what behavior user wants :( but Introduction of Filter was not only due to storing custom headers for TraceFilter, but other filter applied on RestTemplate tracing etc. :D

adriancole:
One of the beauties of latency analysis is that it has been going on for a decade, and at least the last several years a lot of work in open source. For this reason, I don't think we need to guess to much to find the core latency-sensitive properties of an http request or response. Citing prior works is a great way to but weight behind choices like this.

With regards to code design I can see that your aim is to customize the headers accepted. My point is that we are adding code that only captures one aspect of an http request. This is the slippery slope. If we are to open new hooks, we owe it to ourselves to make the code count. Each special-hook adds weight to the project, particularly when such hooks and properties that drive them are untested and undocumented. There's no shame in providing an alternate class until a tenable design surfaces.

`protected addRequestAnnotations(HttpServletRequest request)` is already a hook, which would allow all sorts of customization to occur, not just headers. Can you try using this for now?

adriancole:
ps I'm going to keep quiet now.. and let other project owners speak up

marcingrzejszczak:
Hi @gauravrmazra ! Thanks for your imput. We really appreciate it.

Before I go into code review - could you please say what problem are you trying to solve? What is the use-case of this feature?

gauravrmazra:
The above PR was in response for #82 

My approach to this is to simplifying and add functionality as well to it. I introduced two things. By default only few headers (Accept, Origin, Host and User-Agent) are added (strict mode) in span but also given user a full freedom to configure which headers he may want to add to annotation (variable mode). The simple use-case is some custom header which he want later for tracing purposes in these traces/span and also with the log he is collecting somewhere (RELK stack maybe :)).

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebAutoConfiguration.java
```diff
@@ -68,9 +76,35 @@ public TraceWebAspect traceWebAspect() {
...
+			}
+		}
+		else {
+			params.add("accept");
+			params.add("host");
+			params.add("origin");
+			params.add("user-agent");
+		}
+		
+		ParamFilter filter = new ParamFilter();
```
gauravrmazra: Forget to add params :(

marcingrzejszczak:
Ok I read the issue and read your code. Actually I fully agree with @adriancole . If we also change the visibility of `addResponseAnnotations` to protected then everyone can write his own implementation of the filter and do whatever they want. 

From our side we would have to tweak the `TraceFilter` class as @adriancole suggested. Let it annotate less. If you run the brewery sample you can see that we're tracing gigantic amount of data. So just let's restrict it and that's pretty much it.

WDYT? @gauravrmazra @adriancole ?

gauravrmazra:
I know it traces a lot of information as annotation for web module. I am using almost each component (except hysterix, feign) in POC with Zipkin. My thinking is to start tracing in any application user only have to plugin the library and enable features in config file. Why user is bothered to provide implementation for core components of tracing. 

For now, I can agree with your approach. I thought of following changes required.
-> In TraceWebAutoConfiguration make TraceFilter as ConditionalBean and then modify TraceFilter to add few things in annotation. 
Isn't it? @adriancole  @marcingrzejszczak

marcingrzejszczak:
Just to be sure we're on the same track... 

For sure this: https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebAutoConfiguration.java#L68 can't be ConditionalOnMissingBean since it's a filter registration bean. If I understand you correctly you want to change that bean to accept TraceFilter as a bean? And that TraceFilter bean would be ConditionalOnMissingBean?

gauravrmazra:
@marcingrzejszczak  Exactly :)

@adriancole @marcingrzejszczak Please keep this PR open for time.

adriancole:
PS when we add core annotations, we would surely document why they are
critical for latency debugging. Want to make a case for headers you find
critical? Perhaps some of your choices will match others.

I recall you added cookie, for example, which while tricky for security has
a use case, as often user or device identity is encoded there. What do you
use this for (ex query in UI?)? What production latency problem did this
help you solve? What would the documentation about it advise? (in your
opinion)

For example, when discussing jdbc query, one user mentioned they use it to
ad-hoc test the latent span.. Literally copy paste into a SQL prompt. It is
stories like this that make the difference between choices and use cases,
if that makes sense.

adriancole:
Makes sense to make overriding the impl more effective. Guarav, how about
changing this PR to do that (or raising another?)

I will take the paring down issue, circulating ideas in the OpenTracing
group first.

Bear in mind, the general direction of the tracing working group is to make
portable tools. It is this that is driving the aim for compatible core
annotations. So we are looking for the least annotations needed to
accomplish that, as for example policy by HTTP method, URI or status (which
aren't headers) should be portable. That leads to portable visualizations
which can understand the semantics of HTTP.

If you are interested in participating, hop on OpenZipkin or OpenTracing on
gitter, or post to the distributed-tracing google group. (Obv can also
watch issues in OpenTracing repos).

PR #94, opened 2016-01-05
Updates to latest zipkin and brave, removing Scribe dependency

adriancole:
Recent versions of zipkin and brave operate with http instead of scribe.
This changes the implementation accordingly, as well adds a sampler to
the POST endpoint.

A notable impact is that we no longer require a collector process, as
the zipkin server's POST endpoint is a collector.

### DIFF
--- a/docs/src/main/asciidoc/README.adoc
+++ b/docs/src/main/asciidoc/README.adoc
@@ -25,7 +25,7 @@ The Ribbon sample makes an interesting demo or playground for learning about zip
 === Running samples with Zipkin
 
 1. Optionally run the https://github.com/openzipkin/zipkin[Zipkin] UI, e.g. via docker compose (there's a `docker-compose.yml` in https://github.com/spring-cloud/spring-cloud-sleuth-samples/spring-cloud-sleuth-sample-zipkin[Spring Cloud Sleuth], or in https://github.com/openzipkin/docker-zipkin[Docker Zipkin]
-7. Run the zipkin sample application (set `sample.zipkin.enabled=false` if you have no Zipkin running). If you are using a VM to run docker you might need to tunnel port 9410 to localhost, or change the `spring.zipkin.host`.
+7. Run the zipkin sample application (set `sample.zipkin.enabled=false` if you have no Zipkin running). If you are using a VM to run docker you might need to tunnel port 9411 to localhost, or change the `spring.zipkin.host`.
 8. Hit `http://localhost:3380`, `http://localhost:3380/call`, `http://localhost:3380/async` for some interesting sample traces (the app callas back to itself).
 9. Goto `http://localhost:8080` for zipkin web (if you are using boot2docker the host will be different)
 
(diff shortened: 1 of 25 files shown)


adriancole:
@dsyer @spencergibb @marcingrzejszczak this is a spike for feedback, particularly around where the collection sample rate should live. I'm going to let this sit a day before working on it again.

spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/ZipkinMessageListener.java
```diff
@@ -196,6 +191,13 @@ public Cloud cloud() {
 			return new CloudFactory().getCloud();
 		}
 
+		// TODO: what config shoud we use? zipkin uses zipkin.collector.sample-rate
+		@Bean
+		TraceIdSampler traceIdSampler() {
```
adriancole: @joshlong @dsyer so we need a config key for the collector's sample rate (ex. 0.7 means keep 70%). What should that be? Once we've decided, I can finish this up

marcingrzejszczak: How does this correlate with https://github.com/spring-cloud/spring-cloud-sleuth/pull/90/files ?

adriancole: So there's 2 tiers of sampling: instrumentation and collection

#90 is a number the app controls (affects instrumentation policy).We'd expect folks to have different values depending on throughput of their apps and probably its canary-ness.

This PR is a global rate, meaning that of all the spans sent from instrumentation, how many do we retain? This defaults to 100%, but you could lower this automatically or manually if, for example the storage system is degraded.

Make sense?

adriancole:
I owe a test on `SamplingZipkinSpanIterator` if you need to merge, go for it and I'll add separately

adriancole:
ok test in (fixed a bug in my impl that it found!)

docs/src/main/asciidoc/README.adoc
```diff
@@ -37,7 +37,7 @@ image::https://raw.githubusercontent.com/spring-cloud/spring-cloud-sleuth/master
 
 === Running samples with Zipkin Stream
 
-Instead of using the native (Scala) Zipkin collector server you can export your span data over https://raw.githubusercontent.com/spring-cloud/spring-cloud-stream[Spring Cloud Stream].
+Instead of using the native (Scala) Zipkin server server you can export your span data over https://raw.githubusercontent.com/spring-cloud/spring-cloud-stream[Spring Cloud Stream].
```
shakuzen: server server

adriancole: thanks! fixed

adriancole:
@marcingrzejszczak heads up! gonna take your offer on the brewery adjustments

PR #95, opened 2016-01-06
Paredown TraceFilter headers #82

gauravrmazra:


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/HeaderKeys.java
```diff
@@ -0,0 +1,9 @@
+package org.springframework.cloud.sleuth.instrument.web;
+
+//Annotation key holding class for request and response headers
+abstract class HeaderKeys {
```
adriancole: bikeshed, but I usually go for final class with private ctor for constant holders.

adriancole: HeaderKeys is a confusing name, especially in http, because a lot of these aren't headers. Moreover header isn't a conceptual name in tracing either.. maybe HttpKeys or something?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
```diff
@@ -98,14 +94,13 @@ protected void doFilterInternal(HttpServletRequest request,
...
-		}
-		else if (skip) {
+		} else if (skip) {
 			addToResponseIfNotPresent(response, Trace.NOT_SAMPLED_NAME, "");
 		}
 
 		String spanId = getHeader(request, response, Trace.SPAN_ID_NAME);
 		String traceId = getHeader(request, response, Trace.TRACE_ID_NAME);
-		String name = "http" + uri;
+		String name = HeaderKeys.HTTP + request.getMethod();
```
adriancole: this looks like a behavior change? make sure you call things out like this so people don't miss them. Ex. often I'll self-comment on my files.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
```diff
@@ -188,55 +177,31 @@ private void publish(ApplicationEvent event) {
...
-			String name = headerNames.nextElement();
-			Enumeration<String> values = request.getHeaders(name);
-			while (values.hasMoreElements()) {
-				String value = values.nextElement();
-				String key = "/http/request/headers/" + name.toLowerCase();
-				this.traceManager.addAnnotation(key, value);
 
-			}
-		}
+		this.traceManager.addAnnotation(HeaderKeys.HTTP_URL, request.getRequestURL().toString());
```
adriancole: note: if you didn't extract the constant class, it would be easier to tell what is a change and what isn't. As it is, we need to scroll up and down (and look at formatting changes) to tell if something changed or not.

adriancole:
I appreciate that you are trying to get things done, but the combination of behaviors, formatting and extraction is making this laborious to review. Maybe just focus on the extensibility aspect (as requested last time?)

As also mentioned before, I'd like to review the namespaces with a larger group before committing a change to the http annotation namespace to avoid thrashing.

Issue #89, opened 2015-12-31
Migrate integration tests from brewery to sleuth


marcingrzejszczak:
we could from maven start for example Zipkin docker container and run some tests against it

adriancole:
Yep. Also the zipkin-java server should be something we can make a junit
rule out of. This is something I think can solve the classpath concerns,
too, as I probably wouldn't involve spring framework code in a junit rule.

marcingrzejszczak:
Done - https://github.com/spring-cloud/spring-cloud-sleuth/commit/4efa4c3f27cc74ad115134554c5c3d476191361f

PR #97, opened 2016-01-07
[#89] Initial approach to the integration tests

marcingrzejszczak:
Using docker-compose in the `pre-integration-test` phase. In that phase it's also booting the app from the jar (we're checking that it's package properly). 

Then in the `integration-test` phase it's executing 2 tests. 

Finally in the `post-integration-test` phase (that's always executed regardless of the `integration-test` phase status) it's killing docker containers, and killing the started apps. Also it's printing 500 lines of logs from the started app (for debugging).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,8 +15,8 @@ install:
 - ./mvnw install -P docs -q -U -DskipTests=true -Dmaven.test.redirectTestOutputToFile=true
 - 'if [ "${TRAVIS_PULL_REQUEST}" = "false" ] && [ "${TRAVIS_BRANCH}" = "master" ]; then ./docs/src/main/asciidoc/ghpages.sh; fi'
 script:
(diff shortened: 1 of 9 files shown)


marcingrzejszczak:
CC @dsyer , @spencergibb , @adriancole

.travis.yml
```diff
@@ -12,7 +12,7 @@ before_install:
   - echo "https://$GH_TOKEN:@github.com" > .git/credentials
   - gem install asciidoctor
 install:
-- ./mvnw install -P docs -q -U -DskipTests=true -Dmaven.test.redirectTestOutputToFile=true
+- ./mvnw install -P docs -q -U -DskipTests=true -Dexec.skip=true -Dmaven.test.redirectTestOutputToFile=true
```
adriancole: add a comment above about exec.skip? it is a rarely used option

adriancole: ex `# don't run integration tests by default`

marcingrzejszczak: I would but ATM it's not working :) I officialy hate Maven

adriancole:
summary of thoughts offline. would be great if the lifecycle of the components could be managed by junit somehow. they'd be easier to run I think

Issue #84, opened 2015-12-31
Thrift exception - No name set in Span


marcingrzejszczak:
Occasionally the following exception is present in the logs and the entry is not there in Zipkin

```
presenting_1 | 2015-12-31 08:26:24.780  WARN 1 --- [pool-7-thread-1] c.g.k.brave.scribe.ScribeClientProvider  : Thrift exception.
presenting_1 | 
presenting_1 | org.apache.thrift.TApplicationException: Internal error processing Log: 'com.twitter.zipkin.common.IncompleteTraceDataException: No name set in Span'
presenting_1 | 	at org.apache.thrift.TApplicationException.read(TApplicationException.java:108) ~[libthrift-0.9.0.jar!/:0.9.0]
presenting_1 | 	at org.apache.thrift.TServiceClient.receiveBase(TServiceClient.java:71) ~[libthrift-0.9.0.jar!/:0.9.0]
presenting_1 | 	at com.twitter.zipkin.gen.scribe$Client.recv_Log(scribe.java:74) ~[brave-spancollector-scribe-3.2.0.jar!/:na]
presenting_1 | 	at com.twitter.zipkin.gen.scribe$Client.Log(scribe.java:61) ~[brave-spancollector-scribe-3.2.0.jar!/:na]
presenting_1 | 	at com.github.kristofa.brave.scribe.SpanProcessingThread.log(SpanProcessingThread.java:120) [brave-spancollector-scribe-3.2.0.jar!/:na]
presenting_1 | 	at com.github.kristofa.brave.scribe.SpanProcessingThread.log(SpanProcessingThread.java:110) [brave-spancollector-scribe-3.2.0.jar!/:na]
presenting_1 | 	at com.github.kristofa.brave.scribe.SpanProcessingThread.call(SpanProcessingThread.java:96) [brave-spancollector-scribe-3.2.0.jar!/:na]
presenting_1 | 	at com.github.kristofa.brave.scribe.SpanProcessingThread.call(SpanProcessingThread.java:36) [brave-spancollector-scribe-3.2.0.jar!/:na]
presenting_1 | 	at java.util.concurrent.FutureTask.run(FutureTask.java:266) [na:1.8.0_66]
presenting_1 | 	at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1142) [na:1.8.0_66]
[...]

PR #101, opened 2016-01-13
[#84] Adding missing name to the filter

marcingrzejszczak:


marcingrzejszczak:
Fixed in https://github.com/spring-cloud/spring-cloud-sleuth/commit/c6b2369a4543009dd37d2490c7acddd8efe65a59

PR #103, opened 2016-01-13
Route build on both container and Trusty infrastructures

cotsog:
Can you try this out?

I'm curious to see if it's rather the way MAVEN_OPTS are specified that makes a difference or the infrastructure.

Thanks!

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,5 @@
 language: java
-jdk:
-  - oraclejdk8
+jdk: oraclejdk8
 matrix:
   include:
     - sudo: false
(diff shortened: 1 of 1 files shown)


spencergibb:
Why? what are you trying to fix?

cotsog:
Hi @spencergibb,

I'm a Customer Support Engineer at Travis CI and I'm investigating a build failure that @marcingrzejszczak reported.

spencergibb:
Thanks for the clarification. I'll let @marcingrzejszczak take care of it then.

dsyer:
We gave up on docker in travis (and everywhere in fact), so this is now redundant I believe.

PR #105, opened 2016-01-14
Updates to zipkin-java 0.4.1, particularly docker images

adriancole:
Zipkin 0.4.1 obviates some duplication of configuration. It is also the
first version that's published as a docker image.

By switching to the published image of zipkin-java, we can be more
consistent, as `spring-cloud-sleuth-zipkin-stream` is derived from that,
not the scala query service.


adriancole:
cc @marcingrzejszczak suspect there's an opportunity in brewery too

Issue #98, opened 2016-01-12
Change names to match OpenTracing where possible


dsyer:

marcingrzejszczak:
Can we close it? @dsyer @adriancole @spencergibb  ?

PR #107, opened 2016-01-14
Introduces HttpZipkinSpanReporter

adriancole:
Brave's span collector, which turned out to not be a great tool for
direct use. Brave internally creates spans before sending to its
collector, so validation is implicit. The flip side of this is using the
collector directly does not validate spans. This means it is easy to
send invalid ones, for example missing span names. The problem is more
difficult as the data is in binary (thrift).

This introduces HttpZipkinSpanReporter, which validates via zipkin-java
classes before sending on the wire. Moreover, this sends in json to make
debugging problems easier.

This does not fully remove the Brave dependency, as further work is
needed. Particularly, Brave is indirectly referenced in other code.

See https://github.com/openzipkin/zipkin-java/issues/68
See #98 (Reporter is an OpenTracing term)

### DIFF
--- a/spring-cloud-sleuth-samples/spring-cloud-sleuth-sample-messaging/pom.xml
+++ b/spring-cloud-sleuth-samples/spring-cloud-sleuth-sample-messaging/pom.xml
@@ -81,6 +81,10 @@
 			<artifactId>lombok</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>com.github.kristofa</groupId>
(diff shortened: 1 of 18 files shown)


marcingrzejszczak:
@adriancole - the branch has conflicts :(

marcingrzejszczak:
PR with rebased version opened here - https://github.com/spring-cloud/spring-cloud-sleuth/pull/108

PR #108, opened 2016-01-14
Introduces HttpZipkinSpanReporter (rebased)

marcingrzejszczak:
Originally - https://github.com/spring-cloud/spring-cloud-sleuth/pull/107

Original note:
Brave's span collector, which turned out to not be a great tool for
direct use. Brave internally creates spans before sending to its
collector, so validation is implicit. The flip side of this is using the
collector directly does not validate spans. This means it is easy to
send invalid ones, for example missing span names. The problem is more
difficult as the data is in binary (thrift).

This introduces HttpZipkinSpanReporter, which validates via zipkin-java
classes before sending on the wire. Moreover, this sends in json to make
debugging problems easier.

This does not fully remove the Brave dependency, as further work is
needed. Particularly, Brave is indirectly referenced in other code.

See openzipkin/zipkin-java#68
See #98 (Reporter is an OpenTracing term)


spring-cloud-sleuth-samples/spring-cloud-sleuth-sample-messaging/pom.xml
```diff
@@ -81,6 +81,10 @@
 			<artifactId>lombok</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>com.github.kristofa</groupId>
```
dsyer: This sample should not require brave. I though the intention of this change was to remove the brave dependency everywhere, but this is the least likely place to see one.

spring-cloud-sleuth-samples/spring-cloud-sleuth-sample-messaging/src/test/java/integration/IntegrationTestZipkinSpanReporter.java
```diff
@@ -27,13 +28,14 @@
...
+public class IntegrationTestZipkinSpanReporter implements ZipkinSpanReporter {
 
-	public List<Span> hashedSpans = Collections.<Span>synchronizedList(new LinkedList<Span>());
+	public List<Span> hashedSpans = Collections.synchronizedList(new LinkedList<>());
 
 	@Override
-	public void collect(Span span) {
-		super.collect(span);
+	public void report(Span span) {
+		log.info(span);
```
dsyer: Can we please not log spans at INFO level? It gets rather verbose.

spring-cloud-sleuth-samples/spring-cloud-sleuth-sample-zipkin/pom.xml
```diff
@@ -86,6 +86,11 @@
 			<groupId>io.zipkin</groupId>
 			<artifactId>zipkin-java-server</artifactId>
 		</dependency>
+		<dependency>
```
dsyer: More brave dependencies (not less)?

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinAutoConfiguration.java
```diff
@@ -31,25 +28,21 @@
...
 @ConditionalOnProperty(value = "spring.zipkin.enabled", matchIfMissing = true)
 public class ZipkinAutoConfiguration {
 
 	@Bean
-	@ConditionalOnMissingBean(SpanCollector.class)
-	public SpanCollector spanCollector() {
+	@ConditionalOnMissingBean(ZipkinSpanReporter.class)
+	public ZipkinSpanReporter reporter() {
 		ZipkinProperties zipkin = zipkinProperties();
 		String url = "http://" + zipkin.getHost() + ":" + zipkin.getPort();
```
dsyer: Is HTTP common for this? I'd expect at least the option to use HTTPS (even if it's not the default)?

spring-cloud-sleuth-samples/spring-cloud-sleuth-sample-messaging/pom.xml
marcingrzejszczak: I'll have to check out those - only there where we have the Zipkin server in samples brave should remain (only in test scope)

adriancole: this is not new. the samples that have brave added here only have to add it because their parent no-longer includes a brave dependency. they need it because of this https://github.com/openzipkin/zipkin-java/issues/68

if you print the dep tree you'll notice there's in fact less brave

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinAutoConfiguration.java
adriancole: yes it is common, and the point of this change wasn't to change to https it was to replace thrift spans with json ones. We can revisit how urls are specified of course, but I think we are getting distracted.

Issue #106, opened 2016-01-14
Change the IDs from UUIDs to random 64bit


marcingrzejszczak:
The `IdGenerator` should no longer have `UUID` but a `long`

```
public interface IdGenerator {

    /**
     * Generate a new identifier.
     * @return the generated identifier
     */
    long generateId();

}
```

The default implementation of the IdGenerator should have a `Random` injected that can be overiden (via a setter for example).

adriancole:
I don't agree with us exposing an interface for this. Sleuth will be less complicated an easier to support if features such as id customization came after diverse need was established. Just use Random.

adriancole:
For example, designing an interface to control ids isn't likely to be as simple as just making a Func0. Ex. in Brave, the PR on this topic was more experienced *and also abandoned*. If we eagerly design a span id generator, we now have a legacy which might mean breaking it later.

If we want sleuth to be a clean and maintainable project, we need to consider aspects such as the [rule of three](http://blog.codinghorror.com/rule-of-three/) when choosing whether or when to create something.

```java
/**
 * Allows you to control how ids are generated. For example, implementations
 * can encode where in the call tree the span originates from, or use alternate
 * random number generators.
 */
// Intentionally single-method to be implementable with a lambda
public interface SpanIdGenerator {

    /** Returns the next trace id, or span id if parent is null. */
    SpanId nextSpanId(@Nullable SpanId parent);

    final class Default implements SpanIdGenerator {
        private final Random random = new Random();

        @Override
        public SpanId nextSpanId(@Nullable SpanId parent) {
            long nextLong = random.nextLong();
            if (parent == null) {
                return SpanId.create(nextLong, nextLong, null);
            }
            return SpanId.create(parent.getTraceId(), nextLong, parent.getSpanId());
        }
    }
}
```

marcingrzejszczak:
This

> Ex. in Brave, the PR on this topic was more experienced and also abandoned. 

Convinces me. So in every place where ATM were using the interface we would call

```
new Random().nextLong()
```

?

marcingrzejszczak:
Done in https://github.com/spring-cloud/spring-cloud-sleuth/commit/82ed9f78a16ef34db48979762e7bec606d20ff9c

PR #111, opened 2016-01-15
[#106] Converted UUID to Long

marcingrzejszczak:
@adriancole can you check this out? Especially where I 

- do the conversions from `Long` to `String`
- modified the PercentageBasedSampler and all related to that


spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/MilliSpanTests.java
```diff
@@ -16,10 +16,10 @@
 
 package org.springframework.cloud.sleuth;
 
-import java.util.Collections;
-
 import org.junit.Test;
 
+import java.util.Collections;
```
spencergibb: We gotta get you setup up for the right import order :-)

marcingrzejszczak: LOL true ;)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/MilliSpan.java
```diff
@@ -35,10 +35,10 @@
 	private final long begin;
 	private long end = 0;
 	private final String name;
-	private final String traceId;
+	private final long traceId;
 	@Singular
-	private List<String> parents = new ArrayList<>();
-	private final String spanId;
+	private List<Long> parents = new ArrayList<>();
+	private final Long spanId;
```
adriancole: intentionally nullable?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/MilliSpan.java
```diff
@@ -66,9 +66,10 @@ public MilliSpan(long begin, long end, String name, String traceId, List<String>
 	private MilliSpan() {
 		this.begin = 0;
 		this.name = null;
-		this.traceId = null;
+		this.traceId = 0;
```
adriancole: hmm.. I'd expect this for a mutable object, but not an immutable one. If we are allowing this to be null, at least a comment why (I know this is an existing thing, but very surprising and can lead to bugs)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/MilliSpan.java
```diff
@@ -66,9 +66,10 @@ public MilliSpan(long begin, long end, String name, String traceId, List<String>
...
 		this.spanId = null;
 		this.processId = null;
+		this.parents = new ArrayList<>();
```
adriancole: oh.. ok actually looking at fields, looks like this object is somewhat mutable. I think this is a separate issue.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
```diff
@@ -41,18 +41,17 @@
 	 * The spanId is immutable and cannot be changed. It is safe to access this from
 	 * multiple threads.
 	 */
-	String getSpanId();
+	Long getSpanId();
```
adriancole: add `TODO: document why this can be null!` or do that :P

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/MilliSpan.java
adriancole: really uncomfortable about this.. maybe add a TODO, as it looks like one could accidentally end up with trace id 0, which is later documented as random.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
```diff
@@ -41,18 +41,17 @@
...
 	 * A pseudo-unique (random) number assigned to the trace associated with this span
 	 */
-	String getTraceId();
+	long getTraceId();
 
 	/**
 	 * Return a unique id for the process from which this Span originated.
 	 * <p/>
 	 * <p/>
-	 * Will never be null.
```
adriancole: hah!! document when it can be null, or TODO if not yet known. The null-ability in various fields has been a bug factory. I'd also encourage us to make an internal Nullable annotation to re-inforce the point.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/util/SpanIdGenerator.java
```diff
@@ -0,0 +1,11 @@
+package org.springframework.cloud.sleuth.util;
+
+/**
+ * Contract for generating id for spans and traces
+ *
+ * @author Marcin Grzejszczak
+ */
+public interface SpanIdGenerator {
```
adriancole: why? This just opens up the implementation for someone to use something that's not a random generator. I'd remove this type (as discussed earlier) and enforce random. Hooks are easier to add than take away.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/util/RandomLongSpanIdGenerator.java
```diff
@@ -0,0 +1,20 @@
...
+
+	private Random random = new Random();
+
+	@Override
+	public long generateId() {
+		return random.nextLong();
+	}
+
+	public void setRandom(Random random) {
+		this.random = random != null ? random : new Random();
```
adriancole: super scary contract.. please remove this type and require a random to be supplied (and never changed)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/util/LongUtils.java
```diff
@@ -0,0 +1,20 @@
...
+
+
+import org.springframework.util.StringUtils;
+
+/**
+ * Helper class to convert Long to String or return null
+ *
+ * @author Marcin Grzejszczak
+ */
+public class LongUtils {
```
adriancole: I think I mentioned this before, but let's use fixed-length lower-hex encoding.

encoding as a integer will lead to overflows and is unnecessarily different than zipkin for example. It also allows for a negative sign, which can be harder to copy/paste. Moreover, this is used in propagation headers, too, and there's no reason to be different there.

Here's a few places where lower-hex encoding is used.
https://github.com/openzipkin/brave/blob/master/brave-core/src/main/java/com/github/kristofa/brave/IdConversion.java
https://github.com/openzipkin/zipkin-java/blob/master/zipkin-java-core/src/main/java/io/zipkin/internal/JsonCodec.java#L55
https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/tracing/Id.scala#L20
https://github.com/openzipkin/zipkin/blob/master/zipkin-common/src/main/scala/com/twitter/zipkin/json/JsonSpan.scala#L49

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/TraceIdToThresholdComparable.java
```diff
@@ -0,0 +1,33 @@
+package org.springframework.cloud.sleuth.sampler;
+
+/**
+ * Given the absolute value of a random 64 bit trace id, we expect inputs to be balanced across
+ * 0-MAX. Threshold is the range of inputs between 0-MAX that we retain.
+ *
+ * @author Marcin Grzejszczak
+ * @author Adrian Cole
+ */
+final class TraceIdToThresholdComparable implements Comparable<Long> {
```
adriancole: just inline it. the comparable aspect won't be reused.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSampler.java
```diff
@@ -40,7 +38,7 @@ public boolean next(Void info) {
 		if (currentSpan == null) {
 			return false;
 		}
-		return new UuidTraceIdToThresholdComparable(configuration.getPercentage(), converter)
+		return new TraceIdToThresholdComparable(configuration.getPercentage())
```
adriancole: just inline this. ex. if `configuration.getPercentage()` is mutable config, make a final copy. This code is simple enough that it can be inlined, and there's no reason to require an allocation or another type to do this math.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
```diff
@@ -36,10 +37,14 @@
 @ConditionalOnProperty(value="spring.sleuth.enabled", matchIfMissing=true)
 public class TraceAutoConfiguration {
 
+	@Autowired(required = false) Random random;
+
 	@Bean
 	@ConditionalOnMissingBean
-	public IdGenerator traceIdGenerator() {
-		return new JdkIdGenerator();
+	public SpanIdGenerator traceIdGenerator() {
```
adriancole: Change to Random. this whole hierarchy is unneeded

adriancole:
Thanks for putting a work in progress out.

I think after some unneeded types and abstracting is removed, it will shred far more code.

Also, we need a test somewhere that proves hex is the representation form. ex some assertion where traceId is equal to literally "abcd..." as a side-effect of the long conversion to string.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/MilliSpan.java
marcingrzejszczak: Yes, if you only have log based tracing then there doesn't have to be a notion of spans

adriancole: seems an odd optimization to me. We should optimize for the most common use case. Ex. NullTracer also wouldn't need span ids. I understand this is an existing thing, but would like this to be impossible. It is better to make LoggingTracers throw away ids (which itself is odd as most features around logging are in fact correlation!)

marcingrzejszczak: So you would suggest that it is NOT possible to have nullable span id?

Can you elaborate more on this:

> . Ex. NullTracer also wouldn't need span ids. I understand this is an existing thing, but would like this to be impossible. It is better to make LoggingTracers throw away ids (which itself is odd as most features around logging are in fact correlation!)

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinSpanListener.java
```diff
@@ -189,17 +182,7 @@ private void addZipkinBinaryAnnotations(io.zipkin.Span.Builder zipkinSpan,
...
-		if (string == null) {
-			return h;
-		}
-		int len = string.length();
-
-		for (int i = 0; i < len; i++) {
-			h = 31 * h + string.charAt(i);
-		}
-		return h;
+	private long valueOrDefault(Long value) {
```
adriancole: I'd at least comment why we are sending invalid spans. Is this id a magic number users will know about? If not, I'd drop them, as there's no invalid span graveyard. Ex. literally, this will create a trace which collects random invalid span data!

adriancole: huh actually seems this is also existing logic. Please add a TODO, as there is no default span or trace id in zipkin, and this behavior is quite surprising even if not new.

adriancole: One way out is to look at the `event.getSpan().isExportable()` contract. A span without an identifier isn't effectively exportable. we could check ids inside that method. Or, before calling convert, guard on ids being present and don't attempt to report/export spans who have no id. Spans without an id cannot be aggregated into a trace, or correlated with other spans. We have a big TODO around how it is even possible to create id-less spans!

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinSpanListener.java
```diff
@@ -16,27 +16,20 @@
...
-
 import io.zipkin.Annotation;
 import io.zipkin.BinaryAnnotation;
 import io.zipkin.Constants;
 import io.zipkin.Endpoint;
-
 import lombok.extern.apachecommons.CommonsLog;
+import org.springframework.cloud.sleuth.Log;
+import org.springframework.cloud.sleuth.Span;
+import org.springframework.cloud.sleuth.event.*;
```
adriancole: nit: your ide is wildcarding

spring-cloud-sleuth-samples/spring-cloud-sleuth-sample-test-core/src/main/java/tools/RequestSendingRunnable.java
```diff
@@ -50,9 +51,9 @@ public void run() {
 		log.info("Received the following response [{}]", responseEntity);
 	}
 
-	private RequestEntity requestWithTraceId(String traceId) {
+	private RequestEntity requestWithTraceId(Long traceId) {
 		HttpHeaders headers = new HttpHeaders();
-		headers.add(Trace.TRACE_ID_NAME, traceId);
+		headers.add(Trace.TRACE_ID_NAME, LongUtils.toString(traceId));
```
adriancole: These util classes :P When you introduce a type, you now have the problem of naming it. Util is a junk drawer, and toString doesn't really tell me anything. I have to look at call sites to see .. ahh this is for span ids! IOTW, this code is less about Longs and more about span ids, why not call it that?

`SpanId.toString(long)/fromString(String)` would be closer, as at least that indicates we are going in and out of the canonical string form of a span id.

adriancole: another option would just to be explicit. ex hexEncode hexDecode.. there are existing utilities for that. ex you don't need to name a class when you don't declare one.

adriancole:
I think this is still missing the change to make span ids hex encoded?

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinSpanListener.java
adriancole: added this issue about the general problem https://github.com/spring-cloud/spring-cloud-sleuth/issues/114

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
```diff
@@ -127,4 +130,26 @@
 	 * Will never be null.
 	 */
 	List<Log> logs();
+
+
+	/**
+	 * Class used for conversions of long ids to their String representation
+	 */
+	class Converter {
```
adriancole: Style feedback

this reads `Span.Converter.toHexString(id)` I think this distracts the span type a bit, as it only applies to its id. Ex converter (or codec) makes me think about the span, and then I get to hexString and notice it is looking for an id.

ps Since hex is already a string, I don't think you need to qualify the methods by appending String to them.

I think SpanId.toHex/fromHex or even Hex.encode/decode would be better. Use `SpanId` or similar, if you want to clarify the relationship about ids. Use `Hex.` if you are just looking to have these codecs somewhere common. Both are valid, imho, but in the end.. I'd probably choose Hex.encode/decode as it is shorter.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
```diff
@@ -36,11 +35,7 @@
 @ConditionalOnProperty(value="spring.sleuth.enabled", matchIfMissing=true)
 public class TraceAutoConfiguration {
 
-	@Bean
-	@ConditionalOnMissingBean
-	public IdGenerator traceIdGenerator() {
-		return new JdkIdGenerator();
-	}
+	@Autowired(required = false) Random random;
```
adriancole: required = false, and not used in this file.. am I missing something?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
marcingrzejszczak: I can't use `Span.toHex` etc. since static methods in interface are allowed in JDK8 :)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
marcingrzejszczak: Good catch - fixed

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
adriancole: hah I ^H'd over a comment where I was guessing java 8 default methods was why!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/integration/AbstractTraceChannelInterceptor.java
```diff
@@ -29,33 +30,42 @@ protected AbstractTraceChannelInterceptor(TraceManager traceManager) {
...
-			if (processId != null) {
-				span.processId(processId);
-			}
-			if (parentId != null) {
-				span.parent(parentId);
-			}
-			span.remote(true);
-			return span.build();
+		if (!hasHeader(message, Trace.TRACE_ID_NAME) || !hasHeader(message, Trace.SPAN_ID_NAME)) {
+			return null;
```
adriancole: love the early escape. maybe add a trailing comment like.. `return null; // cannot build a span without ids` or similar

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/integration/AbstractTraceChannelInterceptor.java
```diff
@@ -29,33 +30,42 @@ protected AbstractTraceChannelInterceptor(TraceManager traceManager) {
...
+		}
+		long spanId = hasHeader(message, Trace.SPAN_ID_NAME) ?
+				getHeader(message, Trace.SPAN_ID_NAME, Long.class) : new Random().nextLong();
```
adriancole: for random to be effective, it has to be kept around. I'd use the same random instance for all callsites that generate ids.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/integration/AbstractTraceChannelInterceptor.java
```diff
@@ -29,33 +30,42 @@ protected AbstractTraceChannelInterceptor(TraceManager traceManager) {
...
+		span.remote(true);
+		return span.build();
 	}
 
 	String getHeader(Message<?> message, String name) {
-		return (String) message.getHeaders().get(name);
+		return getHeader(message, name, String.class);
+	}
+
+	<T> T getHeader(Message<?> message, String name, Class<T> type) {
```
adriancole: suggestion, if you make this getHeaderAsLong.. you are down to 2 args, and slightly less complex looking call sites

spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/ZipkinMessageListener.java
```diff
@@ -94,15 +84,15 @@ public void sink(Spans input) {
...
 		if (span.getParents().size() > 0) {
 			if (span.getParents().size() > 1) {
 				log.error("zipkin doesn't support spans with multiple parents.  Omitting "
 						+ "other parents for " + span);
 			}
-			zipkinSpan.parentId(hash(span.getParents().get(0)));
+			zipkinSpan.parentId(span.getParents().get(0));
 		}
-		zipkinSpan.id(hash(span.getSpanId()));
+		zipkinSpan.id(span.getSpanId());
```
adriancole: yes!

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinSpanListener.java
```diff
@@ -189,17 +186,4 @@ private void addZipkinBinaryAnnotations(io.zipkin.Span.Builder zipkinSpan,
 		}
 	}
 
-	private static long hash(String string) {
```
adriancole: very happy that there's no default needed anymore

adriancole:
Much better. :ship: when ready

marcingrzejszczak:
Done in https://github.com/spring-cloud/spring-cloud-sleuth/commit/82ed9f78a16ef34db48979762e7bec606d20ff9c

PR #115, opened 2016-01-18
Pares down default http tags, but opens apis for adding more

### LINKED ISSUE #82: Pare down what TraceFilter is logging

adriancole:
Previously, sleuth logged more data than necessary. This led to issues
including spans approaching megabytes each. This pares down default keys
to a minimal set, but allows users to opt-into logging more by extending
`TraceFilter`.

Closes #82
Closes #95 (Obviates)

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/TraceKeys.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/TraceKeys.java
@@ -1,0 +1,95 @@
+/*
+ * Copyright 2013-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.sleuth.instrument;
+
+/**
+ * Well-known {@link org.springframework.cloud.sleuth.Span#tag(String, String) span tag} keys.
+ *
+ * <h3>Overhead of adding Trace Data</h3>
+ *
+ * Overhead is directly related to the size of trace data exported out of process. Accordingly, it
+ * is better to tag what's important for latency troubleshooting, ie a whitelist, vs collecting
+ * everything and filtering downstream. The keys listed here are very common in tracing tools, and
+ * are considerate to the issue of overhead.
+ *
+ * <p>When evaluating new keys, consider how much additional data it implies, and if that data is
+ * critical to classifying, filtering or displaying traces. More data often means larger systems,
+ * less retention, or a lower sample rate.
+ *
+ * <p>For example, in zipkin, a thrift-encoded span with an "sr" annotation is 82 bytes plus the
+ * size of its name and associated service. The maximum size of an HTTP cookie is 4096 bytes,
+ * roughly 50x that. Even if compression helps, if you aren't analyzing based on cookies, storing
+ * them displaces resources that could be used for more traces. Meanwhile, you have another system
+ * storing private data! The takeaway isn't never store cookies, as there are valid cases for this.
+ * The takeaway is to be conscious about what's you are storing.
+ */
+public final class TraceKeys {
+
+  /**
+   * The domain portion of the URL or host header. Ex. "mybucket.s3.amazonaws.com"
+   *
+   * <p>Used to filter by host as opposed to ip address.
+   */
+  public static final String HTTP_HOST = "http/host";
+
+  /**
+   * The HTTP method, or verb, such as "GET" or "POST".
+   *
+   * <p>Used to filter against an http route.
+   */
+  public static final String HTTP_METHOD = "http/method";
+
+  /**
+   * The absolute http path, without any query parameters. Ex. "/objects/abcd-ff"
+   *
+   * <p>Used to filter against an http route, portably with zipkin v1.
+   *
+   * <p>In zipkin v1, only equals filters are supported. Dropping query parameters makes the number
+   * of distinct URIs less. For example, one can query for the same resource, regardless of signing
+   * parameters encoded in the query line. This does not reduce cardinality to a HTTP single route.
+   * For example, it is common to express a route as an http URI template like
+   * "/resource/{resource_id}". In systems where only equals queries are available, searching for
+   * {@code http.uri=/resource} won't match if the actual request was "/resource/abcd-ff".
+   *
+   * <p>Historical note: This was commonly expressed as "http.uri" in zipkin, eventhough it was most
+   * often just a path.
+   */
+  public static final String HTTP_PATH = "http/path";
+
+  /**
+   * The entire URL, including the scheme, host and query parameters if available. Ex.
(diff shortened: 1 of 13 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebAutoConfiguration.java
```diff
@@ -65,12 +66,17 @@ public TraceWebAspect traceWebAspect() {
 	}
 
 	@Bean
-	public FilterRegistrationBean traceWebFilter(ApplicationEventPublisher publisher) {
+	@ConditionalOnMissingBean
+	public TraceFilter traceFilter(ApplicationEventPublisher publisher) {
 		Pattern pattern = StringUtils.hasText(this.skipPattern) ? Pattern.compile(this.skipPattern)
```
spencergibb: Not positive, but you might get double registrations as boot automatically registers filters. @dsyer could day for sure.

spencergibb: Should have been one line above.

adriancole: Happy to adjust this in any way. just lemme know how

spencergibb: I don't think you need the filter as a bean and the filter registration bean, just one of the two.

adriancole: thx only added it since it was in #95. reverted

adriancole:
green now.. forgot to add the file with the tags in it! PTAL

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/HttpTags.java
```diff
@@ -0,0 +1,72 @@
...
+   * <p>Note: This is not logged in as many systems as {@link #HTTP_URI}. Proceed with caution as
+   * the URL may include private data encoded as query parameters, and be considerably larger than
+   * other fields.
+   */
+  public static final String HTTP_URL = "http/url";
+
+  /**
+   * The HTTP response code, such as 500.
+   *
+   * <p>Used to filter for error status.
```
adriancole: note: in brave, this is only logged for non-2xx range. I think this makes sense, as it is only interesting in error case.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/TraceKeys.java
```diff
@@ -0,0 +1,94 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.sleuth.instrument;
+
+/**
+ * Well-known {@link org.springframework.cloud.sleuth.Span#tag(String, String) span tag} keys.
```
adriancole: @kristofa @jcarres-mdsol I'm thinking about adding the same file in brave, and the ruby tracer. The key namespaces are encoded differently than some zipkin implementations, but I'd consider this a one-punch type of problem.

With the exception of `http/uri`, these keys are literally the same as google cloud trace, but also quite similar to existing ones.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebAutoConfiguration.java
```diff
@@ -65,6 +66,7 @@ public TraceWebAspect traceWebAspect() {
 	}
 
 	@Bean
+	@ConditionalOnMissingBean
```
marcingrzejszczak: This is bad. We can't have filters with `ConditionalOnMissingBean`. If anyone has his own filter this one won't be registered

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/TraceKeys.java
jcarres-mdsol: We already do this in the ruby-tracer in different ways 
HOST is set in the endpoint.
METHOD is set in the rpc_name
PATH , we call it in ruby "http.uri"
URL, we are not logging it. OK to have support but would like to make it optional as it may be some huge urls in there. 
STATUS we are logging it as "http.status"

adriancole: @jcarres-mdsol cool. I did caveat the URL, but I'll be more explicit about it being optional.

Do you feel strongly about "http.uri" vs "http/path"? I feel the latter is more honest as we strip query params and there's nothing about "uri" that indicates that.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebAutoConfiguration.java
adriancole: ok will revert to the previous approach cc @spencergibb

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/TraceKeys.java
jcarres-mdsol: yes the only reason is 'http.uri' now is because the original code was using that key.
http/path seems better

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebAutoConfiguration.java
adriancole: done

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/TraceKeys.java
kristofa: :+1:

PR #120, opened 2016-01-19
[Match OpenTracing] Matching Tracer

marcingrzejszczak:
cc @adriancole @dsyer @spencergibb 

## Removed notion of Trace

According to https://github.com/opentracing/opentracing-python/blob/master/example/zipkin_like/tracer.py#L35 

there is no notion of Trace. There is always a Span. So I've removed Trace and Span knows about a previously saved Span.

The next thing would be to make Span even more intelligent so that it can have things like `Span.continue()` etc. WDYT?


dsyer:
I quite like to have a `Trace`. How do you keep track of the fact that there is already a `Trace` when a new one starts (the `savedTrace` in the existing implementation)? That state has to be stored somewhere (and it can't be in the `Span` because there is not a unique parent).

PR #122, opened 2016-01-20
[Match OpenTracing] Matching Sampler

marcingrzejszczak:
Changed sampler method to be consistent with open tracing API:

https://github.com/opentracing/opentracing-python/blob/master/example/zipkin_like/sampler.py#L36

cc @adriancole @dsyer @spencergibb


adriancole:
LGTM

Issue #99, opened 2016-01-12
Basic health metrics


adriancole:
Arising from some discussions, it seems the most basic primitives we need to assess the health of the tracing system are spans accepted vs dropped.

If we have these metrics on the producer side (ex boot apps) and the consumer side (stream listener), we can get at least a basic signal of health.

This topic was discussed in brave, but I can add more context and opinions here, if it could help.

FWIW, here is the callback brave provides to address this issue. Note that the consuming callback would be quite similar. In any case, this sort of callback is easy to bind to any number of metrics systems.

```java
public interface SpanCollectorMetricsHandler {

    /**
     * Called when spans are submitted to SpanCollector for processing.
     *
     * @param quantity the number of spans accepted.
     */
    void incrementAcceptedSpans(int quantity);

    /**
     * Called when spans become lost for any reason and won't be delivered to the target collector.
     *
     * @param quantity the number of spans dropped.
     */
    void incrementDroppedSpans(int quantity);

}
```

See https://github.com/openzipkin/brave/pull/101

marcingrzejszczak:
We can somehow connect this with Codehale metrics and add this to health endpoint

marcingrzejszczak:
In progress - https://github.com/spring-cloud/spring-cloud-sleuth/pull/123/files

PR #123, opened 2016-01-20
[#99] Basic Health Metrics

marcingrzejszczak:


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
```diff
@@ -51,4 +58,39 @@ public DefaultTracer traceManager(Sampler sampler,
 									ApplicationEventPublisher publisher) {
 		return new DefaultTracer(sampler, random(), publisher);
 	}
+
+	@Configuration
+	@ConditionalOnProperty(value = "spring.sleuth.health", matchIfMissing = true)
```
spencergibb: `spring.sleuth.health.enabled`

dsyer:
Wouldn't it be better to build this on the `GaugeService` and `CounterService`? There is a wrapper for Dropwizard if users have that on the classpath.-

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/health/SpanCollectorMetricsHandler.java
```diff
@@ -0,0 +1,29 @@
+package org.springframework.cloud.sleuth.health;
+
+public interface SpanCollectorMetricsHandler {
```
spencergibb: `SpanMetricsCollector`? `SpanCollectorMetricsHandler` seems unwieldy.

spencergibb: Maybe even just `SpanMetrics`.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/health/SpanCollectionHealthIndicator.java
```diff
@@ -0,0 +1,26 @@
+package org.springframework.cloud.sleuth.health;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.actuate.health.AbstractHealthIndicator;
+import org.springframework.boot.actuate.health.Health;
+
+/**
+ * Health indicator that shows the number of accepted / dropped spans.
+ *
+ * //TODO: should this be a health indicator (it's never down) or an Endpoint ?
```
shakuzen: If there were a (configurable) threshold of dropped spans (a percent), then there could be a down state, and it would be more health-like. Though I don't know if the application should necessarily be considered down because it is dropping too many spans.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/health/InMemorySpanCollectorMetricsHandler.java
```diff
@@ -0,0 +1,32 @@
+package org.springframework.cloud.sleuth.health;
+
+import java.util.concurrent.atomic.AtomicLong;
+
+/**
+ * Implementation of the metrics statistics held in-memory.
+ *
+ * @author Marcin Grzejszczak
+ */
+public class InMemorySpanCollectorMetricsHandler implements SpanCollectorMetricsHandler {
```
adriancole: I don't think this should be the default, no-op should be the default. There's no benefit to incrementing atomic longs no-one will read.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/health/SpanCollectorMetricsHandler.java
```diff
@@ -0,0 +1,29 @@
...
+	 * Called when spans become lost for any reason and won't be delivered to the target collector.
+	 *
+	 * @param quantity the number of spans dropped.
+	 */
+	void incrementDroppedSpans(long quantity);
+
+	/**
+	 * Returns the number of accepted spans
+	 */
+	long getAcceptedSpans();
```
adriancole: I don't think we should be exposing getters. real metrics systems will export these metrics, either to a local endpoint (along with other metrics of use) or outside the process. The concern of collecting metrics doesn't need to be conflated with returning them

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/health/SpanCollectionHealthIndicator.java
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * @author Marcin Grzejszczak
+ */
+public class SpanCollectionHealthIndicator extends AbstractHealthIndicator {
```
adriancole: I'd really skip this for now, and just focus on collecting metrics. Let's leave health, thresholds etc. in the space of metrics systems (as opposed to a hard-wired aspect of `SpanCollectorMetricsHandler`). If someone wants to read stats like below, they could just as well look at their metrics endpoint locally or otherwise.

adriancole: ex this does a pretty good job separating the act of reporting stats (like accept vs dropped) from health check

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/health/SpanCollectionHealthIndicator.java
adriancole: mentioned elsewhere, but I totally agree that healthcheck is a different concern, or at least one harder to pin down with a ratio of accept vs drop without any history.

Local, health checks are usually something you can answer yes/no to. For example, is the client busted? One way is to inject the transport itself (ex the reporter) and expose a state of the connection.
https://github.com/Netflix/suro/blob/master/suro-server/src/main/java/com/netflix/suro/server/HealthCheck.java

@kristofa any thoughts on this?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/health/SpanCollectorMetricsHandler.java
adriancole: I think this was a C&P? the literal binding is to SpanReporter I think. Maybe we rename things in this file about the act of reporting. Ex. SpanReporterMetrics or similar

spring-cloud-sleuth-zipkin/src/test/java/org/springframework/cloud/sleuth/zipkin/HttpZipkinSpanReporterTest.java
```diff
@@ -16,10 +18,11 @@
 public class HttpZipkinSpanReporterTest {
 
 	@Rule public final MockWebServer server = new MockWebServer();
+	SpanCollectorMetricsHandler spanCollectorMetricsHandler = new InMemorySpanCollectorMetricsHandler();
```
adriancole: yep. ex in-memory could be in the test tree, as it only really makes sense here.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/health/SpanCollectionHealthIndicator.java
smaldini: As an orthogonal note, we do work on exposing state for reactive, asynchronous (and even synchronous) flows in Spring 5 and any other Reactor supported projects. I plan to come back to you at some point but we are able to do this kind of crazy graphing : 
screen shot 2016-01-10 at 8 25 18 pm

adriancole: ooooo :) thanks for the preview!!

marcingrzejszczak: This looks amazing!

smaldini: btw the massive circle on the left is the realtime monitored "flatMap" or dynamic merge, but in this screenshot I disabled cleaning "completed state" to accumulate a bit and replay (we will offer blackbox recorder that shows step by step the state changing within the console UI). What we don't do is of course, tracing, and the combination of both will prove massively useful for next gen of reactive microservices!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
```diff
@@ -51,4 +55,19 @@ public DefaultTracer traceManager(Sampler sampler,
 									ApplicationEventPublisher publisher) {
 		return new DefaultTracer(sampler, random(), publisher);
 	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public CounterService noOpSpanCounterService() {
+		return new NoOpSpanCounterService();
```
dsyer: I don't see why we need this. We should just opt out of collecting metrics if there is no `CounterService` available.

hanskoff:
awesome ;)

spring-cloud-sleuth-samples/spring-cloud-sleuth-sample-zipkin-stream/src/main/java/example/ZipkinStreamServerApplication.java
```diff
@@ -9,7 +9,7 @@
 public class ZipkinStreamServerApplication {
 
 	public static void main(String[] args) throws Exception {
-		SpringApplication.run(ZipkinStreamServerApplication.class, args);
+		SpringApplication.run(ZipkinStreamServerApplication.class, "--debug");
```
adriancole: careful!

adriancole:
I like where you are going with this.

marcingrzejszczak:
https://github.com/spring-cloud/spring-cloud-sleuth/commit/301c6b230ae771d512aa2436ce5b396ac6e5065f

PR #129, opened 2016-01-23
Changes http key delimiter to dot and adds size keys

adriancole:
per discussion here https://github.com/openzipkin/zipkin/pull/909


spring-cloud-sleuth-zipkin-stream/src/test/java/org/springframework/cloud/sleuth/zipkin/stream/ZipkinMessageListenerTests.java
```diff
@@ -36,7 +36,7 @@
 	@Test
 	public void convertsTimestampAndDurationToMicroseconds() {
 		long start = System.currentTimeMillis();
-		this.span.log("http/request/retry"); // System.currentTimeMillis
+		this.span.log("hystrix/retry"); // System.currentTimeMillis
```
adriancole: changed this only to dodge the http namespace issue, as the event was contrived anyway :P

PR #130, opened 2016-01-25
Prefers the word event when adding to a Span's timeline

adriancole:
Span.logs are really timestamped events. This clarifies the naming and
documentation around that.

See https://github.com/opentracing/opentracing-go/pull/39


marcingrzejszczak:
LGTM

Issue #128, opened 2016-01-22
Message producer not starting new span


dsyer:
This is a new feature and I think it's a bug. Let's say you receive an http request that means you have to send a message. The result should be a new span. If we count the spans in the brewery, I bet they have changed since last week.

marcingrzejszczak:
Assuming that we take the set up into account we should have the following flow:

We should have 6 spans:
- first HTTP incoming message with its parent (2 spans)
- then there is the message sent and received (2 spans)
- then I'm calling rest template to myself with the HTTP message (1 span)
- then I'm receiving that HTTP message (2 spans  but one is the same as the one in the first step)

2 + 2 + 1 + 1 = 6 spans. ATM we're missing this part: 

> then there is the message sent and received

PR #131, opened 2016-01-25
[#128] Added spans for messaging

marcingrzejszczak:
After this fix an additional span is created after the message has been sent

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/integration/TraceChannelInterceptor.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/integration/TraceChannelInterceptor.java
@@ -16,14 +16,14 @@
 
 package org.springframework.cloud.sleuth.instrument.integration;
 
+import java.util.Random;
+
 import org.springframework.cloud.sleuth.Span;
 import org.springframework.cloud.sleuth.Tracer;
 import org.springframework.cloud.sleuth.instrument.TraceKeys;
 import org.springframework.cloud.sleuth.sampler.IsTracingSampler;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageChannel;
-
-import java.util.Random;
 
 /**
  * @author Dave Syer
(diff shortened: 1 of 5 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/integration/TraceChannelInterceptor.java
```diff
@@ -16,6 +16,8 @@
 
 package org.springframework.cloud.sleuth.instrument.integration;
 
+import java.util.Random;
```
adriancole: I agree.. well let me take that back.. I think I agree.. or ..

spring-cloud-sleuth-samples/spring-cloud-sleuth-sample-messaging/src/test/java/integration/MessagingApplicationTests.java
```diff
@@ -43,6 +45,8 @@
 @TestPropertySource(properties="sample.zipkin.enabled=true")
 public class MessagingApplicationTests extends AbstractIntegrationTest {
 
+	private static final String SERVER_SENT = "ss";
+	private static final String CLIENT_RECEIVED = "cr";
```
adriancole: super-duper-nit.. RECV allows these both to be.. wait for it.. the same length!

adriancole: oh wait.. you have zipkin-java in the classpath, so you could use Constants.^^

spring-cloud-sleuth-samples/spring-cloud-sleuth-sample-messaging/src/test/java/integration/MessagingApplicationTests.java
```diff
@@ -99,14 +103,56 @@ private void thenAllSpansHaveTraceIdEqualTo(long traceId) {
...
+
+	private Optional<Span> findLastHttpSpan() {
+		return this.integrationTestSpanCollector.hashedSpans.stream()
 				.filter(span -> "http/foo".equals(span.name)).findFirst();
-		then(firstSpan.isPresent()).isTrue();
-		then(lastSpan.isPresent()).isTrue();
-		then(lastSpan.get().parentId).isEqualTo(firstSpan.get().id);
+	}
+
+	private Optional<Span> findSpanWithClientReceivedAnnotation(List<Span> eventSpans) {
```
adriancole: this logic could be deduped by adding a param for the annotation value. I think it would read almost as well.

```
findSpanWithServerSentAnnotation(eventSpans)
findSpanWithAnnotation(eventSpans, SERVER_SENT)
```

adriancole:
cool looks like you've squashed something otherwise hard to catch!

adriancole:
:shipit:

PR #140, opened 2016-02-02
Delomboking

marcingrzejszczak:


adriancole:
presuming this is a mechanical change LGTM: ok eventhough there's more code

cemo:
@marcingrzejszczak Have you ever experienced an issue with Lombok? What is the motivation to remove it from project? 

BTW thanks for your all efforts for this project. I am sure this will be a huge step in Spring land.

adriancole:
This is a week old, so we should decide to merge it or dump it.

Personal viewpoint in favor of merging this, even if I'd rather dump than keep it unmerged for longer:

Lombok was a distraction for me, when I first started looking deeply at code. After I learned how to setup the build's profile, and the lombok plugin, things work. Lombok seems to value less lines, but not just good ones. Ex. it makes it less lines to do `@SneakyThrows`. When I see things like this, I take a deep breath.. and try to ignore it rather than getting upset.

Personally, I like the google auto project, as you can see the lines written, and they are quite quality. Eventhough it is magical (doing lots of code for you), it is less magical (as source lines are generated directly), and also there's no runtime dependency (as annotations are source retention). Most of all, the feature scope is narrow: it addresses the most commodity change (like toString/equals/hashCode, etc), in quality ways (ex preferring not null).

If given the choice between lombok and auto, definitely auto, especially in an instrumentation library. If given the choice between lombok or auto vs nothing (ex to avoid polarizing technology)... personally I'd choose nothing.

spencergibb:
We decided last week to do delombok, just a matter of when.

marcingrzejszczak:
Closing this one and will create a new one

PR #154, opened 2016-02-10
Updates to zipkin-java 0.5

adriancole:
This includes better dependency aggregation than before. For example,
you no longer need mysql to view the dependency graph, as the in-memory
server supports aggregation now. Also, if you log "ca" on a root span,
the associated service will show on the dependency graph on the far
left. This allows you to show uninstrumented clients of note. Similarly,
if you log "sa" on a leaf node, the destination will show on the far
right. You'd use "sa" for uninstrumented services like databases, auth
servers or cloud apis.

Here's an example. Note you can now see the database service on the far right.

screen shot 2016-02-10 at 11 28 46 pm

### DIFF
--- a/spring-cloud-sleuth-dependencies/pom.xml
+++ b/spring-cloud-sleuth-dependencies/pom.xml
@@ -17,7 +17,7 @@
 		<brave.version>3.4.0</brave.version>
 		<spring-cloud-netflix.version>1.1.0.BUILD-SNAPSHOT</spring-cloud-netflix.version>
 		<aspectj.version>1.8.4</aspectj.version>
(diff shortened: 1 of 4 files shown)


marcingrzejszczak:
+1

adriancole:
going to bump to 0.5.1 which is based against scala server 1.33.0 tomorrow (then merge)

adriancole:
just waiting on maven central sync

adriancole:
@marcingrzejszczak might be interesting to re-look at the dependency graph of brewery now

marcingrzejszczak:
Sweeeeeet. 

deps_are_working

PR #164, opened 2016-02-15
Improve support of async spans

marcingrzejszczak:
- Replaced `SpanName` back with the String
- `SpanHolder` is a dumb structure
- `SpanStarter` is a utility class that operates on `SpanHolder` to either start or continue / detach or close a span
- Added a lot of continuing of Spans instead of starting new ones
- What used to be a span name now ended up as a `origin` tag e.g. `async:className#method=methodName`


adriancole:
There's a lot of types of change in this PR. Single-purpose PRs are nicer
to the reviewer as they allow them to say "awesome"! faster. For example,
I'd certainly +1 rolling back span name, but there's other things here that
would hold up merging that, such as...

SpanStarter feels like an unnecessary addition.. should this be a helper on
Tracer?
SpanHolder is just all-together weird. It looks like state that should be
in Span

When you decouple decouplable things, then it is easier to review and push
forward some issues that are not necessarily blocking others.

Also, if the intent is to help reviewers say yes, please at least minimally
document your change, if not in javadoc on new types, in the PR
description. If this isn't meant for review, make a note of that or just
push a private branch instead of a pull request.

marcingrzejszczak:
I've added little documentation in the PR description. Actually it depicts all the changes (I think)

I don't like SpanStarter either - we had this discussion on Gitter that it seems like a good idea for now to have it. And yeah it seems like a good idea to annotate it with @Internal . most likely it will get merged to tracer at one point.

Of course I didn't plan to merge both things together in one PR but "it escalated quickly" ;)

marcingrzejszczak:
I'm closing this PR and will create a new one

Issue #168, opened 2016-02-16
Improve support of async spans


marcingrzejszczak:
`@Async` affects the structure of the trace, as their execution is tracked in a child span. This is a child span because they are moved off the critical path of the remote caller. Because this child span is created in-process, we can consider it a local span.

* Span.tags include LOCAL_COMPONENT("lc") which allocates the span to a component, such as "hystrix", or "async" if unspecified or unknown.
* Span.name is the operation name and component-specific. For example, if the component is "hystrix", it might be the command key. If the component is unspecified or unknown ("async"), it is the name of the java method with the `@Async` annotation

The components can indicate a common tag namespace, eg in TraceKeys, but developing that is more about the component in question than the nature of async. Hence, it can be done separately, like we did with http.

PR #169, opened 2016-02-17
[#168] Introduced local components for async

marcingrzejszczak:
* ADded TraceKeys for Async and HTTP
* Async from Callables in RestControllers continue span
* ALl the other async are creating a local component span


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/LazyTraceExecutor.java
```diff
@@ -47,7 +54,17 @@ public void execute(Runnable command) {
 				this.delegate.execute(command);
 			}
 		}
-		this.delegate.execute(new TraceRunnable(this.tracer, command));
+		if (this.traceKeys == null) {
+			try {
+				this.traceKeys = this.beanFactory.getBean(TraceKeys.class);
```
adriancole: Why is there no default? I'd expect it to be the other way around, assume the normalized keys and expect the occasional override

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/TraceContinuingCallable.java
```diff
@@ -0,0 +1,42 @@
...
+
+package org.springframework.cloud.sleuth.instrument.async;
+
+import java.util.concurrent.Callable;
+
+import org.springframework.cloud.sleuth.Span;
+import org.springframework.cloud.sleuth.Tracer;
+
+/**
+ * @author Spencer Gibb
```
adriancole: spencer is awesome

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/TraceDelegate.java
```diff
@@ -52,7 +52,7 @@ protected Span startSpan() {
 
 	protected String getSpanName() {
 		return this.name == null ?
-				ASYNC_COMPONENT + ":" + Thread.currentThread().getName()
```
adriancole: kindof a crappy name :P if span name ~ operation name, the name async doesn't really clarify anything especially as there's a component tag already named async. Even the thread name isn't that great. Why not fallback to toString on the delegate?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/TraceKeys.java
```diff
@@ -283,4 +303,96 @@ public void setHeaders(Collection<String> headers) {
...
+		 */
+		private String threadNameKey = "thread";
+
+		/**
+		 * Name of the class from which the asynchronous process started
+		 */
+		private String classNameKey = "class";
+
+		/**
+		 * Name of the method from which the asynchronous process started
```
adriancole: is this analogous to "name of the method annotated with `@Async` ?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/TraceKeys.java
```diff
@@ -283,4 +303,96 @@ public void setHeaders(Collection<String> headers) {
...
+
+		/**
+		 * Prefix for header names if they are added as tags.
+		 */
+		private String prefix = "";
+
+		/**
+		 * Name of the thread from which the async processing started
+		 */
+		private String threadNameKey = "thread";
```
adriancole: name of the thread that executed the async method

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/TraceKeys.java
```diff
@@ -283,4 +303,96 @@ public void setHeaders(Collection<String> headers) {
...
+	 */
+	public static class Hystrix {
+
+		/**
+		 * Prefix for header names if they are added as tags.
+		 */
+		private String prefix = "";
+
+		/**
+		 * Name of the command key
```
adriancole: link

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
```diff
@@ -70,6 +71,12 @@ public SleuthMetricProperties sleuthMetricProperties() {
 		return new SleuthMetricProperties();
 	}
 
+	@Bean
+	@ConditionalOnMissingBean
```
adriancole: this is why I don't understand the defensive lookup below.. if this allows a default, why would we guard on TraceKeys not found?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/TraceDelegate.java
```diff
@@ -24,7 +24,7 @@
  */
 public abstract class TraceDelegate<T> {
```
adriancole: is this type used anywhere except from a runnable or a callable? If so, I'd probably flatten the type hierarchy to just runnable, callable as the types aren't interchangeable anyway, and the work here isn't high value.

adriancole:
This change is positive, but it has highlighted some pretty intense hierarchy :)

I'd axe the "delegate" hierarchy stuff as it indirects things and seems to only serve two incompatible entry points: callable and runnable.

What if instead, you killed everything underneath two types:  LocalSpanCallable, LocalSpanRunnable?

I suspect the line count and indirection would go way down at the cost of duplicating some rather easy to test annotation adding (even that could be a package private utility)

marcingrzejszczak:
That's actually what I had initially and I changed it to this. So if we both thought about this then most likely it v makes more sense ;)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/SpanName.java
```diff
@@ -0,0 +1,36 @@
...
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ *
+ * Annotation to provide the name for the Span
+ *
+ * @author Marcin Grzejszczak
+ */
+@Target(ElementType.TYPE)
```
adriancole: type? or would this also be method retention? in clean-up time, there'd be a code block showing an example, such as on an Async (or even better documenting it is experimental and would only work on Async + whatever else)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/TraceKeys.java
```diff
@@ -283,4 +303,96 @@ public void setHeaders(Collection<String> headers) {
...
+		 */
+		private String threadNameKey = "thread";
+
+		/**
+		 * Name of the class from which the asynchronous process started
+		 */
+		private String classNameKey = "class";
+
+		/**
+		 * Name of the method annotated with @Async
```
adriancole: does this render? maybe wrap it in a code tag? `{@code @Async}`

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/TraceKeys.java
```diff
@@ -283,4 +303,96 @@ public void setHeaders(Collection<String> headers) {
...
+		 */
+		private String prefix = "";
+
+		/**
+		 * Name of the thread that executed the async method
+		 */
+		private String threadNameKey = "thread";
+
+		/**
+		 * Name of the class from which the asynchronous process started
```
adriancole: simple name? fully qualified? is the class the defining class of the async annotated method? if so, maybe say that and see methodNameKey

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/LazyTraceExecutor.java
```diff
@@ -47,7 +54,18 @@ public void execute(Runnable command) {
 				this.delegate.execute(command);
 			}
 		}
-		this.delegate.execute(new TraceRunnable(this.tracer, command));
+		// due to some race conditions trace keys might not be ready yet
```
adriancole: thx

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/TraceCallable.java
```diff
@@ -20,18 +20,39 @@
...
+ *     <ul>from the toString() of the delegate</ul>
+ * </li>
+ *
+ * @see org.springframework.cloud.sleuth.SpanName
+ *
  * @author Spencer Gibb
+ * @author Marcin Grzejszczak
  */
-public class TraceCallable<V> extends TraceDelegate<Callable<V>> implements Callable<V> {
+public class TraceCallable<V> implements Callable<V> {
```
adriancole: is there a subtype besides LocalComponent? (curious) or is this used directly?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/TraceContinuingCallable.java
```diff
@@ -0,0 +1,44 @@
...
+
+package org.springframework.cloud.sleuth.instrument.async;
+
+import java.util.concurrent.Callable;
+
+import org.springframework.cloud.sleuth.Span;
+import org.springframework.cloud.sleuth.Tracer;
+
+/**
+ * Trace Callable that continues a span instead of creating a new one
```
adriancole: great line

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/util/SpanNameRetrievalUtil.java
```diff
@@ -0,0 +1,38 @@
...
+/**
+ * Utility class that tries to get the Span name from
+ * SpanName annotation value if one is present. If that's not the case
+ * then it delegates to toString() method of the object
+ *
+ * @see org.springframework.cloud.sleuth.SpanName
+ *
+ * @author Marcin Grzejszczak
+ */
+public class SpanNameRetrievalUtil {
```
adriancole: hehe I bet you know already I don't like this name :)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/TraceCallable.java
```diff
@@ -45,4 +66,42 @@ public V call() throws Exception {
...
+	protected String getSpanName() {
+		String spanName = this.name != null ? this.name : SpanNameRetrievalUtil.getSpanName(this.delegate);
+		// If there is no overridden toString method we'll put constant value
+		if (isDefaultToString(spanName)) {
+			return "async";
+		}
+		return spanName;
+	}
+
+	private boolean isDefaultToString(String spanName) {
```
adriancole: should this other code go into the span retriever function?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/TraceCallable.java
```diff
@@ -45,4 +66,42 @@ public V call() throws Exception {
 		}
 	}
 
+	protected Span startSpan() {
+		return this.tracer.joinTrace(getSpanName(), this.parent);
+	}
+
+	protected String getSpanName() {
+		String spanName = this.name != null ? this.name : SpanNameRetrievalUtil.getSpanName(this.delegate);
```
adriancole: I'd collapse this logic into the span namer..

ex.

```java
if (this.name != null) return this.name;
return spanNamer.name(this.delegate, "async"); // last param is the fallback on null;
```

adriancole:
I'd collapse any single-child hierarchies where only the leaf is used (if that exists). For example, if only LocalSpanFooable is used (not its parent), squash the parent. Also, I think SpanNamer is a better name for what's being done, although (forgetting the language level) it could be a function interface vs a concrete type that's actually injected.

besides that, lgtm

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/TraceCallable.java
marcingrzejszczak: There is also the `TraceContinuingCallable`

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/SpanName.java
marcingrzejszczak: Come to think of it - a @Async method could be also annotated with @SpanName - but I'd leave that for later. For now let's assume that it's only classes?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/TraceCallable.java
marcingrzejszczak: yeah it should

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/SpanName.java
adriancole: constraints are fine with me.. what sort of classes would be annotated with
this? for example, if they are those that are passed to executor services
(runnables, etc). That's the thing we'd explain in javadoc and/or example,
test etc.

marcingrzejszczak: Fair enough - I'll add it to the javadoc. Good point

marcingrzejszczak:
Done in https://github.com/spring-cloud/spring-cloud-sleuth/commit/cf6f62966382a72fb4202ad4cf43414b01bcd433

PR #171, opened 2016-02-18
Updates to zipkin 0.5.3

adriancole:
This includes updates to the latest zipkin-web image and zipkin-java
dependency.

zipkin-web's notable as it does a lot more in javascript, ex it no
longer does server-side mustache template rendering. It also makes one
less call to the services endpoint when rendering the search page.

zipkin-java's in-memory span store is now in the core jar and also can
render the dependency tree. This means it is now feature-parity with
mysql (but still merely a test span-store). zipkin-java also now tests
that it can run without optional dependencies.


marcingrzejszczak:
:shipit:

PR #173, opened 2016-02-18
[#161] Fixed the way collections are passed between spans

### LINKED ISSUE #161: assymetry in requests from load-balanced feign vs load-balanced `RestTemplate`
i've got two endpoints on node A calling the same endpoint on node B:

```java

@FeignClient(serviceId = MessageClient.ZIPKIN_CLIENT_B)
interface MessageServiceClient {

	@RequestMapping(method = RequestMethod.GET, value = "/",
			consumes = MediaType.APPLICATION_JSON_VALUE)
	Map<String, String> readMessage();
}

@RestController
@RequestMapping("/message")
class MessageClientRestController {

	@Autowired
	private RestTemplate restTemplate;

	@Autowired
	private MessageServiceClient messageClient;

	@RequestMapping("/template")
	ResponseEntity<Map<String, String>> template() {

		String url = "http://" + MessageClient.ZIPKIN_CLIENT_B;

		ParameterizedTypeReference<Map<String, String>> ptr =
				new ParameterizedTypeReference<Map<String, String>>() {};

		ResponseEntity<Map<String, String>> responseEntity =
				this.restTemplate.exchange(url, HttpMethod.GET, null, ptr);

		return ResponseEntity
				.ok()
				.contentType(responseEntity.getHeaders().getContentType())
				.body(responseEntity.getBody());
	}

	@RequestMapping("/feign")
	Map<String, String> feign() {
		return this.messageClient.readMessage();
	}
}

``` 

when i run both paths i get more `span`s in the `RestTemplate` call than in the Feign call, which seem to me otherwise identical. Both are load-balanced.

screen shot 2016-02-13 at 5 37 58 pm
screen shot 2016-02-13 at 5 39 18 pm

marcingrzejszczak:
When a span is continued a new one is created from the previous one. We've been copying span collections thus the changes in the continued instance were not reflected in the parent.

Fixes #161

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
@@ -63,9 +63,9 @@ public class Span {
 	private final long spanId;
 	private boolean remote = false;
 	private boolean exportable = true;
-	private final Map<String, String> tags = new LinkedHashMap<>();
+	private final Map<String, String> tags;
 	private final String processId;
-	private final List<Log> logs = new ArrayList<>();
+	private final List<Log> logs;
 	private final Span savedSpan;
 
 	public Span(Span current, Span savedSpan) {
@@ -78,8 +78,8 @@ public class Span {
 		this.remote = current.isRemote();
 		this.exportable = current.isExportable();
 		this.processId = current.getProcessId();
-		this.tags.putAll(current.tags());
-		this.logs.addAll(current.logs());
+		this.tags = current.tags;
+		this.logs = current.logs;
 		this.savedSpan = savedSpan;
 	}
 
@@ -102,6 +102,8 @@ public class Span {
 		this.exportable = exportable;
 		this.processId = processId;
 		this.savedSpan = savedSpan;
+		this.tags = new LinkedHashMap<>();
+		this.logs = new ArrayList<>();
 	}
 
 	public static SpanBuilder builder() {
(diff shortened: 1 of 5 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
```diff
@@ -78,8 +78,8 @@ public Span(Span current, Span savedSpan) {
 		this.remote = current.isRemote();
 		this.exportable = current.isExportable();
 		this.processId = current.getProcessId();
-		this.tags.putAll(current.tags());
-		this.logs.addAll(current.logs());
+		this.tags = current.tags;
```
adriancole: I'm worried this might get un-fixed at some point.

I'd put in a test case that proves this reference is shared and/or comment why we are sharing it (ex note the use case, even if it takes two lines).

adriancole:
very subtle! I don't quite understand the lifecycle of these collections, but LGTM once you've defended the change with a comment and/or test case.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
marcingrzejszczak: I've noticed that I've forgotten to add a test - good catch. I'll also add a javadoc to the constructor

marcingrzejszczak:
Done in https://github.com/spring-cloud/spring-cloud-sleuth/commit/785dab0908c6466b6934b33067b046d2dfa5220c

PR #177, opened 2016-02-19
[#159] Tagging hystrix events

### LINKED ISSUE #159: We should add events for Hystrix as we have in HTTP
Like in `TraceKeys.Http`

marcingrzejszczak:
when a hystrix command was explicitly created we can tag the following values
* command key
* command group
* thread pool key

fixes #159

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/TraceKeys.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/TraceKeys.java
@@ -318,6 +318,21 @@ public class TraceKeys {
 		 */
 		private String commandKey = "commandKey";
 
+		/**
+		 * Name of the command group
+		 */
+		private String commandGroup = "commandGroup";
+
+		/**
+		 * Name of the cache key
+		 */
+		private String cacheKey = "cacheKey";
+
+		/**
+		 * Name of the thread pool key
+		 */
+		private String threadPoolKey = "threadPoolKey";
+
 		public String getPrefix() {
 			return this.prefix;
 		}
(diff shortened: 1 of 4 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/TraceKeys.java
```diff
@@ -318,6 +318,21 @@ public void setHeaders(Collection<String> headers) {
 		 */
 		private String commandKey = "commandKey";
 
+		/**
+		 * Name of the command group
```
adriancole: add link

adriancole: or `@see`

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/TraceKeys.java
```diff
@@ -318,6 +318,21 @@ public void setHeaders(Collection<String> headers) {
...
+		 */
+		private String commandGroup = "commandGroup";
+
+		/**
+		 * Name of the cache key
+		 */
+		private String cacheKey = "cacheKey";
```
adriancole: trick with tags is adding the minimum, so that it doesn't expand the spans with unused data. do you expect people to look up by cache key or thread pool key? What if we didn't define these now? (ex instead wait for someone to ask for them). If you look at the http keys, you'll notice documentation about why the keys are important.. prefer less, better documented keys. Ex. command key is used for X, ex you can re-use this in a turbine query like so, to investigate further.

adriancole:
basic feedback is that I think we should only add things that become helpful in latency troubleshooting in concrete ways.

Have a look at the javadocs on Http keys. I really think a higher quality few keys are a better choice. Keys are easy to add later.

```
		/**
		 * The entire URL, including the scheme, host and query parameters if available.
		 * Ex.
		 * "https://mybucket.s3.amazonaws.com/objects/abcd-ff?X-Amz-Algorithm=AWS4-HMAC-SHA256&X-Amz-Algorithm=AWS4-HMAC-SHA256..."
		 * Combined with {@link #method}, you can understand the fully-qualified
		 * request line. This is optional as it may include private data or be of
		 * considerable length.
		 */
		private String url = "http.url";
```

adriancole:
here's an old comment about some of the relationships of hystrix keys. This, people + docs might be a good place to mine advisory javadocs on keys we feel are important enough to add.

https://github.com/Netflix/feign/issues/189#issuecomment-95688652

marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-sleuth/commit/3eca10bcc7af10387014b8b9b35b84c838390f1f

PR #181, opened 2016-02-24
Refactors HTTP SpanReporter tests to use ZipkinRule

adriancole:
This replaces direct usage of MockWebServer with ZipkinRule.

See https://github.com/openzipkin/zipkin-java/pull/83


marcingrzejszczak:
:shipit:

PR #185, opened 2016-02-25
Updates to latest zipkin, adding spring.zipkin.compression.enabled flag

adriancole:
This updates to latest zipkin, adding a flag to gzip spans before
posting them. Note this is set to false by default as it requires servers
to also be latest. We can flip the default once we can assume that.
    
Ex. `spring.zipkin.compression.enabled = true`

### DIFF
--- a/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinProperties.java
+++ b/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinProperties.java
@@ -31,6 +31,7 @@ public class ZipkinProperties {
 	private String baseUrl = "http://localhost:9411/";
 	private boolean enabled = true;
 	private int flushInterval = 1;
+	private Compression compression = new Compression();
 
 	public String getBaseUrl() {
 		return this.baseUrl;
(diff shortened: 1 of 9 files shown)


spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinProperties.java
```diff
@@ -27,6 +27,7 @@
 	private String baseUrl = "http://localhost:9411/";
 	private boolean enabled = true;
 	private int flushInterval = 1;
+	private boolean gzip = false;
```
adriancole: @dsyer @marcingrzejszczak is this a decent property name? or should be call it compress or something else?

marcingrzejszczak: Theoretically there can be different types of compressions, right? Maybe it should be an enum compression that defaults to NONE, but can be equal to GZIP. The question is if we support via Mvc any other types of compression - I don't really know.

adriancole: per discussion, following convention of boot: `compression.enabled`, so.. `spring.zipkin.compression.enabled = true`

PR #186, opened 2016-02-25
[#184] Stopped overriding tags via messaging

### LINKED ISSUE #184: Via messaging don't override the header values
We have to check the headers for existence of an entry. We shouldn't override the existing values.

org.springframework.cloud.sleuth.instrument.messaging.SpanMessageHeaders

marcingrzejszczak:
fixes #184

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/AbstractTraceChannelInterceptor.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/AbstractTraceChannelInterceptor.java
@@ -20,6 +20,9 @@ import org.springframework.util.ClassUtils;
  */
 abstract class AbstractTraceChannelInterceptor extends ChannelInterceptorAdapter implements ExecutorChannelInterceptor {
 
+	/**
+	 * A default prefix for span name for discerning messaging origin of the span
+	 */
 	protected static final String MESSAGE_COMPONENT = "message";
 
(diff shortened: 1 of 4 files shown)


marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-sleuth/commit/6948d92d134b7eab986c9faa51c5081e8b965c8a

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/AbstractTraceChannelInterceptor.java
```diff
@@ -20,6 +20,9 @@
  */
 abstract class AbstractTraceChannelInterceptor extends ChannelInterceptorAdapter implements ExecutorChannelInterceptor {
 
+	/**
+	 * A default prefix for span name for discerning messaging origin of the span
```
adriancole: I have no idea what this means..  "discerning messaging origin of the span" 
do you mean that spans with this naming prefix came from messaging components? If so, maybe reinforce with an example? Because in this change it looks like this is a tag prefix, not a span name prefix, or maybe it is both?

marcingrzejszczak: Fixed in https://github.com/spring-cloud/spring-cloud-sleuth/commit/7c7043910caad26e863fca7a820c6aefb2d6934d

Issue #191, opened 2016-02-29
ServiceName missing for certain spans


marcingrzejszczak:
For example:

![duration_again2](https://cloud.githubusercontent.com/assets/3297437/13401049/de8c2bde-df0a-11e5-9821-a7eb1fe4a8a7.png)

PR #195, opened 2016-03-01
Issues #191 missing span name

marcingrzejszczak:
@dsyer @adriancole @spencergibb 

* added local component annotation if a span doesn't have any Zipkin constants
* added SA annotation if a span has a client related annotation

### DIFF
--- a/spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIterator.java
+++ b/spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIterator.java
@@ -15,7 +15,9 @@
  */
 package org.springframework.cloud.sleuth.zipkin.stream;
 
+import java.util.Arrays;
 import java.util.Iterator;
+import java.util.List;
 import java.util.NoSuchElementException;
 
 import org.apache.commons.logging.Log;
@@ -38,6 +40,14 @@ import zipkin.Span.Builder;
  * @since 1.0.0
  */
 final class SamplingZipkinSpanIterator implements Iterator<zipkin.Span> {
+	private static final List<String> ZIPKIN_ANNOTATIONS = Arrays.asList(
+			Constants.CLIENT_ADDR, Constants.CLIENT_RECV, Constants.CLIENT_SEND,
+			Constants.CLIENT_RECV_FRAGMENT, Constants.CLIENT_SEND_FRAGMENT,
+			Constants.SERVER_ADDR, Constants.SERVER_RECV, Constants.SERVER_SEND,
+			Constants.SERVER_RECV_FRAGMENT, Constants.SERVER_SEND_FRAGMENT,
+			Constants.LOCAL_COMPONENT,
+			Constants.WIRE_RECV, Constants.WIRE_SEND
+	);
 
 	private static final Log log = org.apache.commons.logging.LogFactory
 			.getLog(SamplingZipkinSpanIterator.class);
@@ -110,17 +120,15 @@ final class SamplingZipkinSpanIterator implements Iterator<zipkin.Span> {
 		// A zipkin span without any annotations cannot be queried, add special "lc" to
 		// avoid that.
 		if (span.logs().isEmpty() && span.tags().isEmpty()) {
-			String processId = span.getProcessId() != null
-					? span.getProcessId().toLowerCase()
-					: ZipkinMessageListener.UNKNOWN_PROCESS_ID;
-			zipkinSpan.addBinaryAnnotation(
-					BinaryAnnotation.create(Constants.LOCAL_COMPONENT, processId, ep));
+			addLocalComponentAnnotation(span, zipkinSpan, ep);
 		}
 		else {
 			ZipkinMessageListener.addZipkinAnnotations(zipkinSpan, span, ep);
 			ZipkinMessageListener.addZipkinBinaryAnnotations(zipkinSpan, span, ep);
 		}
-
+		if (!spanContainsAnyZipkinConstant(span)) {
+			addLocalComponentAnnotation(span, zipkinSpan, ep);
+		}
 		zipkinSpan.timestamp(span.getBegin() * 1000);
 		zipkinSpan.duration(span.getAccumulatedMillis() * 1000);
 		zipkinSpan.traceId(span.getTraceId());
@@ -137,5 +145,23 @@ final class SamplingZipkinSpanIterator implements Iterator<zipkin.Span> {
 			zipkinSpan.name(span.getName());
 		}
 		return zipkinSpan.build();
+	}
+
+	private static void addLocalComponentAnnotation(Span span, Builder zipkinSpan,
+			Endpoint ep) {
+		String processId = span.getProcessId() != null
+				? span.getProcessId().toLowerCase()
+				: ZipkinMessageListener.UNKNOWN_PROCESS_ID;
+		zipkinSpan.addBinaryAnnotation(
+				BinaryAnnotation.create(Constants.LOCAL_COMPONENT, processId, ep));
+	}
+
+	private static boolean spanContainsAnyZipkinConstant(Span span) {
+		for (org.springframework.cloud.sleuth.Log log : span.logs()) {
+			if (ZIPKIN_ANNOTATIONS.contains(log.getEvent())) {
+				return true;
+			}
+		}
+		return false;
 	}
 }
--- a/spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIterator.java
+++ b/spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIterator.java
@@ -15,6 +15,7 @@
  */
 package org.springframework.cloud.sleuth.zipkin.stream;
 
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
@@ -40,14 +41,23 @@ import zipkin.Span.Builder;
  * @since 1.0.0
  */
 final class SamplingZipkinSpanIterator implements Iterator<zipkin.Span> {
-	private static final List<String> ZIPKIN_ANNOTATIONS = Arrays.asList(
+	private static final List<String> ZIPKIN_CLIENT_ANNOTATIONS = Arrays.asList(
 			Constants.CLIENT_ADDR, Constants.CLIENT_RECV, Constants.CLIENT_SEND,
-			Constants.CLIENT_RECV_FRAGMENT, Constants.CLIENT_SEND_FRAGMENT,
-			Constants.SERVER_ADDR, Constants.SERVER_RECV, Constants.SERVER_SEND,
-			Constants.SERVER_RECV_FRAGMENT, Constants.SERVER_SEND_FRAGMENT,
-			Constants.LOCAL_COMPONENT,
-			Constants.WIRE_RECV, Constants.WIRE_SEND
+			Constants.CLIENT_RECV_FRAGMENT, Constants.CLIENT_SEND_FRAGMENT
 	);
+	private static final List<String> ZIPKIN_ANNOTATIONS = zipkinAnnotations();
+
+	private static List<String> zipkinAnnotations() {
+		List<String> annotations = new ArrayList<>();
+		annotations.addAll(Arrays.asList(
+				Constants.SERVER_ADDR, Constants.SERVER_RECV, Constants.SERVER_SEND,
+				Constants.SERVER_RECV_FRAGMENT, Constants.SERVER_SEND_FRAGMENT,
+				Constants.LOCAL_COMPONENT,
+				Constants.WIRE_RECV, Constants.WIRE_SEND
+		));
+		annotations.addAll(ZIPKIN_CLIENT_ANNOTATIONS);
+		return annotations;
+	}
 
 	private static final Log log = org.apache.commons.logging.LogFactory
 			.getLog(SamplingZipkinSpanIterator.class);
@@ -128,6 +138,8 @@ final class SamplingZipkinSpanIterator implements Iterator<zipkin.Span> {
 		}
 		if (!spanContainsAnyZipkinConstant(span)) {
 			addLocalComponentAnnotation(span, zipkinSpan, ep);
+		} else if (spanContainsAnyZipkinClientConstantAndSaIsNotSet(span)) {
+			addServerAddressAnnotation(zipkinSpan, ep);
 		}
 		zipkinSpan.timestamp(span.getBegin() * 1000);
 		zipkinSpan.duration(span.getAccumulatedMillis() * 1000);
@@ -156,6 +168,16 @@ final class SamplingZipkinSpanIterator implements Iterator<zipkin.Span> {
 				BinaryAnnotation.create(Constants.LOCAL_COMPONENT, processId, ep));
 	}
 
+	private static void addServerAddressAnnotation(zipkin.Span.Builder zipkinSpan,
+			Endpoint ep) {
+		BinaryAnnotation component = new BinaryAnnotation.Builder()
+				.type(BinaryAnnotation.Type.STRING)
+				.key(Constants.SERVER_ADDR)
+				.value(ep.serviceName)
+				.endpoint(ep).build();
+		zipkinSpan.addBinaryAnnotation(component);
+	}
+
 	private static boolean spanContainsAnyZipkinConstant(Span span) {
 		for (org.springframework.cloud.sleuth.Log log : span.logs()) {
 			if (ZIPKIN_ANNOTATIONS.contains(log.getEvent())) {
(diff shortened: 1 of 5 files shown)


spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIterator.java
```diff
@@ -138,4 +158,45 @@ public void remove() {
...
+			}
+		}
+		return false;
+	}
+
+	private static boolean spanContainsAnyZipkinClientConstantAndSaIsNotSet(Span span) {
+		boolean containsAnyZipkinClientConstant = false;
+		for (org.springframework.cloud.sleuth.Log log : span.logs()) {
+			if (ZIPKIN_CLIENT_ANNOTATIONS.contains(log.getEvent())) {
+				containsAnyZipkinClientConstant = true;
```
adriancole: I think logically it would be the same to `return !span.tags().containsKey(Constants.SERVER_ADDR)` here instead of break

spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIterator.java
```diff
@@ -138,4 +158,45 @@ public void remove() {
...
+		String processId = span.getProcessId() != null
+				? span.getProcessId().toLowerCase()
+				: ZipkinMessageListener.UNKNOWN_PROCESS_ID;
+		zipkinSpan.addBinaryAnnotation(
+				BinaryAnnotation.create(Constants.LOCAL_COMPONENT, processId, ep));
+	}
+
+	private static void addServerAddressAnnotation(zipkin.Span.Builder zipkinSpan,
+			Endpoint ep) {
+		BinaryAnnotation component = new BinaryAnnotation.Builder()
```
adriancole: `BinaryAnnoation.address(String key, Endpoint endpoint) ` saves you from defining this method

spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIterator.java
```diff
@@ -110,17 +130,17 @@ public void remove() {
 		// A zipkin span without any annotations cannot be queried, add special "lc" to
 		// avoid that.
 		if (span.logs().isEmpty() && span.tags().isEmpty()) {
```
adriancole: I'd replace this test with `!spanContainsAnyZipkinConstant(span)` to reduce the if statements.

adriancole: actually.. just get rid of this "if" statement, and move the comment to your new if statement

spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIterator.java
```diff
@@ -38,6 +41,23 @@
  * @since 1.0.0
  */
 final class SamplingZipkinSpanIterator implements Iterator<zipkin.Span> {
+	private static final List<String> ZIPKIN_CLIENT_ANNOTATIONS = Arrays.asList(
```
adriancole: this can be made a lot simpler.. just look for CLIENT_SEND. A span that has CLIENT_RECV_FRAGMENT, but not CLIENT_SEND is a bug. Also CLIENT_ADDR shouldn't be here (this is on a server span to indicate its caller)

adriancole: ps I understand why you did this, just that the logic in zipkin is a little overly cautious :P

spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIterator.java
```diff
@@ -38,6 +41,23 @@
...
+			Constants.CLIENT_ADDR, Constants.CLIENT_RECV, Constants.CLIENT_SEND,
+			Constants.CLIENT_RECV_FRAGMENT, Constants.CLIENT_SEND_FRAGMENT
+	);
+	private static final List<String> ZIPKIN_ANNOTATIONS = zipkinAnnotations();
+
+	private static List<String> zipkinAnnotations() {
```
adriancole: the most key annotations are CLIENT_SEND, SERVER_RECEIVE, CLIENT_ADDR, and SERVER_ADDR.  Since we control the instrumentation, we can focus on these.. ex we wouldn't send CLIENT_RECEIVE without CLIENT_SEND.

Bear in mind that Constants.LOCAL_COMPONENT is a tag/binaryAnnotation, so be careful with contains checks.

spring-cloud-sleuth-zipkin-stream/src/test/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIteratorTests.java
```diff
@@ -76,6 +79,54 @@ public boolean isSampled(long l) {
...
+		assertThat(result)
+				.flatExtracting(s -> s.binaryAnnotations)
+				.extracting(input -> input.key)
+				.contains(Constants.LOCAL_COMPONENT);
+	}
+
+	@Test
+	public void appendsServerAddressTagIfClientLogIsPresent() {
+		Span span = span("foo");
+		span.logEvent(Constants.CLIENT_RECV);
```
adriancole: switch all the tests s/CLIENT_RECV/CLIENT_SEND as it makes more sense (because it would be an error to have only a RECV annotation)

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinSpanListener.java
```diff
@@ -44,6 +47,24 @@
  * @since 1.0.0
  */
 public class ZipkinSpanListener {
```
adriancole: If stream depends on this, make a utility and share it.. otherwise, I understand why c&p, just carry over the same comments here.

spring-cloud-sleuth-zipkin/src/test/java/org/springframework/cloud/sleuth/zipkin/ZipkinSpanListenerTests.java
```diff
@@ -135,6 +139,45 @@ public void rpcAnnotations() {
...
+		assertThat(result.binaryAnnotations)
+				.filteredOn("key", Constants.SERVER_ADDR)
+				.extracting(input -> input.value)
+				.containsOnly("unknown".getBytes(Charset.forName("UTF-8")));
+	}
+
+	@Test
+	public void shouldReuseServerAddressTag() {
+		this.parent.logEvent(Constants.CLIENT_SEND);
+		this.parent.tag(Constants.SERVER_ADDR, "fooservice");
```
adriancole: I understand why you are re-using this, but it doesn't make as much sense here as it looks like. the value of SERVER_ADDR is a struct that includes the IP address. I'd use the tag "peer.service" from OpenTracing in sleuth and in the zipkin converter, make that either a CLIENT_ADDR or SERVER_ADDR instead. https://github.com/opentracing/opentracing-go/blob/master/ext/tags.go#L21

In the zipkin v2, I'd expect to process "peer.service" directly. I've already started playing with this here: https://github.com/openzipkin/zipkin-java/blob/master/zipkin/src/main/java/zipkin/internal/DependencyLinkSpan.java

spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIterator.java
```diff
@@ -110,17 +130,17 @@ public void remove() {
...
+			addLocalComponentAnnotation(span, zipkinSpan, ep);
 		}
 		else {
 			ZipkinMessageListener.addZipkinAnnotations(zipkinSpan, span, ep);
 			ZipkinMessageListener.addZipkinBinaryAnnotations(zipkinSpan, span, ep);
 		}
-
+		if (!spanContainsAnyZipkinConstant(span)) {
+			addLocalComponentAnnotation(span, zipkinSpan, ep);
+		} else if (spanContainsAnyZipkinClientConstantAndSaIsNotSet(span)) {
```
adriancole: add a comment here why we are adding the "server address".

The guidance from the docs is "When logging CLIENT_SEND, instrumentation should also log the SERVER_ADDR", so for example you can link to that: https://github.com/openzipkin/zipkin-java/blob/master/zipkin/src/main/java/zipkin/Constants.java#L28

adriancole:
most notes are about how to pare down logic, otherwise LG

spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIterator.java
marcingrzejszczak: :+1:

spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIterator.java
marcingrzejszczak: Good to know!

spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIterator.java
marcingrzejszczak: :+1:

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinSpanListener.java
marcingrzejszczak: AFAIR they don't depend on each other - that's why there's quite a few c&p

spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIterator.java
marcingrzejszczak: it makes little sense to move the comment to that method actually IMO. I'd leave it where it is

spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIterator.java
marcingrzejszczak: Hmm but this `BinaryAnnotation.address` allows only to take info from `Endpoint`. If one explicitly provides the service name from `peer.service` it will be ignored

spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIterator.java
marcingrzejszczak: :+1:

spring-cloud-sleuth-zipkin-stream/src/test/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIteratorTests.java
marcingrzejszczak: :+1:

spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIterator.java
```diff
@@ -38,6 +41,11 @@
  * @since 1.0.0
  */
 final class SamplingZipkinSpanIterator implements Iterator<zipkin.Span> {
+	private static final List<String> ZIPKIN_ANNOTATIONS = Arrays.asList(
```
adriancole: I don't think the _ADDR things need to be here, as I don't think a sleuth span would add these as tags, right? I think Span.SPAN_PEER_SERVICE_TAG_NAME is what's being tagged instead now

adriancole: similarly LOCAL_COMPONENT is a tag, not an annotation (log)

adriancole: basically, I'd probably just call this `ZIPKIN_START_EVENTS = Arrays.asList(CLIENT_SEND, SERVER_RECV);` effectively these are the events that indicate it is an RPC span, and also whether or not an _ADDR binary annotation should be logged.

spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIterator.java
```diff
@@ -138,4 +146,43 @@ public void remove() {
 		}
 		return zipkinSpan.build();
 	}
+
+	private static void addLocalComponentAnnotation(Span span, Builder zipkinSpan,
+			Endpoint ep) {
+		if (span.tags().containsKey(Constants.LOCAL_COMPONENT)) {
```
adriancole: nice early out

spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIterator.java
```diff
@@ -138,4 +146,43 @@ public void remove() {
 		}
 		return zipkinSpan.build();
 	}
+
+	private static void addLocalComponentAnnotation(Span span, Builder zipkinSpan,
```
adriancole: nit: `ensureLocalComponent` as that allows you to completely dodge Annotation vs BinaryAnnotation (latter being what's in question here)

spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIterator.java
```diff
@@ -138,4 +146,43 @@ public void remove() {
...
+	private static boolean spanContainsAnyZipkinConstant(Span span) {
+		for (org.springframework.cloud.sleuth.Log log : span.logs()) {
+			if (ZIPKIN_ANNOTATIONS.contains(log.getEvent())) {
+				return true;
+			}
+		}
+		return false;
+	}
+
+	private static boolean spanContainsAnyZipkinClientConstantAndSaIsNotSet(Span span) {
```
adriancole: bad name :P .. I think the flow you want is..

```java
if (hasClientSend(span)) {
  ensureServerAddr(span);
}

```

spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIterator.java
marcingrzejszczak: :+1:

spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/SamplingZipkinSpanIterator.java
```diff
@@ -109,18 +121,14 @@ public void remove() {
 
 		// A zipkin span without any annotations cannot be queried, add special "lc" to
 		// avoid that.
-		if (span.logs().isEmpty() && span.tags().isEmpty()) {
-			String processId = span.getProcessId() != null
-					? span.getProcessId().toLowerCase()
-					: ZipkinMessageListener.UNKNOWN_PROCESS_ID;
-			zipkinSpan.addBinaryAnnotation(
-					BinaryAnnotation.create(Constants.LOCAL_COMPONENT, processId, ep));
+		if (!spanContainsAnyZipkinConstant(span)) {
```
adriancole:
```java
if (notClientOrServer(span)) {
  ensureLocalComponent(span);
}
```

Issue #79, opened 2015-12-24
Update documentation to reflect recent changes


marcingrzejszczak:

marcingrzejszczak:
@adriancole: so in your work, I'd probably put in a comment or doc the "style" being used. Ex. single-host vs client+server span .. commentary about how the UI presents single-host spans is here (feel free to add) openzipkin/zipkin#963

PR #197, opened 2016-03-02
Issues #79 updating documentation

marcingrzejszczak:
Feedback is more than welcome! What else can we put here?


docs/src/main/asciidoc/README.adoc
```diff
@@ -49,6 +49,11 @@ The UI should look like the screenshot above.
 
 include::https://raw.githubusercontent.com/spring-cloud/spring-cloud-build/master/docs/src/main/asciidoc/building.adoc[]
 
+IMPORTANT: There are 2 different versions of JDK used in Spring Cloud Sleuth. JDK7 is used for main sources and
```
adriancole: is this true? I think only one JDK is used. The comment is more about language level than JDK, right?

docs/src/main/asciidoc/features.adoc
```diff
@@ -8,18 +8,30 @@
 2016-02-02 15:31:01.936  INFO [bar,46ab0d418373cbc9,46ab0d418373cbc9,false] 23030 --- [nio-8081-exec-4] ...
 ----
 +
-(notice the `[appname,traceId,spanId,exportable]` entries from the MDC).
+(notice the `[appname,traceId,spanId,exportable]` entries from the MDC. `appname`, `traceId` , `spanId` seem to be quite
```
adriancole: if it isn't too much trouble, just enumerate all three vs assuming people know what they are. Particularly appname has been hard to pin down.

* spanId - the id of a specific operation that took place
* appname - the name of the application that logged the span
* traceId - the id of the latency graph that contains the span

docs/src/main/asciidoc/features.adoc
```diff
@@ -8,18 +8,30 @@
...
+straightforward, `exportable` means whether the log should be exported to Zipkin or not).
 
-* Optionally log span data in JSON format for harvesting in a log aggregator (set `spring.sleuth.log.json.enabled=true`).
+* Provides an abstraction over common distributed tracing data models: traces, spans (forming a DAG), annotations,
+key-value annotations. Loosely based on HTrace, but Zipkin (Dapper) compatible.
 
-* Provides an abstraction over common distributed tracing data models: traces, spans (forming a DAG), annotations, key-value annotations. Loosely based on HTrace, but Zipkin (Dapper) compatible.
+* Instruments common ingress and egress points from Spring applications (servlet filter, async endpoints,
```
adriancole: good

docs/src/main/asciidoc/features.adoc
```diff
@@ -8,18 +8,30 @@
 2016-02-02 15:31:01.936  INFO [bar,46ab0d418373cbc9,46ab0d418373cbc9,false] 23030 --- [nio-8081-exec-4] ...
 ----
 +
-(notice the `[appname,traceId,spanId,exportable]` entries from the MDC).
+(notice the `[appname,traceId,spanId,exportable]` entries from the MDC. `appname`, `traceId` , `spanId` seem to be quite
+straightforward, `exportable` means whether the log should be exported to Zipkin or not).
 
-* Optionally log span data in JSON format for harvesting in a log aggregator (set `spring.sleuth.log.json.enabled=true`).
+* Provides an abstraction over common distributed tracing data models: traces, spans (forming a DAG), annotations,
```
adriancole: Not sure if it goes here, but here's a stab..

Sleuth records timing information to aid in latency analysis. Using sleuth, you can pinpoint causes of latency in your applications. Sleuth is written to not log too much, and to not cause your production application to crash.
* propagates structural data about your call-graph in-band, and the rest out-of-band.
* includes opinionated instrumentation of layers such as HTTP 
* includes sampling policy to manage volume
* can report to a Zipkin system for query and visualization

docs/src/main/asciidoc/features.adoc
```diff
@@ -8,18 +8,30 @@
...
 
-* Provides an abstraction over common distributed tracing data models: traces, spans (forming a DAG), annotations, key-value annotations. Loosely based on HTrace, but Zipkin (Dapper) compatible.
+* Instruments common ingress and egress points from Spring applications (servlet filter, async endpoints,
+rest template, scheduled actions, message channels, zuul filters, feign client).
 
-* Instruments common ingress and egress points from Spring applications (servlet filter, rest template, scheduled actions, message channels, zuul filters, feign client).
+* Provides simple metrics of accepted / dropped spans.
 
-* If `spring-cloud-sleuth-zipkin` then the app will generate and collect Zipkin-compatible traces (using Brave). By default it sends them via HTTP to a Zipkin server on localhost (port 9411). Configure the location of the service using `spring.zipkin.[host,port]`.
+* If `spring-cloud-sleuth-zipkin` then the app will generate and collect Zipkin-compatible traces (using Brave).
```
adriancole: take out brave, since it doesn't use brave anymore iirc

docs/src/main/asciidoc/features.adoc
```diff
@@ -8,18 +8,30 @@
...
+By default it sends them via HTTP to a Zipkin server on localhost (port 9411).
+Configure the location of the service using `spring.zipkin.baseUrl`.
 
-* If `spring-cloud-sleuth-stream` then the app will generate and collect traces via Spring Cloud Stream. Your app automatically becomes a producer of tracer messages that are sent over your broker of choice (e.g. RabbitMQ, Apache Kafka, Redis).
+* If `spring-cloud-sleuth-stream` then the app will generate and collect traces via Spring Cloud Stream. Your app
```
adriancole: link to spring cloud stream. maybe say an example stream could be rabbitmq

adriancole: whoops already did that

docs/src/main/asciidoc/intro.adoc
```diff
@@ -6,10 +6,74 @@ Spring Cloud Sleuth implements a distributed tracing solution for http://cloud.s
...
+span, you must stop it at some point in the future.
 
-*Trace:* A set of spans forming a tree-like structure.  For example, if you are running a distributed big-data store, a trace might be formed by a put request.
+*Trace:* A set of spans forming a tree-like structure.  For example, if you are running a distributed
+big-data store, a trace might be formed by a put request.
 
+*Annotation:*  is used to record existence of an event in time. Some of the core annotations used to define
+the start and stop of a request are:
 
+    - *cs* - Client Sent - The client has made a request. This annotation depicts the start of the span.
```
adriancole: good stuff

docs/src/main/asciidoc/intro.adoc
```diff
@@ -6,10 +6,74 @@ Spring Cloud Sleuth implements a distributed tracing solution for http://cloud.s
...
+
+image::zipkin-ui.png[Traces Info propagation]
+
+NOTE: When picking a particular trace you will see merged spans. That means that if there were 2 spans sent to
+Zipkin with Server Received and Server Sent annotations then they will presented as a single span.
+
+You don't have to use Zipkin to profit from tracing. When grepping the logs of those four applications by trace id
+equal to e.g. `2485ec27856c56f4` one would get the following:
+
+==== Distributed tracing with log collectors
```
adriancole: log correlation? I think this is probably more to the point

docs/src/main/asciidoc/intro.adoc
```diff
@@ -6,10 +6,74 @@ Spring Cloud Sleuth implements a distributed tracing solution for http://cloud.s
...
+image::zipkin-traces.png[Traces]
+
+However if you pick a particular trace then you will see *7 spans*:
+
+image::zipkin-ui.png[Traces Info propagation]
+
+NOTE: When picking a particular trace you will see merged spans. That means that if there were 2 spans sent to
+Zipkin with Server Received and Server Sent annotations then they will presented as a single span.
+
+You don't have to use Zipkin to profit from tracing. When grepping the logs of those four applications by trace id
```
adriancole: this comment should be moved to the next system, and I wouldn't necessarily suggest this is profitable :) rather that if you aren't running a zipkin system, you can still perform log correlation.

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
```diff
@@ -60,6 +57,123 @@ NOTE: Remember that tags are only collected and exported if there is a
 danger of accidentally collecting too much data without configuring
 something).
 
+NOTE: Currently the instrumentation in Spring Cloud Sleuth is eager - it means that
```
adriancole: I think it means more than this, correct? basically timing events are captured even when sleuth isn't exporting data to a tracing system.

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
```diff
@@ -60,6 +57,123 @@ NOTE: Remember that tags are only collected and exported if there is a
...
+we're actively trying to pass the tracing context between threads.
+This approach may change in the future towards being lazy on this matter.
+
+== Span lifecycle
+
+You can do the following operations on the Span by means of *Tracer* interface:
+- *start* - when you start a span its name is assigned and start timestamp is recorded.
+- *close* - the span gets *stopped* (the end time of the span is recorded) and if
```
adriancole: close, stopped and finished! fwiw OpenTracing uses one word: finish for all this.

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
```diff
@@ -60,6 +57,123 @@ NOTE: Remember that tags are only collected and exported if there is a
...
+the span is *exportable* then it will be eligible for collection to Zipkin.
+The span is also removed from the current thread.
+- *continue* - a new instance of span will be created whereas it will be a copy of the
```
adriancole: this sentence is confusing. and the example is also confusing as it doesn't say why I wouldn't just use the reference of spanToContinue

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
```diff
@@ -60,6 +57,123 @@ NOTE: Remember that tags are only collected and exported if there is a
...
+one that it continues. In other words:
+[source,java]
+----
+Span continuedSpan = this.tracer.continueSpan(spanToContinue);
+assertThat(continuedSpan).isEqualTo(spanToContinue);
+----
+- *detach* - the span doesn't get stopped or closed. It only gets removed from the current thread.
```
adriancole: aside, but in finagle this is called clear() (actually Trace.letClear). Edge-case things like this and continue should have an example of where someone would use them. If they aren't general purpose, then they could just be documented in javadoc and not here.

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
```diff
@@ -60,6 +57,123 @@ NOTE: Remember that tags are only collected and exported if there is a
...
+- *join* - you can create a new span and set an explicit parent to it
+
+=== Creating spans
```
adriancole: thing about docs more about stories vs operation names. Yes, we are creating a span, but maybe use a real-world example. Like something someone would actually do... ex time a critical operation

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
```diff
@@ -60,6 +57,123 @@ NOTE: Remember that tags are only collected and exported if there is a
...
+
+[source,java]
+----
+include::../../../../spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/documentation/SpringCloudSleuthDocTests.java[tags=manual_span_creation,indent=0]
+----
+
+In this example we could see how to create a new instance of span. Assuming that there already
+was a span present in this thread then it would become the parent of that span.
+
+=== Continuing spans
```
adriancole: maybe there can be a link to this when introducing continue?

docs/src/main/asciidoc/README.adoc
marcingrzejszczak: yup

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
```diff
@@ -60,6 +57,123 @@ NOTE: Remember that tags are only collected and exported if there is a
...
+    only a technical implementation detail that you wouldn't necessarily want to reflect in tracing as a separate being.
+
+To continue a span you can use the *Tracer* interface.
+
+[source,java]
+----
+include::../../../../spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/documentation/SpringCloudSleuthDocTests.java[tags=manual_span_continuation,indent=0]
+----
+
+What's important is to always clean after you create a span. The golden rule would be to:
```
adriancole: should this note be in the continue span section?

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
```diff
@@ -60,6 +57,123 @@ NOTE: Remember that tags are only collected and exported if there is a
...
+----
+
+IMPORTANT: After having joined the span remember to close it. Otherwise you will see a lot of warnings in your logs
+ related to the fact that you have a span present in the current thread other than the one you're trying to close.
+ What's worse your spans won't get closed properly thus will not get collected to Zipkin.
+
+== Naming spans
+
+Picking a span name is not a trivial task. Span name should depict an operation
+but shouldn't be too long (you don't want the Zipkin UI to render for minutes,
```
adriancole: this isn't about length, this is about cardinality. operation names should be low cardinality, ex not include identifiers.

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
marcingrzejszczak: hmm these are the methods from `Span` class. I can add a note that OpenTracing uses finish

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
marcingrzejszczak: a better example is presented below

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
adriancole: I meant that the description of close suggests it needs a different name.

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
marcingrzejszczak: This section is just an introduction. As you can see in the subsequent points I'm elaborating more on this

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
marcingrzejszczak: the same response as above

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
marcingrzejszczak: :+1:

marcingrzejszczak:
Done in https://github.com/spring-cloud/spring-cloud-sleuth/commit/4b61439ab98aec9009f0047acf7465adc337c2fb

PR #200, opened 2016-03-04
Simplified Tracer API

marcingrzejszczak:


adriancole:
+1 This is analogous to OpenTracing's `parent.start_child(operation_name)`

I like this a lot as it dodges the word join, which has a completely different meaning in OpenTracing (join is for an existing span that crossed a process boundary, ex joining from http headers)
```
span = tracer.join(
    operation_name=operation,
    format=opentracing.HTTP_HEADER_FORMAT,
    carrier=request.headers
)
```

Issue #198, opened 2016-03-03
With latest AbstractTraceHttpRequestInterceptor, spans cannot be closed when exceptions occur while sending HTTP request


dnathanson:
In `TraceRestTemplateInterceptor`, `publishStartEvent()` is called before sending the HTTP request.  `publishStartEvent()` now creates a child span and sets it as current span on thread.  The child span is expected to be closed when `TraceHttpResponse.close()` is called.  But if an exception happens in `ClientHttpRequestExecution.execute()`, a response object is not returned and so the span is never closed.
This causes an error when `Tracer` tried to close or detach the current span (in my case invoked from `SleuthHystrixConcurrencyStrategy`.

```
o.s.cloud.sleuth.util.ExceptionUtils    : Tried to detach trace span but it is not the current span: [Trace: 14dd71dd28e39053, Span: 14dd71dd28e39053, exportable=true]. You may have forgotten to close or detach [Trace: 14dd71dd28e39053, Span: e7d1934fd517e593, exportable=true]
```

marcingrzejszczak:
Do you have a sample project you could share?

marcingrzejszczak:
You can check this branch and see if it's working fine now - https://github.com/spring-cloud/spring-cloud-sleuth/compare/issues_%23198_rest_templ_interceptor

dnathanson:
That fixes the issue. Thanks!

One thing...  Since you are rethrowing the exception, there is no need to include the exception/stack trace in your log message.  I'm using Spring's `@Retryable` annotation and I end up with three of your stack traces, followed by my own stack trace in the logs.

marcingrzejszczak:
True, good point!

PR #201, opened 2016-03-04
Issues #198 ribbon hystrix

marcingrzejszczak:
Fixes the Sleuth Feign support issue related to not closing spans that were interrupted by an exception

cc @dsyer , @spencergibb, @adriancole


spencergibb:
I worry about the Feign integration (I think it predates this PR). The moment someone specifies a `@FeignClient(configuration = MyConfig.class)` and specifies a `Decoder` or `Builder` (ie non-hystrix), all of this is broken. Maybe that is something we can live with or document.

marcingrzejszczak:
Yeah the easiest would be to make the impls public (the methods not necessarily) and write in the docs that you have to wrap your implementations in our decorators... I was thinking also about a post processor that would wrap all of these

marcingrzejszczak:
How about now? I've added some docs, some classes are public so one can wrap his own decoder / retryer in a sleuth solution. Also I've added a bean post processor that wraps those things in our custom impls.

marcingrzejszczak:
Any comments? @dsyer @spencergibb @adriancole ? If not then I'd merge this

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/TraceFeignRetryer.java
```diff
@@ -0,0 +1,66 @@
...
+	public TraceFeignRetryer(Tracer tracer, Retryer delegate) {
+		this.tracer = tracer;
+		this.delegate = delegate;
+	}
+
+	@Override
+	public void continueOrPropagate(RetryableException e) {
+		try {
+			this.feignRequestContext.putSpan(this.tracer.getCurrentSpan(), true);
+			this.tracer.getCurrentSpan().logEvent("retry");
```
adriancole: "feign.retry"

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/TraceFeignRequestInterceptor.java
```diff
@@ -0,0 +1,122 @@
...
+		}
+		Long parentId = getParentId(span);
+		if (parentId != null) {
+			setHeader(template, Span.PARENT_ID_NAME, Span.idToHex(parentId));
+		}
+		setHeader(template, Span.PROCESS_ID_NAME, span.getProcessId());
+		publish(new ClientSentEvent(this, span));
+	}
+
+	protected String getSpanName(RequestTemplate template) {
```
adriancole: I'd always envisioned `MethodMetadata.configKey()` being used for operation name lookups, but that's not available at this call site.. Logger is another way to hack this stuff in

adriancole:
I see what you have going on, but it seems like a lot of work. The additional scope of the time spent decoding doesn't seem enough value to justify this much code.

Since tracing is often very similar to logging, I'd recommend just making a Logger decorator that handles these facets. There are a couple package-protected methods that we could open up: logRetry, logIOException.

ex
```java
class TracingLogger extends Logger {
  private final Logger delegate;

// override all methods, tracing, then delegating
--snip--
```

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/TraceFeignDecoder.java
```diff
@@ -0,0 +1,82 @@
...
+import feign.codec.Decoder;
+
+/**
+ * A decoder that closes a span upon decoding the response.
+ *
+ * @author Marcin Grzejszczak
+ *
+ * @since 1.0.0
+ */
+public class TraceFeignDecoder implements Decoder, ApplicationEventPublisherAware {
```
adriancole: since not all responses are decoded, you might need to play with Response.Body itself. Ex closing the span on body.close unless there's no body.

adriancole:
PS totally fine to have some of these things refactored later, particularly if types are hidden while they shape up

marcingrzejszczak:
Done in - https://github.com/spring-cloud/spring-cloud-sleuth/commit/19249cdc23a655125a58d022525276e1f74aae09

PR #203, opened 2016-03-07
Added more documentation

marcingrzejszczak:


docs/src/main/asciidoc/spring-cloud-sleuth.adoc
```diff
@@ -273,10 +274,117 @@ the number of dropped spans will get increased.
...
+    - the Span will be tagged with that method's class name and the method name too
+
+==== Executor, ExecutorService and ScheduledExecutorService
+
+We're providing `LazyTraceExecutor`, `TraceableExecutorService` and `TraceableScheduledExecutorService`. Those implementations
+are creating Spans each time a new task is submitted, invoked or scheduled.
+
+=== Messaging
+
+To disable these features just set the `spring.sleuth.integration.enabled` property to `false`.
```
adriancole: nit: I think you have at least 3 sentence variants of this instruction. For each section, 

// <20 word overview and how to disable
Foo instrumentation (adds X, tags Y, etc). To disable Foo instrumentation, set `spring.sleuth.foo.enabled` to false.

// if more info is available, detail it in a new paragraph

Ex 1. Spring Integration
> http://projects.spring.io/spring-integration/[Spring Integration] instrumentation creates spans for publish and subscribe  events. To disable Spring Integration instrumentation, set `spring.sleuth.integration.enabled` to false.

Ex 2. Zuul
> https://github.com/Netflix/zuul/[Zuul] instrumentation propagates trace information on outbound via http headers. To disable Zuul instrumentation, set `spring.sleuth.zuul.enabled` to false.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/documentation/SpringCloudSleuthDocTests.java
```diff
@@ -61,8 +63,8 @@ public Sampler defaultSampler() {
 	}
 
 	// tag::span_name_annotation[]
-	@SpanName("calculateTax")
-	class TaxCountingRunnable implements Runnable {
+	@SpanName("processTax")
```
adriancole: supernit: was this rename intentional? I think calculateTax is a decent name and also doesn't incidentally join the word Process with Runnable

adriancole:
looks like progress. thanks!

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/documentation/SpringCloudSleuthDocTests.java
marcingrzejszczak: yeah that was exactly why I did it. Ok I'll rename it then back to what it was.

marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-sleuth/commit/6b6f40aed2e5ba3d1983c265084e9465231e8592

PR #207, opened 2016-03-09
Updates to latest zipkin 0.7

adriancole:
This is a maintenance update to zipkin which has no api impact.

* zipkin-server supports cassandra now
* zipkin-web is now javascript (even if still hosted on a scala process)


marcingrzejszczak:
:shipit:

PR #210, opened 2016-03-10
[#209] Skip pattern for Scheduled

### LINKED ISSUE #209: With Consul Config, schedule KV watcher is creating spans
Spring Cloud Consul's config component schedules a watcher that checks is KV store in Consul has been updated.  Sleuth is automatically creating spans anything @Scheduled so in Zipkin there are lots of spans for Local Component: scheduled, Class: ConfigWatch, Method: watchConfigKeyValues.

There should be way to exclude scheduled classes from tracing via configuration.

marcingrzejszczak:
fixes #209


spencergibb:
:+1: nice!

Issue #211, opened 2016-03-11
Decide on one way of span creation


marcingrzejszczak:
We have things like acquire release etc and we also have a tracer api with start/finish etc. We should decide on one - the API most likely.

marcingrzejszczak:
Fixed in https://github.com/spring-cloud/spring-cloud-sleuth/commit/05540b9d2e2f9644afc612f2cd85d6aa9a4c9417

PR #212, opened 2016-03-12
Removed events and replaced them with method calls

### LINKED ISSUE #156: Refactor event propagation to interface execution
The current event propagation seems to be introducing unnecessary complexity without any visible benefits. Maybe it would be worth to refactor the code to use plain old interfaces. Direct coupling would improve debuggability.

marcingrzejszczak:
Fixed Travis script to start building branches

fixes #156 #211


.travis.yml
```diff
@@ -23,15 +20,21 @@ script:
...
     MILESTONE=${MILESTONE}
     MVN_PROFILE=${MVN_PROFILE}
+    TRAVIS_BRANCH=${TRAVIS_BRANCH}
+    TRAVIS_REPO_SLUG=${TRAVIS_REPO_SLUG}
+    TRAVIS_PULL_REQUEST=${TRAVIS_PULL_REQUEST}
 
     EOF
+- ./mvnw install -P docs -q -U -DskipTests=true -Dmaven.test.redirectTestOutputToFile=true
+- '[ "${MVN_GOAL}" == "deploy" ] && ./docs/src/main/asciidoc/ghpages.sh || echo "Not updating docs"'
+script:
```
adriancole: nit: when changing things that seem unrelated, self-note in the PR what's going. it isn't a problem to change other things per-se, just distracting to figure out what if anything this has to do with the core change.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/NoOpSpanReporter.java
```diff
@@ -14,20 +14,18 @@
...
+ * @author Marcin Grzejszczak
  *
- * @author Spencer Gibb
+ * @since 1.0.0
  */
-@SuppressWarnings("serial")
-public class SpanContinuedEvent extends SpanContainingEvent {
+public class NoOpSpanReporter implements SpanReporter {
+	@Override
+	public void reportForCollection(Span span) {
```
adriancole: reportForCollection is unnecessarily long and maybe lies. Not all implementations will have a collection tier, ex what if it just reports to a log file? I'd simplify by just saying report

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/SpanReporter.java
```diff
@@ -14,23 +14,18 @@
...
-
-	public ClientSentEvent(Object source, Span span) {
-		super(source, span);
-	}
-
-}
\ No newline at end of file
+public interface SpanReporter {
+	/**
+	 * Reports a Span for collection
```
adriancole: Reports a completed span out of band, usually out of process. This is typically to a trace depot (ex zipkin) or a log file.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
```diff
@@ -202,6 +193,19 @@ else if (skip) {
 		}
 	}
 
+	private void logServerReceived(Span parent) {
+		if (parent != null && parent.isRemote()) {
+			parent.logEvent(Span.SERVER_RECV);
+		}
+	}
+
+	private void logServerSent(SpanReporter spanReporter, Span parent) {
```
adriancole: I think these two methods are better inlined. For example, this method conditionally logs server sent, and also reports! The extraction doesn't help line count that much and the method name hides some pretty important details.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/AbstractTraceHttpRequestInterceptor.java
```diff
@@ -96,22 +85,16 @@ private String uriScheme(URI uri) {
...
 	/**
-	 * Close the current span and emit the ClientReceivedEvent
+	 * Close the current span and log the client received event
 	 */
 	public void finish() {
 		if (!isTracing()) {
 			return;
 		}
-		publish(new ClientReceivedEvent(this, currentSpan()));
+		currentSpan().logEvent(Span.CLIENT_RECV);
```
adriancole: I think that this is done directly, here. I can see without navigation, that this task results in a "cr"

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/log/SpanLogger.java
```diff
@@ -14,27 +14,37 @@
...
- * @author Spencer Gibb
- * @since 1.0.0
+ * @author Marcin Grzejszczak
  *
- * @see ClientSentEvent
+ * @since 1.0.0
  */
-@SuppressWarnings("serial")
-public class ServerReceivedEvent extends SpanParentContainingEvent {
+public interface SpanLogger {
```
adriancole: it seems this component's primary concern is synchronizing MDC context with span id context. At least that's the only implementation, right?

adriancole:
This change is certainly a step in the right direction.

SpanLogger seems the most likely to thrash (before or after this change), as the name is very generalized, yet it is designed for a very specific in-process span propagation concern. I think we should try to buff this out before 1.0

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/AbstractTraceHttpRequestInterceptor.java
marcingrzejszczak: I don't get it

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/log/SpanLogger.java
marcingrzejszczak: Yes for now MDC is the only impl other than NoOp. Maybe someone wants to add it for commons or sth (I don't know if that's even possible)

.travis.yml
marcingrzejszczak: Added info to the PR

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/AbstractTraceHttpRequestInterceptor.java
adriancole: I meant to type 'I like that this is done here'! Yeah you got it

Issue #19, opened 2015-08-12
Support for configurable header names


dsyer:
The zipkin headers are here: http://zipkin.io/Instrumenting.html#instrumenting-other-libraries

> Ids are encoded as hex strings
> X-B3-TraceId: 64 encoded bits
> X-B3-SpanId: 64 encoded bits
> X-B3-ParentSpanId: 64 encoded bits
> X-B3-Sampled: Boolean (either “1” or “0”)
> X-B3-Flags: a Long

There is an obvious immediate problem: Sleuth uses UUID for the ids (64 bits is really too small and Zipkin admits it).

We could also support Zipkin headers as an enhancement (include both sets). Seems like a waste of headers (we only need 3, but we end up with 7 or 8).

dsyer:
Zipkin also offers instrumentation for Thrift services. If we are going for interop we should do the same.

spencergibb:
@adriancole thoughts? Have those headers changed?

adriancole:
`X-B3-Flags: a Long` was a mistake imho. I'll see if finagle folks want to remove this. The others are stable. PS there's a nuance hex strings are lowercase fixed-width.

marcingrzejszczak:
What do we do about this? We can close it I guess?

dsyer:
I think we still need it. But maybe not urgent for 1.0.

adriancole:
The bit size is aligned at this point. I'd recommend preferring B3 headers as the majority of instrumentation use these.

For example, there's no ruby-sleuth, and we'd needlessly re-start traces as a side-effect of passing through sleuth. The only architecture where the current choice doesn't hurt is when then entire architecture is made of sleuthed boot apps.

Here's the effect of being incompatible in polyglot or otherwise zipkin architecture:

Trace starts in zipkin instrumented entrypoint. This calls out to sleuth which ignores the headers and starts a new trace. Sleuth calls out downstream with an incompatible header. The receiver ignores that header and can't see the original headers so thinks its a new trace again. So 3 traces for the same trace in the same zipkin span store.

drpotato:
@marcingrzejszczak thanks for picking this up, looking forward to seeing it done!

marcingrzejszczak:
I'll put some of @adriancole 's notes from Gitter

> I think the core issue here is externalizing how to deal with out-of-process propagation
similarly to how reporting is configurable, yet doesn't change the structure of the span
[...]

PR #213, opened 2016-03-14
Introduction of Joiners and Injectors

marcingrzejszczak:
That's a big one...

* I've added two interfaces `SpanInjector` and `SpanJoiner`.
* `Tracer` extends those interfaces and I'm manually passing `SpanJoinerComposite` and `SpanInjectorComposite` to `DefaultTracer`. The two Composite classes have all `SpanInjector` and `SpanJoiner` interfaces injected. That way you have to just register your version of `SpanInjector` and `SpanJoiner` and it will be automatically seen by `Tracer`. 
    * The good side of this is that from the client's perspective it's enough to call the `Tracer's` method - the internals are hidden
    * We're aligned with OpenTracing on this directly
    * The problem (related to last point) with this approach is that `Tracer` CANNOT be injected to any of those two interfaces (`Injector` and `Joiner`) since there's going to be a circular dependency.
* I've added `SpanInjector` and `SpanJoiner` for Http filter, Feign, Zuul and RestTemplate. That way only a handful of production classes (other than implementations of those two interfaces) + tests use Span header names directly. Thanks to this it will be easier to do step 2 which is introduce configurable headers via properties.
* Removed the constraint that headers are set only if tracing is on (that there is a span in the current thread) for Zuul, Feign and Rest Template. Iv'e done it cause the constraint was set only for the additional Sleuth related headers - there was no consistency. 
    * Also due to technical reasons - I would have to (which I did in another branch) refactor the whole code to separate `SpanAccessor` from `Tracer`.
    * For that reason `Tracer` can't extend `SpanAccessor` cause if you want to use `SpanAccessor` in the `SpanInjector` and `SpanJoiner` you will have a circular dependency

@spencergibb @dsyer @adriancole  - WDYT?

Related to issue #19


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/TraceHeaders.java
```diff
@@ -0,0 +1,128 @@
...
+ *     <li>{@code X-B3-ParentSpanId} 64 encoded bits</li>
+ *     <li>{@code X-B3-Sampled} Boolean (either “1” or “0”)</li>
+ * </ul>
+ *
+ * @author Marcin Grzejszczak
+ *
+ * @since 1.0.0
+ */
+@ConfigurationProperties("spring.sleuth.headers")
+public class TraceHeaders {
```
marcingrzejszczak: I'm not using this anywhere yet - you can disregard that class

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
```diff
@@ -62,10 +69,24 @@ public Sampler defaultTraceSampler() {
...
+			SpanNamer spanNamer, SpanLogger spanLogger,
+			SpanReporter spanReporter, List<SpanJoiner> spanJoiners,
+			List<SpanInjector> spanInjectors) {
 		return new DefaultTracer(sampler, random, spanNamer, spanLogger,
-				spanReporter);
+				spanReporter, new SpanJoinerComposite(spanJoiners),
+				new SpanInjectorComposite(spanInjectors));
+	}
+
+	// have to register a NoOp - otherwise autowiring won't work for trace manager
```
spencergibb: Not if it were moved to a static inner class and `@Autowired(required = false) List<SpanJoiner> spanJoiners = new ArrayList<>();` was used.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/MessagingSpanInjector.java
```diff
@@ -29,24 +30,30 @@
...
-			return (Span) object;
+	private final TraceKeys traceKeys;
+
+	public MessagingSpanInjector(TraceKeys traceKeys) {
+		this.traceKeys = traceKeys;
+	}
+
+	@Override
+	@SuppressWarnings("unchecked")
+	public <T> void inject(Span span, T carrier) {
```
spencergibb: Why not put `<T>` on `SpanInjector`?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/MessagingSpanJoiner.java
```diff
@@ -0,0 +1,92 @@
...
+import org.springframework.messaging.Message;
+
+/**
+ * Creates a {@link SpanBuilder} from {@link Message}
+ *
+ * @author Marcin Grzejszczak
+ *
+ * @since 1.0.0
+ */
+public class MessagingSpanJoiner implements SpanJoiner {
```
spencergibb: Same for `SpanJoiner<T>`.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/MessagingSpanInjector.java
marcingrzejszczak: I even did that once :) But this is exactly how it is done in OpenTracing.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/MessagingSpanJoiner.java
marcingrzejszczak: It's exactly how it is done in OpenTracing

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
marcingrzejszczak: I've already tried that and it won't work cause Tracer is also a `SpanJoiner` so you will have a circular dependency cycle.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/MessagingSpanInjector.java
spencergibb: OK, what does that matter?

marcingrzejszczak: Cause:

- we want to be compatible with OpenTracing 
- it doesn't change much cause still I have to do sth like "accept" from Observator pattern

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/MessagingSpanJoiner.java
adriancole: Actually OT doesn't have this type, it has these methods on Tracer. When extracted into a type whose only concern is injection for a specific carrier, it makes sense to move the carrier type param to the type.

adriancole: Marcin think of this like gson or similar. Gson itself can serialize any
type, but to teach it a specific type, you use a jsonadapter which has a
type target parameter for the class being serialized.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/MessagingSpanInjector.java
adriancole: I think my other notes follow this, but this doesn't have to do with OT
unless the method is on the Tracer itself. The Tracer accepts variable
types as it is used in many contexts (ex messaging and also HTTP). This
class has a limited concern.

I would expect an implementation of OT to internally have inject adapters
for different carriers and the composite (tracer) would have a registry of
those adapters.

OT or not OT, methods are designed differently depending on the shape and
scope of the call site. It is a subtle, but important difference. I can
work with you on this today, and thanks for raising a straw man!

adriancole:
summary of notes, based on some lengthy discussion :)

* bind generic type params to `Joiner` and `Injector`, similar to https://github.com/opentracing/opentracing-java/pull/11/files#diff-eb9c3460aba76aabc0de04b05e4a2b3d
* autowire `Joiner|Injector<X>` where the types are used. This makes it easier to understand.
* drop `Span*Composite` if this is a new feature. doing so lowers the scope of this PR.

Ex.
```java
// both are functional interfaces

// carrier needs to be mutable here. ex Message.Builder
interface Injector<T> {
  void inject(Span span, T carrier);
}

// carrier is not affected, so can be immutable. ex. Message
interface Joiner<T> {
  Span join(T carrier);
}
```

marcingrzejszczak:
Ok I've refactored the code to have `SpanInjector` and `SpanExtractor`. I had gigantic issues with generics thus there's a couple of TODOs.

cc @dsyer @adriancole @spencergibb

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/TraceHeaders.java
adriancole: probably PropagationHeaders

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/MessagingSpanInjector.java
```diff
@@ -29,24 +30,26 @@
...
-		}
-		return null;
+	private final TraceKeys traceKeys;
+
+	public MessagingSpanInjector(TraceKeys traceKeys) {
+		this.traceKeys = traceKeys;
+	}
+
+	@Override
+	@SuppressWarnings("unchecked")
```
adriancole: weird that this needed a warning supression

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/MessagingSpanInjector.java
```diff
@@ -55,27 +58,28 @@ public static Span getSpanFromHeader(Message<?> message) {
 	 * a tag set it will not get overridden.
 	 *
 	 * @param traceKeys - the global configuration for trace keys
-	 * @param message - message to which headers will be added
+	 * @param messageBuilder - message builder to which headers will be added
 	 * @param span - span from which headers will be taken
 	 * @return the input message with updated headers
 	 */
-	public static Message<?> addSpanHeaders(TraceKeys traceKeys, Message<?> message,
+	public void addSpanHeaders(TraceKeys traceKeys, MessageBuilder messageBuilder,
```
adriancole: I'd take out the method dispatch as you'll notice the docs have drifted anyway

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/websocket/TraceWebSocketAutoConfiguration.java
```diff
@@ -46,12 +56,28 @@ public void registerStompEndpoints(StompEndpointRegistry registry) {
...
 	}
 
 	@Override
 	public void configureClientInboundChannel(ChannelRegistration registration) {
 		registration.setInterceptors(
-				new TraceChannelInterceptor(this.tracer, this.traceKeys, this.random));
+				new TraceChannelInterceptor(this.tracer, this.traceKeys, this.spanExtractor, this.spanInjector));
+	}
+
+	// TODO: Qualifier + ConditionalOnProp cause autowiring generics doesn't work
```
adriancole: thx for the TODOs

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/HttpServletResponseInjector.java
```diff
@@ -0,0 +1,46 @@
...
+ * @since 1.0.0
+ */
+class HttpServletResponseInjector implements SpanInjector<HttpServletResponse> {
+
+	@Override
+	public void inject(Span span, HttpServletResponse carrier) {
+		addResponseHeaders(span, carrier);
+	}
+
+	private void addResponseHeaders(Span span, HttpServletResponse response) {
```
adriancole: I don't think dispatching here helps much.. how about inlining it?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
```diff
@@ -199,13 +153,32 @@ else if (skip) {
...
-		if (span != null) {
-			if (!response.containsHeader(Span.SPAN_ID_NAME)) {
-				response.addHeader(Span.SPAN_ID_NAME, Span.idToHex(span.getSpanId()));
-				response.addHeader(Span.TRACE_ID_NAME, Span.idToHex(span.getTraceId()));
+	/**
+	 * Creates a span and appends it as the current request's attribute
+	 */
+	private Span createSpan(HttpServletRequest request, HttpServletResponse response,
+			boolean skip, Span spanFromRequest, String name) {
+		if (spanFromRequest == null) {
```
adriancole: the level of indenting here makes it hard to read. I'd early return here

ex 

```java
if (spanFromRequest != null) return spanFromRequest;
// now everything can shift left one tab

File: spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/AbstractTraceHttpRequestInterceptor.java [hunk 10]
```diff
@@ -34,38 +34,16 @@
...
-				this.tracer.isTracing()) {
-			request.getHeaders().add(name, value);
-		}
-	}
-
-	private void setIdHeader(HttpRequest request, String name, Long value) {
-		if (value != null) {
-			setHeader(request, name, Span.idToHex(value));
-		}
+	private void enrichWithTraceHeaders(Span span, HttpRequest request) {
```
adriancole: loose the dispatch and inline. that way, we call the same operation the same thing (injecting a span into a request as opposed to enriching with trace headers)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/FeignRequestTemplateInjector.java
```diff
@@ -0,0 +1,62 @@
...
+import feign.RequestTemplate;
+
+/**
+ * Span injector that injects tracing info to {@link RequestTemplate}
+ *
+ * @author Marcin Grzejszczak
+ *
+ * @since 1.0.0
+ */
+class FeignRequestTemplateInjector implements SpanInjector<RequestTemplate> {
```
adriancole: I actually like this.. using the common term injector allows me to relate to all the different code easily. thanks

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/FeignResponseHeadersHolder.java
```diff
@@ -0,0 +1,35 @@
...
+import java.util.Map;
+
+/**
+ * Mutable holder for Feign Response headers
+ *
+ * @author Marcin Grzejszczak
+ *
+ * @since 1.0.0
+ */
+class FeignResponseHeadersHolder {
```
adriancole: thanks for making this package private.. I'm thinking maybe we add a feign response builder at some point.

adriancole:
thanks I know this was hard work. You've noted some follow-up concerns which can be done post merge

:shipit:

marcingrzejszczak:
Done in https://github.com/spring-cloud/spring-cloud-sleuth/commit/7c5af8c84e7cdb4d821975feea966eb49cf7e2d8

PR #214, opened 2016-03-15
Added Trace Headers

### LINKED ISSUE #19: Support for configurable header names

marcingrzejszczak:
* instead of direct passing of headers `TraceHeaders` is introduced
* by changing one property you can change trace-id to for example 'correlationId'
* updated docs to reflect the change
* defaults to Zipkin headers
* the brewery is working fine (on a branch)
* changed the way we set pass info about sampling spans (until now we had "not-sampled")

fixes #19

cc @spencergibb , @dsyer , @adriancole

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/TraceHeaders.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/TraceHeaders.java
@@ -38,16 +38,64 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 @ConfigurationProperties("spring.sleuth.headers")
 public class TraceHeaders {
 
-	private Zipkin zipkin;
-	private Sleuth sleuth;
+	public static final String SPAN_SAMPLED = "1";
+	public static final String SPAN_NOT_SAMPLED = "0";
 
-	public Zipkin getZipkin() {
-		return this.zipkin;
+	public static final String ZIPKIN_TRACE_ID_HEADER_NAME = "X-B3-TraceId";
+	public static final String ZIPKIN_SPAN_ID_HEADER_NAME = "X-B3-SpanId";
+	public static final String ZIPKIN_PARENT_SPAN_ID_HEADER_NAME = "X-B3-ParentSpanId";
+	public static final String ZIPKIN_SAMPLED_HEADER_NAME = "X-B3-Sampled";
+	public static final String ZIPKIN_PROCESS_ID_HEADER_NAME = "X-Process-Id";
+
+	private String traceId = ZIPKIN_TRACE_ID_HEADER_NAME;
+	private String spanId = ZIPKIN_SPAN_ID_HEADER_NAME;
+	private String parentSpanId = ZIPKIN_PARENT_SPAN_ID_HEADER_NAME;
+	private String sampled = ZIPKIN_SAMPLED_HEADER_NAME;
+	private String processId = ZIPKIN_PROCESS_ID_HEADER_NAME;
+
+	public String getTraceId() {
+		return this.traceId;
 	}
 
-	public void setZipkin(Zipkin zipkin) {
-		this.zipkin = zipkin;
+	public void setTraceId(String traceId) {
+		this.traceId = traceId;
 	}
+
+	public String getSpanId() {
+		return this.spanId;
+	}
+
+	public void setSpanId(String spanId) {
+		this.spanId = spanId;
+	}
+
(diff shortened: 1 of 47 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/TraceHeaders.java
```diff
@@ -38,76 +38,88 @@
...
+	public static final String SPAN_SAMPLED = "1";
+	public static final String SPAN_NOT_SAMPLED = "0";
+
+	public static final String ZIPKIN_TRACE_ID_HEADER_NAME = "X-B3-TraceId";
+	public static final String ZIPKIN_SPAN_ID_HEADER_NAME = "X-B3-SpanId";
+	public static final String ZIPKIN_PARENT_SPAN_ID_HEADER_NAME = "X-B3-ParentSpanId";
+	public static final String ZIPKIN_SAMPLED_HEADER_NAME = "X-B3-Sampled";
+	public static final String ZIPKIN_PROCESS_ID_HEADER_NAME = "X-Process-Id";
+
+	private String traceId = ZIPKIN_TRACE_ID_HEADER_NAME;
```
spencergibb: `traceIdName`? `setTraceId` looks funny to me.

shakuzen:
I think `features.adoc` needs to be updated as well. The note about the default logging pattern at the bottom will be incorrect if this gets merged, since the default header names changed to match Zipkin.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/TraceHeaders.java
adriancole: I think these constant names are long winded and a bit imprecise.

Ex the _NAME suffix is wasted characters IMHO, and unnecessarily causes
line wraps.

There's a ZIPKIN_ header that has nothing to do with zipkin or b3 (process
ID)

Here's an example of the complete set of zipkin/b3 headers. My only
suggestion here would to be rename it to PropagationHeaders. Also, if you
do that, you'll notice you can drop the HEADERS part, too! The whole thing
is wildly more succinct

https://github.com/openzipkin/brave/blob/master/brave-http/src/main/java/com/github/kristofa/brave/http/BraveHttpHeaders.java

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/MessagingSpanExtractor.java
```diff
@@ -33,29 +34,32 @@
...
+	public MessagingSpanExtractor(Random random, TraceHeaders traceHeaders) {
 		this.random = random;
+		this.traceHeaders = traceHeaders;
 	}
 
 	@Override 
 	public Span joinTrace(Message carrier) {
-		if (!hasHeader(carrier, Span.TRACE_ID_NAME)
-				|| !hasHeader(carrier, Span.SPAN_ID_NAME)) {
+		if (!hasHeader(carrier, this.traceHeaders.getTraceId())
```
adriancole: This is fragile. This assumes that a change to how inject/extract work will only include the header name (and not the encoding of the value). Thats an odd special case that costs instructure to support.

The type injector/extractor are made precisely to support concrete changes to propagation headers, for example where values are encoded differently or bundled all into a single header.

If I were you, I would do this:

First, is to change the names and rename the injector/joiner to Zipkin (or B3).
Second, wait for a concrete use case before adding customization, as injector/extractor are already possible to adjust.

marcingrzejszczak:
I had a chat with @adriancole and we're going to do things in another way...

* This PR will be ditched - it's too complex too configure all of the headers like that
* We will rename current headers to make them Zipkin compatbile
* We will change the "not sampled" to "sampled"
* We will write the docs and examples to show how can one write his own injector to propagate spans from custom headers

PR #215, opened 2016-03-16
Made headers Zipkin compatible

marcingrzejszczak:


adriancole:
that was easy to review! LGTM (ack that you have a TODO to converge docs)

docs/src/main/asciidoc/features.adoc
```diff
@@ -30,6 +30,9 @@ latency in your applications. Sleuth is written to not log too much, and to not
 * Instruments common ingress and egress points from Spring applications (servlet filter, async endpoints,
 rest template, scheduled actions, message channels, zuul filters, feign client).
 
+* Via `SpanInjector` and `SpanExtractor` interface allows to customize the way spans are
```
adriancole: not sure if this is too long winded..

Sleuth includes default logic to join a trace across http or messaging boundaries. For example, http propagation works via Zipkin-compatible request headers. This propagation logic is defined and customized via `SpanInjector` and `SpanExtractor` implementations.

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
```diff
@@ -191,7 +191,120 @@ adding a Channel Binder implementation
...
 automatically turn your app into a producer of messages with payload
-type `Spans`. 
+type `Spans`.
+
+== Customizations
+
+Thanks to the `SpanInjector` and `SpanExtractor` you can customize the way spans
+are created and propagated.
+
+Essentially with Spring Cloud Span there are two ways span tracing data can be passed to
```
adriancole: There are currently two built-in ways to pass tracing information between processes:

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
```diff
@@ -191,7 +191,120 @@ adding a Channel Binder implementation
...
+	@Qualifier("messagingSpanInjector")
+	@ConditionalOnProperty(value = "spring.sleuth.integration.injector.enabled", matchIfMissing = true)
+	public SpanInjector<MessageBuilder> messagingSpanInjector() {
+		...
+	}
+----
+
+As you can see you can disable them with the properties:
+
+* `Message` span extractor - `spring.sleuth.integration.extractor.enabled`
```
adriancole: interesting properties.. I think the idea is that you want to stop propagation, right? like ingress and egress?

these nobs say for example, you have a scenario where you want to take spans, but not pass them down and visa versa.

do we need a knob like this yet? or would it be simpler to just disable the feature? that way, the more jargony terms extractor injector don't need to leak into property namespace.

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
```diff
@@ -191,7 +191,120 @@ adding a Channel Binder implementation
...
+
+There is also a WebSockets version:
+
+* `Message` span extractor - `spring.sleuth.integration.websocket.extractor.enabled`
+* `MessageBuilder` span injector - `spring.sleuth.integration.websocket.injector.enabled`
+
+You can override them either by disabling with a property and providing your own implementation
+or by adding a `@Primary` annotation to your bean definition. Remember to provide the following qualifiers:
+
+* `Message` span extractor - `messagingSpanExtractor`
```
adriancole: this seems like documenting a TODO :) are we sure these qualifiers aren't needed? or do we just think we should document how things are today in case they don't improve?

adriancole: are these needed bc messaging, stomp, etc might all be needed in the same spring context? maybe make or reuse a qualifier annotation? just thoughts as strings like this seem fragile

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
```diff
@@ -191,7 +191,120 @@ adding a Channel Binder implementation
...
+
+Essentially with Spring Cloud Span there are two ways span tracing data can be passed to
+ a process.
+
+ * via Spring Integration
+ * via HTTP
+
+=== Spring Integration
+
+For Spring Integration these are the beans responsible for creation of a Span from a `Message`
```
adriancole: before sections like this. overview in a sentence or two how propagation works. Forgive the rough sentences.

ex. For HTTP, propagation is via http headers. For example, span ids are extracted from Zipkin-compatible (B3) headers, to start or join an existing trace. Trace information is injected into any outbound requests so the next hop can extract them.

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
marcingrzejszczak: I didn't want to use @Primary but actually it's better to use it instead of passing props. I'll remove those properties

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
marcingrzejszczak: > this seems like documenting a TODO :) are we sure these qualifiers aren't needed? or do we just think we should document how things are today in case they don't improve?

yes - if someone wants to override the behaviour he needs to pass the qualifier. We can add annotations instead. @MessagingSpanExtractor or sth like that (actually that one is already reserved for the impl).

> are these needed bc messaging, stomp, etc might all be needed in the same spring context? maybe make or reuse a qualifier annotation? just thoughts as strings like this seem fragile

There was an issue with configs - I fixed that

marcingrzejszczak:
Done in https://github.com/spring-cloud/spring-cloud-sleuth/commit/3bb7916fffd33072e0b3bd66531a85420fc11aaf

@adriancole I removed qualifiers and properties :)

adriancole:
yay

PR #220, opened 2016-03-18
Updates to zipkin 0.9

adriancole:
Notably, this removes the "zipkin-web" dependency, as the UI is now
collocated on zipkin-server.

Other notable changes are Cassandra and Kafka support, as well removing
the Brave dependency.


marcingrzejszczak:
:shipit:

PR #221, opened 2016-03-21
Updates to zipkin 0.9.3, which fixes a UI recursion bug

adriancole:


adriancole:
see https://github.com/openzipkin/zipkin/issues/1056

marcingrzejszczak:
:shipit:

adriancole:
bumped to 0.9.3 which fixes some bugs found backporting tests

spring-cloud-sleuth-samples/spring-cloud-sleuth-sample-zipkin-stream/pom.xml
```diff
@@ -54,6 +55,11 @@
 			<artifactId>spring-cloud-sleuth-zipkin-stream</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>io.zipkin</groupId>
```
adriancole: @marcingrzejszczak added this as otherwise the example won't show the UI (also latest zipkin-web is a dumb proxy to the query server, so deploying it would just point to 404 unless we host it)

PR #222, opened 2016-03-21
Scrubs docs about zipkin-web as it is now embedded

adriancole:


adriancole:
@dsyer PTAL I know we don't need zipkin-web anymore, but not sure if we should have similar CF instructions for zipkin-server or not

marcingrzejszczak:
Hmm if we have spring-cloud-stream then we need a UI of some sort. That's why we also deploy Zipkin Web

adriancole:
zipkin's UI is now hosted on the zipkin server itself! so basically, if you
hit http://localhost:9411 you'll see the UI now.

Here's the spring config:

https://github.com/openzipkin/zipkin-java/blob/master/zipkin-server/src/main/java/zipkin/server/ZipkinUiConfiguration.java#L40

spring-cloud-sleuth-zipkin-stream/README.md
```diff
@@ -24,24 +24,11 @@ public class ZipkinStreamServerApplication {
 }
 ```
 
+Zipkin has a web UI, which is enabled by default when you depend on `io.zipkin:zipkin-ui`.
```
adriancole: @marcingrzejszczak wdyt about this?

marcingrzejszczak: Sweet!

PR #233, opened 2016-03-28
Tightens Log to forbid null events and fixes Span json serialization

adriancole:
The Log object should never have a null entry as it is defined. This
change fixes that, backfills equals/hashCode for convenience, and adds
Jackson annotations that use the validating constructor.

This also adds a json annotation to span that allows logs to be
serialized. This is tested by round-tripping the json.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Log.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Log.java
@@ -16,6 +16,9 @@
 
 package org.springframework.cloud.sleuth;
 
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 /**
  * Represents an event in time associated with a span. Every span has zero or more Logs,
  * each of which being a timestamped event name.
@@ -30,7 +33,7 @@ public class Log {
 	private final long timestamp;
 
 	/**
-	 * Event (if not null) should be the stable name of some notable moment in the lifetime of a span.
+	 * Event should be the stable name of some notable moment in the lifetime of a span.
 	 * For instance, a span representing a browser page load might add an Event for each of the
 	 * Performance.timing moments here: https://developer.mozilla.org/en-US/docs/Web/API/PerformanceTiming
 	 *
@@ -40,7 +43,12 @@ public class Log {
 	 */
 	private final String event;
 
-	public Log(long timestamp, String event) {
+	@JsonCreator
+	public Log(
+			@JsonProperty(value = "timestamp", required = true) long timestamp,
+			@JsonProperty(value = "event", required = true) String event
+	) {
+		if (event == null) throw new NullPointerException("event");
 		this.timestamp = timestamp;
 		this.event = event;
 	}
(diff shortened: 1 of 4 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Log.java
```diff
@@ -30,7 +33,7 @@
 	private final long timestamp;
 
 	/**
-	 * Event (if not null) should be the stable name of some notable moment in the lifetime of a span.
+	 * Event should be the stable name of some notable moment in the lifetime of a span.
```
adriancole: This was only allowed to be null in OpenTracing as the object was a union type. (Ex there were other fields which might be missing, but one had to be present).

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
```diff
@@ -66,6 +67,7 @@
  * like scoped tracers. Sleuth spans are DTOs, whose sole responsibility is the current
  * span in the trace tree.
  */
+@JsonAutoDetect(fieldVisibility = JsonAutoDetect.Visibility.ANY)
```
adriancole: took a lazier approach here, since span has so many fields.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Log.java
```diff
@@ -40,7 +43,12 @@
 	 */
 	private final String event;
 
-	public Log(long timestamp, String event) {
+	@JsonCreator
+	public Log(
+			@JsonProperty(value = "timestamp", required = true) long timestamp,
```
marcingrzejszczak: Is this JsonProperty necessary?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Log.java
```diff
@@ -40,7 +43,12 @@
...
+			@JsonProperty(value = "event", required = true) String event
+	) {
+		if (event == null) throw new NullPointerException("event");
```
marcingrzejszczak: Maybe IllegalArgumentException?

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/SpanTest.java
```diff
@@ -80,4 +81,18 @@ public void getTimelineAnnotationsReadOnly() {
 
 		then(serializedName).isNotEmpty();
 	}
-}
\ No newline at end of file
+
+	@Test public void should_properly_serialize_logs() throws IOException {
```
marcingrzejszczak: Sweet. We were missing this test. Thanks a lot

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Log.java
adriancole: if you use a constructor as opposed to reflection, it is. The "required"
flag isn't .. uhh.. required. I can switch to the reflective annotations,
if you like.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Log.java
adriancole: NPE was designed to report null :P main thing here is just giving it a
label.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Log.java
adriancole: The problem with using reflection is that you need to make a dummy
ctor, and also have the chance for null, still. Ex. the following
still allows a later set of event to null.

@SuppressWarnings("unused") // used for reflective deserialization
private Log() {
 this.timestamp = 0L;
 this.event = "";
}

marcingrzejszczak: Nah I definitely prefer the constructor injection. Initially I did the same but for some reason it failed

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Log.java
marcingrzejszczak: :+1:

PR #234, opened 2016-03-29
Removed docker compose

### LINKED ISSUE #230: Remove docker from zipkin examples, and instead use java -jar
Before, docker was helpful as zipkin had 3 processes, whereas now zipkin has only one: zipkin-server. Since zipkin-server is a spring boot application, it should be familiar to have folks start it directly. 

The hope is that by removing docker from examples, people can focus on the actual integration of note, and maintainers can be more productive by answering less questions.

marcingrzejszczak:
fixes #230


adriancole:
probably need to do a README sweep

marcingrzejszczak:
@adriancole README sweep done :)

adriancole:
except... there's no more reference on how to start zipkin!

PR #238, opened 2016-03-31
Messages don't log events even though they should

marcingrzejszczak:


dsyer:
LGTM (although I assume we'll have to adjust them later when we decide how to deal with "multiple hops").

Issue #235, opened 2016-03-29
Feature Request: Supporting rxjava observables with Rxjava Schedulers


shivangshah:
So far, the tracer's context is lost when using DeferredResult+Observable+Schedulers. It'd be nice to have this supported out of the box with spring-sleuth. Looking at the example here: http://blog.mabn.pl/2014/11/rxjava-logback-and-mdc-threadlocal.html, a RxJavaSchedulersHook can be created (and registered as an RxJava Plugin) and sleuth related variables can be copied over to the newly created thread.

marcingrzejszczak:
Pull requests are more then welcome :)

shivangshah:
@marcingrzejszczak I will give it a go. Would you be able to point me in the direction of which threadlocal variables specific to sleuth need to be copied over?

marcingrzejszczak:
Sure!

There are the `SpanInjector` and `SpanExtractor` interfaces that will help you with that. Let's take a look at https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/HttpRequestInjector.java .

One passing tracing information FROM Sleuth to SOMETHING we're using a `SpanInjector`. It means that we will leave Sleuth and get outside the process most likely (but no necessarily). In that example you can see that from a Span we're putting information to a `HttpRequest`. You can see which information is important for us. All the header names are there in the `Span` class as static constants.

If you want to get information from SOMETHING and build a Span then you're using the `SpanExtractor`. Example for messaging - https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/MessagingSpanExtractor.java . As you can see from a carrier (Message / HttpRequest etc.) that has the tracing information we will build a Span - by means of a `Span.builder()`.

Don't hesitate to ping me here or on Gitter in case of any further questions https://gitter.im/spring-cloud/spring-cloud-sleuth?utm_source=badge&utm_medium=badge&utm_campaign=pr-badge&utm_content=badge

shivangshah:
@marcingrzejszczak :point_up_2:

shivangshah:
@marcingrzejszczak I have pushed the code changes to my own fork (for now, no PR yet). You can find the code here: 
[...]

PR #243, opened 2016-04-02
Support rx java

shivangshah:
Adding support for `RxJava`: https://github.com/spring-cloud/spring-cloud-sleuth/issues/235
We're registering a custom https://github.com/ReactiveX/RxJava/wiki/Plugins#rxjavaschedulershook [`RxJavaSchedulersHook`] that wraps all `Action0` instances into their Sleuth representative - the `TraceAction`. The hook either starts or continues a span depending on the fact whether tracing was already going on before the Action was scheduled. To disable the custom RxJavaSchedulersHook set the `spring.sleuth.rxjava.schedulers.hook.enabled` to `false`.

1) Because RxJavaPlugins is protected does not expose `reset()`, a wrapper had to be introduced. More details can be found here: ReactiveX/RxJava#2297
2) The implementation (and testing) strategy was followed per HystrixConcurrencyStrategy implementation.
3) Adding integration test cases


shivangshah:
One comment I'd like to make here is when I run the integration test case I wrote by itself: 
`mvn test -Dtest=org.springframework.cloud.sleuth.instrument.rxjava.SleuthRxJavaIntegrationTests
` it runs fine. But when I run the full `mvn test` that one specific test fails. I couldn't find the root cause, maybe you guys can help. 
@adriancole @marcingrzejszczak @dsyer

PR #244, opened 2016-04-03
Updates to zipkin 0.11

adriancole:
The only notable change for Sleuth is an artifact id change from
spanstore to storage.


marcingrzejszczak:
:shipit:

PR #252, opened 2016-04-21
Removing entries from MDC

### LINKED ISSUE #248: MDC  question
i have some question about below code:

```
   	@Override
	public void logStoppedSpan(Span parent, Span span) {
		log("Stopped span|{}", span);
		if (parent != null) {
			MDC.put(Span.SPAN_ID_NAME, Span.idToHex(parent.getSpanId()));
			MDC.put(Span.SPAN_EXPORT_NAME, String.valueOf(parent.isExportable()));
		}
		else {
			MDC.remove(Span.SPAN_ID_NAME);
			MDC.remove(Span.SPAN_EXPORT_NAME);
			MDC.remove(Span.TRACE_ID_NAME);
		}
	}
```

when SERVICE_A call SERVICE_B, SERVICE_B always run to "if(parent != null)",but the request is over.

marcingrzejszczak:
when iterating over saved spans during span closing we can apply a function that should be executed on each iteration.

fixes #248

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
@@ -144,7 +144,6 @@ public class TraceFilter extends OncePerRequestFilter {
 				} else {
 					spanFromRequest.logEvent(Span.SERVER_SEND);
 				}
-				// Double close to clean up the parent (remote span as well)
 				this.tracer.close(spanFromRequest);
 			}
 		}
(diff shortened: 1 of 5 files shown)


adriancole:
not quiet sure if this is too much a hammer for the job (ex MDC removal is innocent, but a function could do anything). regardless, maybe a test that shows what the MDC use case is..

marcingrzejszczak:
Note that it's package scope so we're protecting ourselves a bit.

Sure, a test makes sense. I'll add it

adriancole:
ahh I didn't notice the package protection. good point

adriancole:
thx for the test! 👍

PR #261, opened 2016-04-28
The algorithm of percentage based sampling malfunctioned

### LINKED ISSUE #260: PercentageBasedSampler only works in increments of 10%
The threshold calculation in org.springframework.cloud.sleuth.sampler.PercentageBasedSampler breaks for percentages with resolution lower than 10%:

`long threshold = Math.abs(Long.MAX_VALUE * (int) (this.configuration.getPercentage() * 100)); // drops fractional percentage.`

This is is fine for settings like
`spring.sleuth.sampler.percentage=0.1`
`spring.sleuth.sampler.percentage=0.2`
`spring.sleuth.sampler.percentage=0.3`

But always returns true for settings such as
`spring.sleuth.sampler.percentage=0.01`
`spring.sleuth.sampler.percentage=0.15`
`spring.sleuth.sampler.percentage=0.79`

marcingrzejszczak:
I've reused the algorithm from Zipkin Sampler to fix it

fixes #260

CC @adriancole @dsyer @spencergibb

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSampler.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSampler.java
@@ -33,13 +33,13 @@ public class PercentageBasedSampler implements Sampler {
 
 	@Override
 	public boolean isSampled(Span currentSpan) {
-		long threshold = Math.abs(Long.MAX_VALUE * (int) (this.configuration.getPercentage() * 100)); // drops fractional percentage.
+		long threshold = (long) (Long.MAX_VALUE * this.configuration.getPercentage()); // drops fractional percentage.
 		if (currentSpan == null || threshold == 0L) {
 			return false;
 		}
 		long traceId = currentSpan.getTraceId();
-		Long mod = Math.abs(traceId % 100);
-		return mod.compareTo(threshold) <= 0;
+		long t = traceId == Long.MIN_VALUE ? Long.MAX_VALUE : Math.abs(traceId);
+		return t <= threshold;
 	}
 
 }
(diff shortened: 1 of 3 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSampler.java
```diff
@@ -33,13 +33,13 @@ public PercentageBasedSampler(SamplerProperties configuration) {
 
 	@Override
 	public boolean isSampled(Span currentSpan) {
-		long threshold = Math.abs(Long.MAX_VALUE * (int) (this.configuration.getPercentage() * 100)); // drops fractional percentage.
+		long threshold = (long) (Long.MAX_VALUE * this.configuration.getPercentage()); // drops fractional percentage.
```
spencergibb: Comment no longer applies?

adriancole: comment should be `// safe cast as percentage < 1` or similar

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/sampler/PercentageBasedSamplerTests.java
```diff
@@ -49,19 +49,36 @@ public void should_pass_given_percent_of_samples() throws Exception {
...
 	}
 
 	@Before
 	public void setupSpan() {
-		this.span = Span.builder().traceId(RANDOM.nextLong()).build();
+		this.span = newSpan();
+	}
+
+	Span newSpan() {
+		return Span.builder().traceId(RANDOM.nextLong()).build();
```
adriancole: yep. most importantly the sampler requires trace ids to be random, and.. the sample rate isn't stable until about 100,000 random ids pass through.

This means we might want to consider re-introducing a counting sampler, or making an option to make this one a counting one. If you have a low-traffic server, counting could be appropriate, given the warts.

@eirslett @kristofa I was thinking about this for brave, too.. wdyt?

adriancole:
LGTM

PR #262, opened 2016-04-29
Draining from queue instead of clearing it

### LINKED ISSUE #259: Posible loss of spans by StreamSpanReporter
Hi,

I've been reading the code of StreamSpanReporter and I've seen that poll() and report() methods could have concurrency issues. At the moment just after copying the spans from the queue into the results and just before clearing the queue, a new span could be added to the queue... and that new span will be cleared from the queue without being sent in the payload of the message.

	@InboundChannelAdapter(value = SleuthSource.OUTPUT)
	public Spans poll() {
		List<Span> result = new ArrayList<>(this.queue);
		// At this point of execution report() method could be invoked
		// and the span just added to the queue wil be cleared and will not be sent to
		// the stream
		this.queue.clear();
		...
	}
	
	@Override
	public void report(Span span) {
		if (span.isExportable()) {
			this.queue.add(span);
		}
	}

Shouldn't some kind of sinchronization be done? Something like:

	@InboundChannelAdapter(value = SleuthSource.OUTPUT)
	public Spans poll() {
		List<Span> result = null;
		synchronized(this.queue){
			result = new ArrayList<>(this.queue);
			this.queue.clear();
		}
		...
	}
	
	@Override
	public void report(Span span) {
		if (span.isExportable()) {
			synchronized(this.queue){
				this.queue.add(span);
			}
		}
	}

marcingrzejszczak:
now instead of first passing all the spans from the queue to the list and then clearing it we're now draining it contents. If in the meantime any spans will arrive they will be rained at next passing

fixes #259

### DIFF
--- a/spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/StreamSpanReporter.java
+++ b/spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/StreamSpanReporter.java
@@ -17,10 +17,10 @@
 package org.springframework.cloud.sleuth.stream;
 
 import java.util.ArrayList;
-import java.util.Collection;
 import java.util.Iterator;
 import java.util.List;
-import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.BlockingQueue;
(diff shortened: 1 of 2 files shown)


adriancole:
+1

PR #263, opened 2016-04-29
Updates to zipkin 0.15

adriancole:
Notably, this removes the accidental transient dependency on scala, by
removing the dependency on `zipkin-cassandra-core`. Related classes have
been vendored into `storage-cassandra`.

This also splits out `CollectorSampler` as its own thing and introduces
two options for instrumentation sampling:

* `BoundaryTraceIdSampler` - For high-volume sites, based on Finagle
* `CountingTraceIdSampler` - Trades lower performance for high accuracy


adriancole:
@marcingrzejszczak mind merging this in my absence it would be nice for latest snapshot to fix the kafka conflict @Arghanil found

adriancole:
rats this won't completely fix the scala dep problem. zipkin-ui will need to bump also, after this is complete https://github.com/openzipkin/zipkin/issues/1105

PR #264, opened 2016-05-03
Updates to zipkin 0.16

adriancole:
This important in so far that it allows us to do metrics. The actual
implementation of stream metrics is something we can work on after 1.0
if there's a time crunch.


spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/ZipkinMessageListener.java
```diff
@@ -150,6 +152,11 @@ CollectorSampler collectorSampler() {
 		}
 
 		@Bean
+		CollectorMetrics collectorMetrics() {
```
marcingrzejszczak: `@ConditionalOnMissingBean` ?

marcingrzejszczak:
:shipit:

PR #270, opened 2016-05-06
Updates to zipkin 0.18

adriancole:
Relating to sleuth, this is mostly cosmetic with a kafka condition fix.

It might be of interest that `GET /api/v1/traces` no longer requires
parameters.

https://github.com/openzipkin/zipkin-java/releases/tag/0.18.0


marcingrzejszczak:
:shipit:

PR #272, opened 2016-05-07
Sampling child span

### LINKED ISSUE #271: Sampler not used when parent is null inside DefaultTracer.createChild() method
Reviewing the code of the createChild() method of DefaultTracer class, I've detected that when the parent is null a new span is always created as exportable.

I've got two questions about createChild() method:
- Is this method going to be invoked with parent being null? I mean, should we create a child without a parent? Maybe the method should assert that parent value is not null instead of creating a new span when the parent is null.
- In case we want to admit that createChild() receives a null parent, I think we should use the sampler to decide if the new span is going to be exportable, shouldn't we?

marcingrzejszczak:
when parent is null, the created child span will be sampled

fixes #271

cc @dsyer , @adriancole - check https://github.com/spring-cloud/spring-cloud-sleuth/issues/271

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/trace/DefaultTracer.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/trace/DefaultTracer.java
@@ -82,11 +82,7 @@ public class DefaultTracer implements Tracer {
 			if (sampler==null) {
 				sampler = this.defaultSampler;
 			}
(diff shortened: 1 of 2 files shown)


adriancole:
The idea of a child span with no parent is confusingly close to a root span (aka new trace). Are they the same or is there something subtle I don't get?

marcingrzejszczak:
Renamed the method, and added null check - https://github.com/spring-cloud/spring-cloud-sleuth/commit/657bba35b3cfb3750125fbbcc388179852217885

PR #289, opened 2016-05-25
Issues #286 custom feign config

marcingrzejszczak:
I’m creating a custom `FeignContext` that before returning a bean is wrapping it with its tracing representation. That way we honour the beans created in custom Feign configurations.

@adriancole @spencergibb @dsyer


adriancole:
Neat. Looks like most lines added are new tests. Glad you've figured out how to get more miles out of this

Issue #290, opened 2016-05-26
Feign external url span details missing


singram:
When using feign to call an external url.  eg.
```
@FeignClient(name = "swapi-service", url = "http://swapi.co/api/")
public interface SwapiService {
  @RequestMapping(method = RequestMethod.GET, value = "/people/{personId}")
  ResponseEntity<SwapiPerson> getPerson(@RequestHeader(value="User-Agent", defaultValue="curl/7.45.0") String userAgent, @PathVariable("personId") long personId);
}
```
Details such as 
-   http.host
-   http.method
-   http.path
-   http.url

Are not captured on the span and visible in Zipkin, as shown in other examples such as those found here. http://brewery-zipkin-server.cfapps.io/traces/38efdb9a8861a00a?serviceName=brewing

Using
- Spring-boot => 1.3.5.Release
- Spring-cloud => Brixton.Release

marcingrzejszczak:
CC @adriancole - this is what we were discussing on Gitter. If we have both apps instrumented (client + server) then once the server side creates the span it adds the details. Whereas when we have a case where we call an external service then we lose those tag information.

PR #292, opened 2016-05-30
Issues #290 client tags

marcingrzejszczak:
This PR introduces additional tags on the HTTP client side. Before sending a request we're tagging the client side with HTTP Trace Keys so that if the other side is not working / not instrumented (e.g. call to an external service) then the debugging information remains.

Feign Client Side:
feign

Rest Template Client Side:
client_side

Server Side:
server_side

Related to #290 
CC @dsyer @adriancole

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/HttpTraceKeysInjector.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/HttpTraceKeysInjector.java
@@ -1,0 +1,55 @@
+package org.springframework.cloud.sleuth.instrument.web;
+
+import java.util.Collection;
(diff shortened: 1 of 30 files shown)


adriancole:
I think this is good, except it looks like we are still adding them unconditionally on the server side. This will double the storage as the largest part of a span are tags like this.

I'd recommend condition adding on the server side. When you are joining a trace you know the client was instrumented, so you can avoid adding tags the client should be adding (the request tags ). When you can't,  you are starting a root span and might choose to add tags from the request.

Such an approach will result in win-win, getting all the tags and not halfing storage.

marcingrzejszczak:
Changes following review

Span structure:
spans

Root span:
parent

Child:
child

@adriancole - now it should be better...

adriancole:
cool! thanks for working towards this

PR #296, opened 2016-06-03
Added aspect that closes spans for wrong http responses

### LINKED ISSUE #278: No tracing in /error handling
When there is an error in an HTTP request handler, the tracing stops at the point where the container hands over to an ERROR dispatcher, which might make it hard to investigate errors.

marcingrzejszczak:
with this change trace filter is not closing all spans. It's closing only spans when the response is successful. If the response status is 4xx,5xx then an exception controller should start processing the response. At the end of the day an aspect will close the span once the controller has finished processing.

fixes #278

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
@@ -34,6 +34,7 @@ import org.springframework.cloud.sleuth.Tracer;
 import org.springframework.cloud.sleuth.sampler.NeverSampler;
 import org.springframework.core.Ordered;
 import org.springframework.core.annotation.Order;
+import org.springframework.http.HttpStatus;
 import org.springframework.util.StringUtils;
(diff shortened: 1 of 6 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
```diff
@@ -147,7 +147,10 @@ protected void doFilterInternal(HttpServletRequest request,
 				} else {
 					spanFromRequest.logEvent(Span.SERVER_SEND);
 				}
-				this.tracer.close(spanFromRequest);
+				// in case of a response with exception status a exception controller will close the span
+				if (response.getStatus() >= 200 && response.getStatus() <= 299) {
```
shakuzen: Could also write the following, which may be more readable/clear to some.

```java
if (HttpStatus.valueOf(response.getStatus()).is2xxSuccessful()) {
```

marcingrzejszczak: :+1:

PR #308, opened 2016-06-22
Throwing exception when tracing is malformed

### LINKED ISSUE #306: Malformed X-B3 headers cause 500 error
If either X-B3-TraceId or X-B3-SpanId request header is malformed, Span throws a NumberFormatException.

It seems to me that this should fail gracefully, creating a new SpanId and/or TraceId instead of propagating the NFE and returning 500.

Maybe handle this in HttpServletRequestExtractor::joinTrace?

marcingrzejszczak:
with this change we will throw an IllegalStateException when the malformed tracing data are sent. In TraceFilter we're catching it and sending back 400 response. In case of messaging the exception gets propagated

fixed #306

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
@@ -374,7 +374,7 @@ public class Span {
 		try {
 			return new BigInteger(hexString, 16).longValue();
 		} catch (NumberFormatException e) {
-			throw new IllegalArgumentException("Malformed id", e);
+			throw new IllegalArgumentException("Malformed id [" + hexString + "]", e);
 		}
 	}
 
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
@@ -371,7 +371,11 @@ public class Span {
 	 */
 	public static long hexToId(String hexString) {
 		Assert.hasText(hexString, "Can't convert empty hex string to long");
-		return new BigInteger(hexString, 16).longValue();
+		try {
+			return new BigInteger(hexString, 16).longValue();
+		} catch (NumberFormatException e) {
+			throw new IllegalArgumentException("Malformed id", e);
+		}
 	}
 
 	@Override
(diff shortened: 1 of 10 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/MessagingSpanExtractor.java
```diff
@@ -100,7 +91,7 @@ private void setParentIdIfApplicable(Message<?> carrier, SpanBuilder spanBuilder
 				spanBuilder.parent(Span.hexToId(parentId));
 			}
 		} catch (Exception e) {
-			log.warn("Exception occurred while trying to set parentId", e);
+			throw new IllegalStateException("Malformed id", e);
```
adriancole: prefer IllegalArgumentException as IllegalStateException is more temporal in nature (ex a state issue that could resolve itself)

the catch Exception is a bit wide, too.. one way could be to make the Span.hexToId throw IllegalArgumentException, and propagate all IllegalArgumentExceptions instead of wrapping them

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
```diff
@@ -152,7 +152,13 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
 		}
 		addToResponseIfNotPresent(response, Span.SAMPLED_NAME, skip ? Span.SPAN_NOT_SAMPLED : Span.SPAN_SAMPLED);
 		String name = HTTP_COMPONENT + ":" + uri;
-		spanFromRequest = createSpan(request, skip, spanFromRequest, name);
+		try {
+			spanFromRequest = createSpan(request, skip, spanFromRequest, name);
+		} catch (IllegalStateException e) {
+			filterChain.doFilter(request, response);
+			response.sendError(HttpStatus.BAD_REQUEST.value(), "Tracing data is malformed");
```
adriancole: I'd re-use the exception message here, as it could be written in such a way so that what to change is clear.

Ex. ` "Exception tracing request " + e.getMessage()`

adriancole:
some polishing notes.. positive change!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
```diff
@@ -371,7 +371,11 @@ public static String idToHex(long id) {
 	 */
 	public static long hexToId(String hexString) {
 		Assert.hasText(hexString, "Can't convert empty hex string to long");
-		return new BigInteger(hexString, 16).longValue();
+		try {
+			return new BigInteger(hexString, 16).longValue();
+		} catch (NumberFormatException e) {
+			throw new IllegalArgumentException("Malformed id", e);
```
adriancole: maybe add the hexString to the exception message?

adriancole:
one nit, otherwise :shipit:

PR #311, opened 2016-06-23
Renaming child server span

marcingrzejszczak:
The change results in converting the artificial span spawned from the first http server span into a span that is generated from an aspect that wraps a controller that processes the request.

cc @dsyer


adriancole:
LGTM tested with a demo app, and the result is 100x more intuitive since the local spans are named based on the calling method

screen shot 2016-06-23 at 6 15 11 pm

https://github.com/adriancole/sleuth-webmvc-example

marcingrzejszczak:
I obviously by mistake have also added some missing zuul related commits from master.

PR #314, opened 2016-06-24
Add basic auth support to HttpZipkinSpanReporter

brockovercash:
I would like to use Basic Auth to make my Zipkin server a little more secure and needed to implement this into the SpanReporter. This is definitely a simple solution and I see some potential for refactoring the Auth element into a class so that things like " Basic b64token" string only has to be created once. I'll update the PR if I do that.

### DIFF
--- a/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/HttpZipkinSpanReporter.java
+++ b/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/HttpZipkinSpanReporter.java
@@ -17,6 +17,9 @@ import java.util.concurrent.LinkedBlockingQueue;
 import java.util.concurrent.ScheduledExecutorService;
 
 import java.util.zip.GZIPOutputStream;
+
+import javax.xml.bind.DatatypeConverter;
+
 import org.apache.commons.logging.Log;
 import org.springframework.cloud.sleuth.metric.SpanMetricReporter;
 
@@ -29,6 +32,7 @@ import static java.util.concurrent.TimeUnit.SECONDS;
  * Submits spans using Zipkin's {@code POST /spans} endpoint.
  *
  * @author Adrian Cole
+ * @author Brock Overcash
  * @since 1.0.0
  */
 public final class HttpZipkinSpanReporter
@@ -38,6 +42,9 @@ public final class HttpZipkinSpanReporter
 	private static final Charset UTF_8 = Charset.forName("UTF-8");
 
 	private final String url;
+	private final boolean basicAuthenticationEnabled;
+	private final String username;
+	private final String password;
 	private final BlockingQueue<Span> pending = new LinkedBlockingQueue<>(1000);
 	private final Flusher flusher; // Nullable for testing
 	private final boolean compressionEnabled;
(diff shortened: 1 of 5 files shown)


spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/HttpZipkinSpanReporter.java
```diff
@@ -17,6 +17,9 @@
 import java.util.concurrent.ScheduledExecutorService;
 
 import java.util.zip.GZIPOutputStream;
+
+import javax.xml.bind.DatatypeConverter;
```
spencergibb: Use Spring's `Base64Utils`

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinProperties.java
```diff
@@ -22,13 +22,17 @@
...
+ * @author Brock Overcash
  * @since 1.0.0
  */
 @ConfigurationProperties("spring.zipkin")
 public class ZipkinProperties {
 	/** URL of the zipkin query server instance. */
 	private String baseUrl = "http://localhost:9411/";
 	private boolean enabled = true;
+	private boolean basicAuthenticated = false;
+	private String username = "admin";
```
spencergibb: Should not have a default.

brockovercash: Which lines specifically? I think basicAuthenticated should default to false, but I agree username and password don't need defaults. I included them because all of the others had defaults.

spencergibb: Username and password. Null is fine.

marcingrzejszczak:
@adriancole - any feedback on this?

adriancole:
sure.. here's my feedback :)

If we consider this, we should also for a sec consider what is the more spring cloud way of addressing outbound HTTP auth. This doesn't mean we don't special-case BASIC, just we should think a little.

For starters, user+password combos change, so it might be cleaner to do something like this with an interceptor. If there are other implicit http clients in the classpath of sleuth, I'd bet there's a less hard-coded, and more spring way to accomplish this.

For example, zipkin doesn't have basic auth, and if it were me, I'd probably use something like expirable API keys or OAuth token as opposed to passwords. Even feign doesn't hard-code basic auth, rather bundles an interceptor to do that.. Is there another option we could use?

adriancole:
ex. one option could be to make a Feign, OkHttp or otherwise reporter for those who want to customize things, using their interceptors. It could be auto-configured.

This would allow us to leave the basic (no pun intended) one simple, and not run down a slope of adding little configurations here and there for the somewhat knee-capped interface of HttpUrlConnection.

brockovercash:
I'm not well versed in much of the Spring ecosystem. What looks like the best way to do this from an integration perspective with the Sleuth libs? Spring-cloud-sleuth-zipkin-feign or work in the sleuth-zipkin project?

marcingrzejszczak:
@brockovercash I took another look at the problem and did sth like this: https://github.com/spring-cloud/spring-cloud-sleuth/pull/316 . I've abstracted the call to Zipkin via a RestTemplate.

marcingrzejszczak:
Closing this one in favour of #316 . Thanks @brockovercash for the idea and feedback!!

PR #316, opened 2016-06-27
Issues #314 basic auth

marcingrzejszczak:
@dsyer @spencergibb @adriancole


brockovercash:
Seems like a dumb question, but will this generate an interception loop since the RestClient is an instrumented library for Sleuth? What prevents it from doing this?

marcingrzejszczak:
That's a good question! I'm creating `RestTemplate` manually - https://github.com/spring-cloud/spring-cloud-sleuth/pull/316/files#diff-0780dc662db6ad165496f36c421ae346R57 via `new`. That's why it won't get instrumented.

brockovercash:
Makes sense! This looks like a better and more extensible solution to me then.

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/HttpZipkinSpanReporter.java
```diff
@@ -128,34 +133,23 @@ public void run() {
 		}
 	}
 
-	void postSpans(byte[] json) throws IOException {
-		// intentionally not closing the connection, so as to use keep-alives
-		HttpURLConnection connection = (HttpURLConnection) new URL(this.url).openConnection();
-		connection.setRequestMethod("POST");
-		connection.addRequestProperty("Content-Type", "application/json");
+	void postSpans(byte[] json) throws Exception {
```
adriancole: RestTemplate throws Exception (vs Runtime and an IOException)?

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/HttpZipkinSpanReporter.java
```diff
@@ -128,34 +133,23 @@ public void run() {
...
+		HttpHeaders httpHeaders = new HttpHeaders();
+		httpHeaders.setContentType(MediaType.APPLICATION_JSON);
+		json = compressRequestIfApplicable(json);
+		RequestEntity<byte[]> requestEntity = new RequestEntity<>(json, httpHeaders, HttpMethod.POST, URI.create(this.url));
+		this.restTemplate.exchange(requestEntity, String.class);
+	}
+
+	private byte[] compressRequestIfApplicable(byte[] json) throws IOException {
 		if (this.compressionEnabled) {
-			connection.addRequestProperty("Content-Encoding", "gzip");
```
adriancole: why don't you need to add content-encoding when gzipping?

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/DefaultZipkinRestTemplateCustomizer.java
```diff
@@ -0,0 +1,71 @@
...
+	public DefaultZipkinRestTemplateCustomizer(
+			ZipkinProperties zipkinProperties) {
+		this.zipkinProperties = zipkinProperties;
+	}
+
+	@Override
+	public void customize(RestTemplate restTemplate) {
+		restTemplate.setRequestFactory(new DefaultZipkinConnectionFactory(this.zipkinProperties));
+		if (this.zipkinProperties.isBasicAuthenticated()) {
+			restTemplate.getInterceptors().add(
```
adriancole: why make a new type? wouldn't it be less code to manage to tell users to add their own Interceptors? Ex basic auth is not necessarily going to be a normal case on apis (ex AWS etc don't use password auth for very good reasons).

I think the goal is to make basic auth *possible*, not make it standard

spring-cloud-sleuth-zipkin/pom.xml
```diff
@@ -38,6 +38,10 @@
 			<artifactId>spring-cloud-sleuth-core</artifactId>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-web</artifactId>
```
adriancole: tradeoff here is that this adds a mandatory dependency. The existing class could have been refactored to allow a second reporter to be used (which would make this optional). I don't know the dependency hygiene, but food for thought.

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/DefaultZipkinConnectionFactory.java
```diff
@@ -0,0 +1,49 @@
...
+		this.zipkinProperties = zipkinProperties;
+		setBufferRequestBody(false);
+	}
+
+	@Override
+	protected void prepareConnection(HttpURLConnection connection, String httpMethod)
+			throws IOException {
+		super.prepareConnection(connection, httpMethod);
+		if (this.zipkinProperties.getCompression().isEnabled()) {
+			connection.addRequestProperty("Content-Encoding", "gzip");
```
adriancole: this seems very indirect. just add the header when you are compressing the content?

adriancole: if there's some weird technical reason why, I'd comment since adding more classes = more to maintain

spring-cloud-sleuth-zipkin/pom.xml
marcingrzejszczak: Since this is a dependency to our Spring module that shouldn't  be a big problem

spring-cloud-sleuth-zipkin/src/test/java/org/springframework/cloud/sleuth/zipkin/HttpZipkinSpanReporterTest.java
```diff
@@ -18,11 +19,11 @@
...
 	SpanMetricReporter spanMetricReporter = new CounterServiceBasedSpanMetricReporter("accepted", "dropped",
 			this.inMemorySpanCounter);
+	RestTemplate restTemplate = defaultRestTemplate();
 
-	HttpZipkinSpanReporter reporter = new HttpZipkinSpanReporter(
-			this.zipkin.httpUrl(),
+	HttpZipkinSpanReporter reporter = new HttpZipkinSpanReporter(restTemplate, this.zipkin.httpUrl(),
 			0, // so that tests can drive flushing explicitly
-			false, // disable compression
+			false,
```
adriancole: nit: I'd leave the comments as they say what false means

spring-cloud-sleuth-zipkin/src/test/java/org/springframework/cloud/sleuth/zipkin/HttpZipkinSpanReporterTest.java
```diff
@@ -68,10 +69,9 @@ public void postsSpans() throws Exception {
 
 	@Test
 	public void postsCompressedSpans() throws Exception {
-		this.reporter = new HttpZipkinSpanReporter(
-				this.zipkin.httpUrl(),
+		this.reporter = new HttpZipkinSpanReporter(restTemplateWithCompression(), this.zipkin.httpUrl(),
 				0, // so that tests can drive flushing explicitly
-				false, // enable compression
+				true,
```
adriancole: ditto here

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/DefaultZipkinRestTemplateCustomizer.java
```diff
@@ -0,0 +1,64 @@
...
+	private final ZipkinProperties zipkinProperties;
+
+	public DefaultZipkinRestTemplateCustomizer(
+			ZipkinProperties zipkinProperties) {
+		this.zipkinProperties = zipkinProperties;
+	}
+
+	@Override
+	public void customize(RestTemplate restTemplate) {
+		if (this.zipkinProperties.getCompression().isEnabled()) {
```
adriancole: I think it would be surprising to have compression nixed when you customize something. It makes this part of the properties special. IOTW I don't think there should be a default customizer, as this compression logic should happen regardless of if someone added a "customizer" or not. There's already a way to not have compression added (which is to set the property false).

spring-cloud-sleuth-zipkin/src/test/java/org/springframework/cloud/sleuth/zipkin/HttpZipkinSpanReporterTest.java
marcingrzejszczak: I just removed it cause it's not necessary

spring-cloud-sleuth-zipkin/src/test/java/org/springframework/cloud/sleuth/zipkin/HttpZipkinSpanReporterTest.java
marcingrzejszczak: I just removed it cause it's not necessary

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/DefaultZipkinRestTemplateCustomizer.java
marcingrzejszczak: Hmm I don't think I agree here. From the code's perspective (the `HttpZipkinSpanReporter`) it's just sending spans. Whether they are compressed or not - he shouldn't care. It's a technical decision of the person configuring the `RestTemplate` so the Http Client. At the moment the default customization of the HTTP client is that it creates a compression mechanism if there is a property. But someone can override the RestTemplate with any sort of an approach he wants to. 

I can agree that we could create a  `RestTemplate` that would by default have the compression mechanism turned on and then the customizer can disable it. But I'd like to stick with the separation of concerns.

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinRestTemplateCustomizer.java
```diff
@@ -0,0 +1,31 @@
...
+import org.springframework.web.client.RestTemplate;
+
+/**
+ * Customizes {@link RestTemplate} used to send Spans to Zipkin
+ *
+ * @author Marcin Grzejszczak
+ *
+ * @since 1.1.0
+ */
+public interface ZipkinRestTemplateCustomizer {
```
adriancole: why we are making an interface vs a qualifier on `ClientHttpRequestInterceptor` should be documented, if retained, but really it shouldn't exist.

As mentioned before, I don't like adding new interfaces to expose activity that's already abstracted.

This is a wider interface than necessary, makes re-using `ClientHttpRequestInterceptor` implementations more difficult, and makes searching for help harder. For example, if I google `ZipkinRestTemplateCustomizer`, I'll find nothing. If I google `ClientHttpRequestInterceptor` I'll find immediately how to do anything from authenticate to log outbound requests.

adriancole:
feel free to merge when you feel this is ready.

I've made comments about some things that I don't agree with, but you are the champion of this project and should make the call on matters like this. Best is subjective :)

spring-cloud-sleuth-zipkin/src/test/java/org/springframework/cloud/sleuth/zipkin/HttpZipkinSpanReporterTest.java
adriancole: not necessary if the reader is expected to know that this parameter index
controls compression.. basically the comment was for the reader, not the
writer. If I didn't comment this, you might haven't as easily noticed a bug
later, right?

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/DefaultZipkinRestTemplateCustomizer.java
adriancole: how is the user supposed to know that the compression property is special?
that it applies unless someone added a customizer? There's no mention of
this caveat unless you read the code.

they're most likely to find out when they notice larger data going across
the network. If you feel strongly that this is a valid separation of
concerns, I'd encourage you to document it, so that the extender knows they
are expected to implement the compression property when customizing.

marcingrzejszczak:
Ok @brockovercash so we had another round of discussions here and we've changed the implementation a little bit. Since Zipkin has no notion of basic authentication and it's not really typical for metric systems to authenticate their requests we've decided to remove the basic authentication from this PR. What we did however is that we added a possibility to customize the `RestTemplate`. There is the `ZipkinRestTemplateCustomizer` and the `DefaultZipkinRestTemplateCustomizer` that gives you the compression. So you could create your own implementation of `ZipkinRestTemplateCustomizer` where you would add basic authentication by adding a basic authentication interceptor. 

For now I guess you'd have to write the interceptor by yourself but there is an issue to move that to Spring - https://github.com/spring-projects/spring-boot/issues/6237 . Here you have an example of the interceptor - https://github.com/spring-projects/spring-boot/blob/master/spring-boot/src/main/java/org/springframework/boot/web/client/BasicAuthorizationInterceptor.java

brockovercash:
Thanks guys!

PR #317, opened 2016-06-29
Exposes getAccumulatedMicros for more precise Span duration

### LINKED ISSUE #312: More precise duration
Sleuth logs "local spans" for in-process activity (ex things like async methods that often complete in less than 2 milliseconds). I've found it difficult to debug sleuth itself on account of Span having only milliseconds precision. For example, some tasks end up as zero milliseconds, and others look longer than they should (ex 2 millis when it should be less than one). While we can use btrace or byteman to help diagnose local perf issues, seems we can kill 2 birds with one stone by changing some units to microseconds.

Durations of local spans are better made in smaller order (like microseconds), as you can more easily compare equivalent tasks. Moreover, we should avoid making end markers with calls to System.currentTimeMillis as it cannot show sub-millis differences and also has a risk of moving backwards.

Here's pseudocode of what we could do to alter the impl to use higher precision:

* when existing field Span.begin is initialized, also initialize startNanos with System.nanoTime
* at Span.stop() set endNanos = System.nanoTime
  * add new field or method: Span.duration() or durationMicros() =  (endNanos - startNanos) * 1000
  * direct existing method Span.getAccumulatedMillis to durationMicros() / 1000

Footnote on choosing micros instead of nanos for duration:

I'd suggest not advertising duration in nanoseconds because we cannot accurately measure at that granularity due to the [overhead and change frequency of System.nanoTime()](http://shipilev.net/blog/2014/nanotrusting-nanotime/#_backoff_measurements). For example, if calling nanoTime has to happen twice, and we should expect 30nanos overhead per call, we can't reliably report duration in orders of nanos. Moreover, eventhough unix can report time in microseconds using the presumably cheaper [gettimeofday](http://man7.org/linux/man-pages/man2/gettimeofday.2.html), it cannot in nanos. In short, micros ends up a pretty decent choice for various reasons even if we didn't consider zipkin which uses micros consistently for everything.

adriancole:
Formerly, `Span.getAccumulatedMillis()` worked, but could returns
imprecise measurements, particularly local spans. This changes the
internals of Span to keep track of a start tick. Using this, it can
return a more precise `Span.getAccumulatedMicros()`.

To ensure this precision isn't lost in serialization, this adds a
json field `durationMicros`, which is only set when the span is stopped.
This is set instead of start tick because `System.nanoTime()` is JVM
specific and so cannot be used across the network. `durationMicros` uses
null instead of zero comparisons because nano time can be negative.

Fixes #312

Example cold trace from https://github.com/adriancole/sleuth-webmvc-example/
screen shot 2016-06-29 at 12 13 57 pm

Example warm trace from https://github.com/adriancole/sleuth-webmvc-example/
screen shot 2016-06-29 at 12 14 23 pm

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
@@ -145,6 +145,11 @@ public class Span {
 	private final List<Log> logs;
 	private final Span savedSpan;
 
+	// Null means we don't know the start tick, so fallback to time
+	@JsonIgnore
+	private final Long startNanos;
+	private Long durationMicros; // serialized in json so micros precision isn't lost
+
 	@SuppressWarnings("unused")
 	private Span() {
 		this(-1, -1, "dummy", 0, Collections.<Long>emptyList(), 0, false, false, null);
@@ -167,6 +172,8 @@ public class Span {
 		this.processId = current.getProcessId();
 		this.tags = current.tags;
 		this.logs = current.logs;
+		this.startNanos = current.startNanos;
+		this.durationMicros = current.durationMicros;
 		this.savedSpan = savedSpan;
 	}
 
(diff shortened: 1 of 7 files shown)


adriancole:
cc @dsyer @marcingrzejszczak @shakuzen

PR #322, opened 2016-07-04
Re-added missing server side span

### LINKED ISSUE #321: Missing server side span
After introducing in SR2 an aspect that wraps controllers in a new span, in some cases a server side span is missing.

This issue has been confirmed with Spring Data REST only at the moment. It's related to a missing local span on the server side. That's why you won't see an additional span in the Zipkin UI.

marcingrzejszczak:
with this change

- HandlerInterceptor is responsible for wrapping requests in spans
- it will not wrap error controller related requests with spans
- tests have been changed to ensure that ALWAYS there is at least one span on the server side (that way it will show up in Zipkin)
- "fallback" mechanism has been added that if a HandlerInterceptor hasn't been used then we are forcing creation of a Span at the server side

fixes #321

cc @dsyer @spencergibb @adriancole

### DIFF
--- a/spring-cloud-sleuth-core/pom.xml
+++ b/spring-cloud-sleuth-core/pom.xml
@@ -75,6 +75,12 @@
 			<artifactId>rxjava</artifactId>
 			<optional>true</optional>
 		</dependency>
+		<!-- Instrumentation of the custom Spring Data REST HandlerInterceptors -->
+		<dependency>
+			<groupId>org.springframework.data</groupId>
+			<artifactId>spring-data-rest-webmvc</artifactId>
+			<optional>true</optional>
+		</dependency>
 		<dependency>
 			<groupId>org.aspectj</groupId>
 			<artifactId>aspectjrt</artifactId>
@@ -113,6 +119,26 @@
 		<dependency>
 			<groupId>pl.pragmatists</groupId>
 			<artifactId>JUnitParams</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-data-jpa</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-orm</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-data-rest</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>com.h2database</groupId>
+			<artifactId>h2</artifactId>
 			<scope>test</scope>
 		</dependency>
 	</dependencies>
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
@@ -128,26 +128,14 @@ public class TraceFilter extends GenericFilterBean {
 				|| Span.SPAN_NOT_SAMPLED.equals(ServletUtils.getHeader(request, response, Span.SAMPLED_NAME));
 		Span spanFromRequest = getSpanFromAttribute(request);
 		if (spanFromRequest != null) {
-			this.tracer.continueSpan(spanFromRequest);
-			if (log.isDebugEnabled()) {
-				log.debug("There has already been a span in the request " + spanFromRequest + "");
-			}
+			continueSpan(request, spanFromRequest);
 		}
 		if (log.isDebugEnabled()) {
 			log.debug("Received a request to uri [" + uri + "] that should be skipped [" + skip + "]");
 		}
 		// in case of a response with exception status a exception controller will close the span
 		if (!httpStatusSuccessful(response) && isSpanContinued(request)) {
-			if (log.isDebugEnabled()) {
-				log.debug(
-						"The span was already detached once and we're processing an error");
-			}
-			try {
-				filterChain.doFilter(request, response);
-			} finally {
-				request.setAttribute(TRACE_ERROR_HANDLED_REQUEST_ATTR, true);
-				this.tracer.close(spanFromRequest);
-			}
+			processErrorRequest(filterChain, request, response, spanFromRequest);
 			return;
 		}
 		addToResponseIfNotPresent(response, Span.SAMPLED_NAME, skip ? Span.SPAN_NOT_SAMPLED : Span.SPAN_SAMPLED);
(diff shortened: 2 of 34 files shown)


marcingrzejszczak:
Sth's wrong with the parent - child structure still so don't merge it yet.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
```diff
@@ -166,33 +167,51 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
...
 			addResponseTags(response, exception);
-			if (spanFromRequest.hasSavedSpan()) {
-				closeParentSpan(spanFromRequest.getSavedSpan());
+			if (spanFromRequest.hasSavedSpan() && requestHasAlreadyBeenHandled(request)) {
+				reportParentSpan(spanFromRequest.getSavedSpan());
+			} else if (!requestHasAlreadyBeenHandled(request)) {
+				spanFromRequest = this.tracer.close(spanFromRequest);
 			}
-			closeParentSpan(spanFromRequest);
+			reportParentSpan(spanFromRequest);
```
adriancole: nit: probably record vs report. reporting happens asynchronously

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHandlerInterceptor.java
```diff
@@ -0,0 +1,172 @@
...
+	public boolean preHandle(HttpServletRequest request, HttpServletResponse response,
+			Object handler) throws Exception {
+		if (isErrorControllerRelated(request)) {
+			if (log.isDebugEnabled()) {
+				log.debug("Skipping creation of a span for error controller processing");
+			}
+			return true;
+		}
+		String spanName = spanName(handler);
+		Span span = getTracer().createSpan(SpanNameUtil.toLowerHyphen(spanName));
```
adriancole: nit: toLowerHyphen might be best inside the spanName function

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceSpringDataBeanPostProcessor.java
```diff
@@ -0,0 +1,88 @@
...
+			HandlerExecutionChain handlerExecutionChain = this.delegate.getHandler(request);
+			if (handlerExecutionChain == null) {
+				return null;
+			}
+			handlerExecutionChain.addInterceptor(new TraceHandlerInterceptor(TraceSpringDataBeanPostProcessor.this.beanFactory));
+			return handlerExecutionChain;
+		}
+	}
+
+	private static List<HandlerMapping> mappings() {
```
adriancole: intentional?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/util/SpanNameUtil.java
```diff
@@ -29,12 +29,22 @@ public static String toLowerHyphen(String name) {
 		StringBuilder result = new StringBuilder();
 		for (int i = 0; i < name.length(); i++) {
 			char c = name.charAt(i);
-			if (c >= 'A' && c <= 'Z') {
-				result.append('-').append((char) (c + 'a' - 'A'));
+			if (i > 0 && isUpercase(c)) {
```
adriancole: misspelled. plus why are we leaving the first character capitalized?

adriancole: oh I think I see why. bc this assumed the input was in lower-camel, but it is now used for upper-camel, too.

another way could be to special case the first character outside the loop and initialize `int i = 1`, leaving the rest alone.

spring-cloud-sleuth-core/pom.xml
```diff
@@ -76,6 +76,11 @@
 			<optional>true</optional>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.data</groupId>
+			<artifactId>spring-data-rest-webmvc</artifactId>
+			<optional>true</optional>
```
adriancole: style, but usually I comment what class or purpose optional deps are needed for, when adding. Helps with any future cleanup or questions.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceSpringDataBeanPostProcessor.java
```diff
@@ -0,0 +1,88 @@
...
+		return bean;
+	}
+
+	@Override
+	public Object postProcessAfterInitialization(Object bean, String beanName)
+			throws BeansException {
+		return bean;
+	}
+
+	class TraceDelegatingHandlerMapping extends DelegatingHandlerMapping {
```
adriancole: when inner classes don't read fields from their enclosing class, mark them static.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHandlerInterceptor.java
```diff
@@ -0,0 +1,172 @@
...
+			return;
+		}
+		getTracer().close(getSpanFromAttribute(request));
+	}
+
+	private Span getSpanFromAttribute(HttpServletRequest request) {
+		return (Span) request.getAttribute(TraceFilter.TRACE_HANDLED_REQUEST_ATTR);
+	}
+
+	private Span getRootSpanFromAttribute(HttpServletRequest request) {
```
adriancole: the names of constants like this and above aren't intuitively related to the method names. Is `TRACE_REQUEST_ATTR` used for something besides a root span? if not, maybe name it so we don't need translation methods?

adriancole:
I don't really follow a lot of this as I'm inexperienced with the complexities around handlers and various use cases. I only made comments on code.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHandlerInterceptor.java
marcingrzejszczak: I'm trying no to do backwards incompatible changes in this branch, but will definitely address that in `master`

spring-cloud-sleuth-core/pom.xml
```diff
@@ -75,6 +75,12 @@
 			<artifactId>rxjava</artifactId>
 			<optional>true</optional>
 		</dependency>
+		<!-- Instrumentation of the custom Spring Data REST HandlerInterceptors -->
```
adriancole: thx

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
```diff
@@ -166,33 +154,75 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
...
 			addToResponseIfNotPresent(response, Span.SAMPLED_NAME, skip ? Span.SPAN_NOT_SAMPLED : Span.SPAN_SAMPLED);
 			detachOrCloseSpans(request, response, spanFromRequest, exception);
 		}
 	}
 
+	private void processErrorRequest(FilterChain filterChain, HttpServletRequest request,
+			HttpServletResponse response, Span spanFromRequest)
+			throws IOException, ServletException {
+		if (log.isDebugEnabled()) {
+			log.debug("The span was already detached once and we're processing an error");
```
adriancole: maybe add the span discussed to the debug message?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
```diff
@@ -166,33 +154,75 @@ public void doFilter(ServletRequest servletRequest, ServletResponse servletRespo
...
+			request.setAttribute(TRACE_ERROR_HANDLED_REQUEST_ATTR, true);
+			this.tracer.close(spanFromRequest);
+		}
+	}
+
+	private void continueSpan(HttpServletRequest request, Span spanFromRequest) {
+		this.tracer.continueSpan(spanFromRequest);
+		request.setAttribute(TraceRequestAttributes.SPAN_CONTINUED_REQUEST_ATTR, "true");
+		if (log.isDebugEnabled()) {
+			log.debug("There has already been a span in the request " + spanFromRequest + "");
```
adriancole: nit: accidental concat of empty string

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHandlerInterceptor.java
```diff
@@ -0,0 +1,191 @@
...
+	public TraceHandlerInterceptor(BeanFactory beanFactory) {
+		this.beanFactory = beanFactory;
+	}
+
+	@Override
+	public boolean preHandle(HttpServletRequest request, HttpServletResponse response,
+			Object handler) throws Exception {
+		if (isErrorControllerRelated(request)) {
+			if (log.isDebugEnabled()) {
+				log.debug("Skipping creation of a span for error controller processing");
```
adriancole: nit: no need to branch when logging a constant.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/util/SpanNameUtil.java
```diff
@@ -20,21 +20,38 @@
...
 
 	public static String toLowerHyphen(String name) {
 		StringBuilder result = new StringBuilder();
-		for (int i = 0; i < name.length(); i++) {
+		appendFirstChar(name, result);
+		for (int i = 1; i < name.length(); i++) {
 			char c = name.charAt(i);
-			if (c >= 'A' && c <= 'Z') {
+			if (isUppercase(c)) {
 				result.append('-').append((char) (c + 'a' - 'A'));
```
adriancole: looking at all the changes here.. in hindsight, I'd special case. Following advice is just what I'm thinking, not what you should do. You can leave the code as is, etc.

ex from the original code, replace

```java
result.append('-').append((char) (c + 'a' - 'A'));
```

with

```java
if (i != 0) result.append('-'); // special-case first character
result.append((char) (c + 'a' - 'A'));
```

That's only a 2 line change (net one additional line) as opposed to a 23 line change. In my mind at least, it accomplishes the same goal.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/SpringDataInstrumentationTests.java
```diff
@@ -0,0 +1,177 @@
...
+	@Autowired Tracer tracer;
+	@Autowired ArrayListSpanAccumulator arrayListSpanAccumulator;
+
+	@Test
+	public void should_create_span_instrumented_by_a_handler_interceptor() {
+		Collection<String> names = names();
+
+		then(names).isNotEmpty();
+		then(this.arrayListSpanAccumulator.getSpans()).isNotEmpty();
+		then(new ListOfSpans(this.arrayListSpanAccumulator.getSpans())).hasASpanWithTagKeyEqualTo("http.class.name");
```
adriancole: maybe assert on the span name (since there's special code around that)

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/TraceFilterIntegrationTests.java
```diff
@@ -119,6 +120,8 @@ public void should_add_a_custom_tag_to_the_span_created_in_controller() throws E
 				.filter(span -> span.tags().containsKey("tag")).findFirst();
 		then(taggedSpan.isPresent()).isTrue();
 		then(taggedSpan.get()).hasATag("tag", "value");
+		then(taggedSpan.get()).hasATag("http.class.method", "deferred");
```
adriancole: I'd probably not use http namespace, since this isn't http (like http.method).

maybe mvc.controller.method or similar?

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/TraceFilterTests.java
```diff
@@ -158,8 +158,10 @@ public void startsNewTraceWithParentIdInHeaders() throws Exception {
 
 		filter.doFilter(this.request, this.response, this.filterChain);
 
-		// this creates a child span which is why we'd expect the parents to include 1L)
-		assertThat(this.span.getParents()).containsOnly(3L);
+		// this creates a child span which is why we'd expect the parents to include the parent id
+		// especially important if no handler interceptors have been used. Then we need to have
+		// an additional span created on the server side so the call graph in Zipkin looks good
```
adriancole: "so the call graph in Zipkin looks good" is something that can be pinned down a bit better. The harder-hitting code comment could be "We add a child span on the server side to show which controller serviced the request, and for how long"

adriancole:
looks like this wasn't easy :)

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/TraceFilterIntegrationTests.java
marcingrzejszczak: I'm bad with naming and `mvc` sounds really good

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/SpringDataInstrumentationTests.java
marcingrzejszczak: Actually I'll assert on both

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/util/SpanNameUtil.java
marcingrzejszczak: Makes sense

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/TraceFilterTests.java
marcingrzejszczak: obviously you're right

PR #326, opened 2016-07-09
Rounds durations of less than zero to one

adriancole:
Duration of 0 is confusing to plot and can be misinterpreted as null.

See https://github.com/openzipkin/zipkin/issues/1174


adriancole:
note: I've not seen this in sleuth, as usually overhead is larger than a microsecond. Just catching an edge case.

marcingrzejszczak:
:shipit:

PR #335, opened 2016-07-14
Fix headers in messaging

### LINKED ISSUE #332: Fix headers in messaging
Currently in messaging we are setting headers with `-` which are not compatible with JMS for instance.

The solution to this is to add camel cased messaging headers and be tolerant readers. First read from the old headers, and if the entry is not there then read from camel case ones.

Also `MessagingSpanExtractor` should be package scoped. `MessagingSpanInjector#SPAN_HEADER` should be package scoped too.

cc @adriancole @dsyer

marcingrzejszczak:
with this change the span related headers are changed to be compatible with messaging specs. The change is backwards compatible so we accept the old ones as input, and as output we're producing both new and old headers.

fixes #332

@dsyer @adriancole


mariuszs:
http://docs.oracle.com/javaee/7/api/javax/jms/Message.html

> Property names must obey the rules for a message selector identifier. 
> Identifiers:
> An identifier is an unlimited-length sequence of letters and digits, the first of which must be a letter. A letter is any character for which the method Character.isJavaLetter returns true. This includes '_' and '$'. A letter or digit is any character for which the method Character.isJavaLetterOrDigit returns true.
> Identifiers cannot be the names NULL, TRUE, and FALSE.
> Identifiers cannot be NOT, AND, OR, BETWEEN, LIKE, IN, IS, or ESCAPE.
> Identifiers are either header field references or property references. The type of a property value in a message selector corresponds to the type used to set the property. If a property that does not exist in a message is referenced, its value is NULL.
> The conversions that apply to the get methods for properties do not apply when a property is used in a message selector expression. For example, suppose you set a property as a string value, as in the following:
> myMessage.setStringProperty("NumberOfOrders", "2");
> The following expression in a message selector would evaluate to false, because a string cannot be used in an arithmetic expression:
> "NumberOfOrders > 1"
> Identifiers are case-sensitive.
> Message header field references are restricted to JMSDeliveryMode, JMSPriority, JMSMessageID, JMSTimestamp, JMSCorrelationID, and JMSType. JMSMessageID, JMSCorrelationID, and JMSType values may be null and if so are treated as a NULL value.
> Any name beginning with 'JMSX' is a JMS defined property name.
> Any name beginning with 'JMS_' is a provider-specific property name.
> Any name that does not begin with 'JMS' is an application-specific property name.

adriancole:
thx this was the sort of "See link" I was interested in.

Issue #323, opened 2016-07-05
Ignorning Hystrix Stream traces when using Spring Cloud Sleuth?


steve-oakey:
I currently have an application that is streaming Hystrix metrics via Spring Cloud Hystrix Stream. When I am using Sleuth in this project, I get traces for both the `@Scheduled` task in `HystrixStreamTask` as well as the integration instrumentation for the `hystrixStreamOutput` `MessageChannel` output. 

I am able to ignore both of these by setting `skipPattern` for the `@Scheduled` task, and disabling the intergration support. The `skipPattern` is not an  issue, but if I have Integration flows that I want tracing for, these would be ignored. I see no way to ignore the tracing only for the `hystrixStreamOutput` channel.

Is there plans to add support for the `patterns` attribute of `@GloablChannelInterceptor` in `TraceSpringIntegrationAutoConfiguration`? Not sure if this annotation accepts a SPEL expression or not.

marcingrzejszczak:
Hey! Sorry for the delay. Currently we didn't have that plan but PRs are more than welcome :)

steve-oakey:
Just as an aside, this PR adds the support for the `pattern` attribute, but it only supports simple patterns so it doesn't appear you can use any sort of negation pattern since it only supports patterns matched by [PatternMatchUtils.simpleMatch(String, String)](https://github.com/spring-projects/spring-framework/blob/master/spring-core/src/main/java/org/springframework/util/PatternMatchUtils.java#L36). So it won't really help to ignore a single channel. Nonetheless, here is the contribution. A change of title for this issue may be appropriate.

Thanks for all the hard work, this is a great project and was easy to integrate!

PR #337, opened 2016-07-15
Added patterns to @GlobalChannelInterceptor

steve-oakey:
Makes use of property place holders to set the `patterns` attribute of the `@GlobalChannelInterceptor` annotation used for the `TraceChannelInterceptor`. As supported by the annotation, only simple
patterns are supported.

Added the Spring configuration metadata for `spring.sleuth.integration` properties.

Fixes gh-323


dsyer:
LGTM

marcingrzejszczak:
@steve-oakey thanks a lot for your contribution! :)

Issue #338, opened 2016-07-15
SleuthStreamReporter fails when multiple Pollers defined but no default Poller


steve-oakey:
When integrating Sleuth into existing applications, we had one Spring Integration application that had multiple `PollerMetadata` beans registered. We did not have a bean registered with a name of `PollerMetadata.DEFAULT_POLLER`. Since the `@InboundChannelAdapter` in `SleuthStreamReporter` relies on a default poller, the application would not start.

Would it make sense to provide a Poller (using `@Poller` within `@InboundChannelAdapter`)?

garyrussell:
Yes, I agree - it's not really a good idea to rely on a default poller, in a component provided by a library - the default poller in the application might have completely inappropriate properties and, as @steve-oakey says, there might not be one present at all.

It would be better to provide an explicit `@Poller` in the component definition, perhaps with configurable properties (and some appropriate opinionated defaults).

marcingrzejszczak:
Thanks @garyrussell ! @steve-oakey - care for another PR? :)

steve-oakey:
Sure. Any preference on the poller configuration? Would providing a configurable fixed rate with 1 second as a default be sufficient?

marcingrzejszczak:
Sure, sounds good

mbogoevici:
Agree with @garyrussell. As background info Spring Cloud Stream provides some default poller metadata, but it's only activated when autoconfiguration is on. It's okay for an application, but not for a library.

garyrussell:
I am not familiar with the application but if the time to accumulate the `Spans` can approach a second, you might want to use `fixed-delay` instead of `fixed-rate`.

Also, this doesn't look very efficient to me...

		List<Span> result = new ArrayList<>();
		this.queue.drainTo(result);
		for (Iterator<Span> iterator = result.iterator(); iterator.hasNext();) {
			Span span = iterator.next();
			if (span.getName() != null && span.getName().equals("message/" + SleuthSource.OUTPUT)) {
				iterator.remove();
			}
		}

If result is large, those remove operations might be expensive because the array has to be shifted down each time.

It would be much better to use a `LinkedList`.

marcingrzejszczak:
Oh yeah - a very good point! Thanks @garyrussell

PR #343, opened 2016-07-16
Added PollerMetadata for StreamSpanReporter

steve-oakey:
Added configurable PollerMetadata bean to be used with the Inbound
Channel Adapter in the StreamSpanReporter. Fixed delay and max messages
per poll are configurable via configuration properties. Implementors can
provide a bean of type PollerMetadata and name
StreamSpanReporter.POLLER to take full control of the poller.

Fixes: gh-338

### DIFF
--- a/spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/SleuthStreamProperties.java
+++ b/spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/SleuthStreamProperties.java
@@ -28,6 +28,7 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 public class SleuthStreamProperties {
 	private boolean enabled = true;
 	private String group = SleuthSink.INPUT;
+	private Poller poller = new Poller();
 
 	public boolean isEnabled() {
 		return this.enabled;
@@ -43,5 +44,37 @@ public class SleuthStreamProperties {
 
 	public void setGroup(String group) {
 		this.group = group;
+	}
+
+	public Poller getPoller() {
+		return this.poller;
+	}
+
+	public static class Poller {
+		/**
+		 * Fixed delay (ms). Default: 1000
+		 */
+		private long fixedDelay = 1000L;
+
+		/**
+		 * Max messages per poll. Default: -1 (unbounded)
+		 */
+		private int maxMessagesPerPoll = -1;
+
+		public long getFixedDelay() {
+			return this.fixedDelay;
+		}
+
+		public int getMaxMessagesPerPoll() {
+			return this.maxMessagesPerPoll;
+		}
+
+		public void setFixedDelay(long fixedDelay) {
+			this.fixedDelay = fixedDelay;
+		}
+
+		public void setMaxMessagesPerPoll(int maxMessagesPerPoll) {
+			this.maxMessagesPerPoll = maxMessagesPerPoll;
+		}
 	}
 }
--- a/spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/StreamSpanReporter.java
+++ b/spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/StreamSpanReporter.java
@@ -27,7 +27,7 @@ import org.springframework.cloud.sleuth.SpanReporter;
 import org.springframework.cloud.sleuth.metric.SpanMetricReporter;
 import org.springframework.integration.annotation.InboundChannelAdapter;
 import org.springframework.integration.annotation.MessageEndpoint;
-
+import org.springframework.integration.annotation.Poller;
 
 /**
  * A message source for spans. Also handles RPC flavoured annotations.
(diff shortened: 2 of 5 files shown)


spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/SleuthStreamProperties.java
```diff
@@ -44,4 +45,40 @@ public void setEnabled(boolean enabled) {
...
+
+	public void setPoller(Poller poller) {
+		this.poller = poller;
+	}
+
+	public static class Poller {
+		/**
+		 * Fixed delay (ms). Default: 1000
+		 */
+		private long fixedDelay = 1000L;
```
adriancole: maybe add javadoc link to the thing this actually affects? eg PeriodicTrigger

spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/SleuthStreamProperties.java
```diff
@@ -44,4 +45,40 @@ public void setEnabled(boolean enabled) {
...
+
+		/**
+		 * Max messages per poll. Default: -1 (unbounded)
+		 */
+		private int maxMessagesPerPoll = -1;
```
adriancole: same comment

spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/SleuthStreamProperties.java
```diff
@@ -44,4 +45,40 @@ public void setEnabled(boolean enabled) {
...
+
+	public Poller getPoller() {
+		return this.poller;
+	}
+
+	public void setPoller(Poller poller) {
+		this.poller = poller;
+	}
+
+	public static class Poller {
```
adriancole: I'd do a quick scan of other spring projects that use properties to configure PollerMetadata, mainly to see if the naming conventions line up

spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/StreamSpanReporter.java
```diff
@@ -38,6 +38,13 @@
 @MessageEndpoint
 public class StreamSpanReporter implements SpanReporter {
 
+	/**
+	 * Bean name for the
+	 * {@link org.springframework.integration.scheduling.PollerMetadata
+	 * PollerMetadata}
+	 */
+	public static final String POLLER = "streamSpanReporterPoller";
```
adriancole: @dsyer etc.. is this a normal spring thing? I've not been doing that, but can. pro and con here is that the name is decoupled from the class so it will stay the same even if the class is renamed.

adriancole:
looks like it could be useful. w/o unit or micro-integration tests, the only way to know it works is manual testing..

spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/StreamSpanReporter.java
dsyer: I don't see any reason to use a bean name that isn't based on the classname.

spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/SleuthStreamProperties.java
steve-oakey: These comments produce the comments in the Spring Boot Configuration Metadata (for content assist in the IDE). The alternative to this is creating the `additional-spring-configuration-metadata.json` file. I am impartial to either, whichever the project prefers.

dsyer: Javadoc links don't work in `@ConfigurationProperties`, so please don't, actually.

adriancole: >
> Javadoc links don't work in @ConfigurationProperties, so please don't,
> actually.
>
ahh sorry to muddy the water

steve-oakey:
@adriancole, @dsyer - Thanks for the feedback. I can add some tests that simply build different `ApplicationContext`'s with different configurations (i.e. one with an provided `PollerMetatdata` bean, one with configured properties and one with defaults) and assert the poller's next execution date is correct and the max messages field is propagated.

I likely will not have time to do this until tomorrow evening. Let me know if you'd like anything else addressed.

marcingrzejszczak:
Do you need some help @steve-oakey with this?

steve-oakey:
Sorry, I am on vacation and was not aware a conflict was introduced. The last time I checked there wasn't one. I do have my laptop with me so I can take a look this afternoon.

marcingrzejszczak:
hmm there is no conflict. But since you're on vacation you should relax instead of coding :P I guess  the questions were asked about adding tests. Will you have some time to add them after you come back from the holidays?

steve-oakey:
I swear a second ago it said the commit had conflicts (I'm using the mobile app ForkHub) but it does not say that now! 

I did add a test case, `SleuthStreamAutoConfigurationTest` that tests the different poller config options and asserts the next execution date using the poller config. I pushed this a while ago and all checks passed. Is there something else I'm missing?

marcingrzejszczak:
Obviously I'm blind and haven't noticed that. Will try to merge your PR ASAP then!

marcingrzejszczak:
Thanks a lot for this awesome contribution!!

PR #360, opened 2016-07-29
Updating the duration setting

### LINKED ISSUE #349: Consider changing the duration when CR, CS logs are set
Currently we can have a diff in duration. Imagine the case

```
Span span = tracer.createSpan("foo"); // span is started t=0
// stuff going on
span.logEvent("CS"); // t=100ms
// stuff going on and response received
span.logEvent("CR"); // t=200ms
// stuff going on
tracer.closeSpan(span); // t=300ms
```

In this case currently the duration is 300ms where in fact it should be 100ms

marcingrzejszczak:
fixes #349 

cc @adriancole

### DIFF
--- a/spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/ConvertToZipkinSpanList.java
+++ b/spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/ConvertToZipkinSpanList.java
@@ -91,7 +91,7 @@ final class ConvertToZipkinSpanList {
 		}
 		zipkinSpan.timestamp(span.getBegin() * 1000);
 		if (!span.isRunning()) { // duration is authoritative, only write when the span stopped
-			zipkinSpan.duration(span.getAccumulatedMicros());
+			zipkinSpan.duration(calculateDurationInMicros(span));
 		}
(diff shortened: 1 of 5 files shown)


spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/ConvertToZipkinSpanList.java
```diff
@@ -146,4 +146,21 @@ private static boolean hasClientSend(Span span) {
 		return false;
 	}
 
+	private static long calculateDurationInMicros(Span span) {
```
adriancole: I think I see what's going on here.. the goal being to not confuse by having a span wider than cr - cs. Seems a shame that this means we can't use the precise duration... probably warrants a comment or javadoc explaining this. iirc, it would be something like..

There could be instrumentation delay between span creation and the semantic start of the span (client send). When there's a difference, spans look confusing. Ex users expect duration to be client receive - send, but it is a little more than that. Rather than have to teach each user about the possibility of instrumentation overhead, we truncate absolute duration (span finish - create) to semantic duration (client receive - send)

marcingrzejszczak:
I've also added a note on the `getAccumulatedNanos` on Span.

adriancole:
:shipit:

PR #361, opened 2016-07-31
Updates to Zipkin 1.6 (which is Spring Boot 1.4)

adriancole:
This updates to Zipkin 1.6, which targets Spring Boot 1.4

Zipkin 1.6 also corrects default values around the UI, which should lead
to better search performance.


marcingrzejszczak:
:shipit:

PR #371, opened 2016-08-09
Added Trace and TraceMessage Headers

### LINKED ISSUE #370: Externalize span / trace id headers
Quite a few users have asked about a possibility to externalize the main header names instead of needing to implement custom extractors.

marcingrzejszczak:
with this change you can customize tracing headers by properties

fixes #370

cc @adriancole

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-sleuth.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-sleuth.adoc
@@ -269,6 +269,24 @@ And you could register them like this:
 include::../../../..//spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/TraceFilterCustomExtractorTests.java[tags=configuration,indent=0]
 ----
 
+=== Custom tracing headers
+
+TIP: Feature available since 1.1.0
+
+If in your system you have different header values for your trace id (e.g. `correlationId`) and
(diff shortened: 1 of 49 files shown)


marcingrzejszczak:
After a chat with @adriancole we've decided to take another approach...

In OpenTracing we have a TextMap - https://github.com/opentracing/opentracing-java/blob/master/opentracing-api/src/main/java/io/opentracing/propagation/TextMap.java

```
/** Generically supports http header propagation */
interface HttpSpanInjector extends SpanInjector<TextMap> {
}

/** Default implemetation for zipkin-compatible propagation */
class ZipkinHttpSpanInjector implements HttpSpanInjector {
    @Override
    public void inject(Span span, TextMap carrier) {
     // X-B3-Foo!
    }
}
```

so that's the first part..

then for each call site we need to convert into TextMap ex.

```
 class HttpRequestTextMap implements TextMap {
  final HttpRequest delegate;

-- do things
}
```

and the usage

```
injector.inject(span, new HttpRequestTextMap(realRequest));
```

We could have 2 of those Injectors/Extractors (Http and Messaging). That way if someone wants something custom there's only 1 place to change. Since we are abstracting all carriers as `TextMap`s (TextMap will be essentially just an abstraction over headers) it will be much easier to manipulate.

adriancole:
:thumbsup:

PR #372, opened 2016-08-09
Simplified Feign

### LINKED ISSUE #202: Simplify Feign integration
We've hidden all the impl details of https://github.com/spring-cloud/spring-cloud-sleuth/commit/19249cdc23a655125a58d022525276e1f74aae09 to make it possible to instrument Feign in a nicer way in the future.

CC @adriancole @spencergibb

marcingrzejszczak:
with this change we no longer treat retries as a continuation of a previous span. That way the Feign code simplifies a lot. RequestInterceptor starts a span and the TraceFeignClient will always close it no matter what's happening.

fixes #202

cc @adriancole


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/TraceFeignClient.java
```diff
@@ -36,42 +39,34 @@
...
 	}
 
 	@Override
 	public Response execute(Request request, Request.Options options) throws IOException {
-		Response response;
 		try {
 			addRequestTags(request);
-			response = this.delegate.execute(request, options);
+			Response response = this.delegate.execute(request, options);
+			logCr();
```
marcingrzejszczak: I'm logging CR only after successful response

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/TraceFeignClient.java
```diff
@@ -36,42 +39,34 @@
...
+			return response;
+		} finally {
+			closeSpan();
```
marcingrzejszczak: If there was an success / exception the span will be closed. Only upon successful response will the span get CR logged

adriancole: on IOException|RuntimeException you could add a binary annotation of "error" -> message, as that will turn the thing red in zipkin's UI

marcingrzejszczak: Nice idea

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/TraceFeignClientAutoConfiguration.java
```diff
@@ -104,26 +104,26 @@ TraceFeignObjectWrapper traceFeignObjectWrapper(BeanFactory beanFactory) {
...
+		return new ResponseEntityDecoder(new SpringDecoder(this.messageConverters)) {
+				@Override
+				public Object decode(Response response, Type type)
+						throws IOException, FeignException {
+					Tracer tracer = beanFactory.getBean(Tracer.class);
+					if (tracer == null) {
+						throw new IllegalStateException("Trace hasn't been yet initialized");
 					}
-				});
+					FeignResponseHeadersHolder feignResponseHeadersHolder = new FeignResponseHeadersHolder(
```
adriancole: why are we injecting headers into the response?

adriancole:
looks good!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/TraceFeignClientAutoConfiguration.java
marcingrzejszczak: Actually that was an approach that we agreed on not doing some time ago... Good catch - I removed it

marcingrzejszczak: Ah! It's this old approach where we injected trace stuff into the response. I think it should be deleted. Good catch!

PR #373, opened 2016-08-12
Deploying documentation to proper folder

marcingrzejszczak:
What we're missing ATM is different documentation versions for different application versions. WIth changing the gh-script what I've managed to achieve is:

- presenting the version of the application in the docs
- checking out the Git branch (e.g. 1.0.x)
- checking out the application version from pom  (e.g. 1.0.5.BUILD-SNAPSHOT)
- checking out what is the name of the main adoc file (e.g. spring-cloud-sleuth)
- pulling the changes from gh-pages after checkout (we're missing that everywhere IMO...)
- in gh-pages creating a folder with name of the version (e.g. /1.0.5.BUILD-SNAPSHOT)
- copying all the `docs/target/generated-docs/` to that folder
  - if the branch from which we're calling the script is NOT master then I'm changing the `spring-cloud-sleuth.html` to `index.html` so that it's easier to access the docs (e.g. http://cloud.spring.io/spring-cloud-sleuth/1.0.5.BUILD-SNAPSHOT/)
  - if it's master then I'm both copying the target docs to proper version subfolder, but also I'm doing things as we did until now. E.g. if master is 1.1.0.BUILD-SNAPSHOT then (e.g. http://cloud.spring.io/spring-cloud-sleuth/1.1.0.BUILD-SNAPSHOT/ and http://cloud.spring.io/spring-cloud-sleuth/spring-cloud-sleuth.html will get updated)

The good thing is that we'll be able to do the same for all releases and milestones. It will be enough to update the contents of this script on a branch / milestone and we're ready to go... Also it will be much easier to copy the release train documentation since the docs will be in their respective version folders.

Example of execution: https://github.com/marcingrzejszczak/spring-cloud-sleuth/tree/gh-pages
Example of a deployed docs: http://toomuchcoding.com/spring-cloud-sleuth/
Example of a deployed docs from master: http://toomuchcoding.com/spring-cloud-sleuth/spring-cloud-sleuth.html
Example of a deployed docs from a branch: http://toomuchcoding.com/spring-cloud-sleuth/1.1.0.BUILD-SNAPSHOT/

cc @dsyer @ryanjbaxter @spencergibb @adriancole

### DIFF
--- a/docs/src/main/asciidoc/ghpages.sh
+++ b/docs/src/main/asciidoc/ghpages.sh
@@ -25,6 +25,9 @@ fi
 # Retrieve properties
 ###################################################################
 
+# Prop that will let commit the changes
+COMMIT_CHANGES="no"
+
 # Get the name of the `docs.main` property
 MAIN_ADOC_VALUE=$(mvn -q \
     -Dexec.executable="echo" \
@@ -37,8 +40,9 @@ echo "Extracted 'main.adoc' from Maven build [${MAIN_ADOC_VALUE}]"
 WHITELISTED_BRANCHES_VALUE=$(mvn -q \
     -Dexec.executable="echo" \
     -Dexec.args='${docs.whitelisted.branches}' \
-    --non-recursive \
-    org.codehaus.mojo:exec-maven-plugin:1.3.1:exec)
+    org.codehaus.mojo:exec-maven-plugin:1.3.1:exec \
+    -P docs \
+    -pl docs)
 echo "Extracted 'docs.whitelisted.branches' from Maven build [${WHITELISTED_BRANCHES_VALUE}]"
 
 # Code getting the name of the current branch. For master we want to publish as we did until now
@@ -72,6 +76,7 @@ if [[ "${CURRENT_BRANCH}" == "master" ]] ; then
             git add -A ${ROOT_FOLDER}/$file
         fi
     done
+    COMMIT_CHANGES="yes"
 else
     echo -e "Current branch is [${CURRENT_BRANCH}]"
     # http://stackoverflow.com/questions/29300806/a-bash-script-to-check-if-a-string-is-present-in-a-comma-separated-list-of-strin
@@ -93,18 +98,21 @@ else
                 fi
             fi
         done
+        COMMIT_CHANGES="yes"
     else
         echo -e "Branch [${CURRENT_BRANCH}] is no ton the whitelist! Won't do anything about this..."
     fi
 fi
 
-git commit -a -m "Sync docs from ${CURRENT_BRANCH} to gh-pages"
+if [[ "${COMMIT_CHANGES}" == "yes" ]] ; then
+    git commit -a -m "Sync docs from ${CURRENT_BRANCH} to gh-pages"
 
-# Uncomment the following push if you want to auto push to
-# the gh-pages branch whenever you commit to master locally.
-# This is a little extreme. Use with care!
-###################################################################
-git push origin gh-pages
+    # Uncomment the following push if you want to auto push to
+    # the gh-pages branch whenever you commit to master locally.
+    # This is a little extreme. Use with care!
+    ###################################################################
+    git push origin gh-pages
+fi
 
 # Finally, switch back to the master branch and exit block
 git checkout ${CURRENT_BRANCH}
--- a/docs/src/main/asciidoc/ghpages.sh
+++ b/docs/src/main/asciidoc/ghpages.sh
@@ -26,7 +26,7 @@ fi
 ###################################################################
 
 # Get the name of the `docs.main` property
-MAIN_ADOC_VALUE=MVN_VERSION=$(mvn -q \
+MAIN_ADOC_VALUE=$(mvn -q \
     -Dexec.executable="echo" \
     -Dexec.args='${docs.main}' \
     --non-recursive \
--- a/docs/src/main/asciidoc/ghpages.sh
+++ b/docs/src/main/asciidoc/ghpages.sh
@@ -25,9 +25,9 @@ fi
 # Retrieve version number, name of the main adoc and name of the current branch
 ###################################################################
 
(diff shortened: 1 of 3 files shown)


docs/src/main/asciidoc/ghpages.sh
```diff
@@ -12,6 +12,40 @@ if ! [ -d docs/target/generated-docs ]; then
...
+
+# Code grepping for the 1st presence of "version>" in pom.xml.
+# First one is project version, second parent version.
+VERSION_NODE=`awk '/version>/{i++}i==1{print; exit}' $ROOT_FOLDER/pom.xml`
+# Extract the contents of the version node
+VERSION_VALUE=$(sed -ne '/version/{s/.*<version>\(.*\)<\/version>.*/\1/p;q;}' <<< "$VERSION_NODE")
+echo "Extracted version from root pom.xml is [${VERSION_VALUE}]"
+
+# Code grepping for the 2nd presence of "version>" in pom.xml.
+# First one is parent, second project version.
```
spencergibb: This might not be consistent, may have to be updated in each project. Wonder if we can use maven to get the version without have to awk/sed/grep etc...

marcingrzejszczak: argh.. true. We already do pass version in Maven but for adoc generation. The problem is that `gh-pages` is executed AFTER the docs have been created. We could call the script from Maven and provide the parameters. That would be SO MUCH better...

spencergibb: MVN_VERSION=$(mvn -q \
    -Dexec.executable="echo" \
    -Dexec.args='${project.version}' \
    --non-recursive \
    org.codehaus.mojo:exec-maven-plugin:1.3.1:exec)

From http://stackoverflow.com/questions/3545292/how-to-get-maven-project-version-to-the-bash-command-line

spencergibb:
how does it work for a release? Most times I haven't pushed when I set the RELEASE, it's only a commit in between snapshots. I guess we should push that commit independently?

marcingrzejszczak:
Actually to backport that for the release we would have to copy paste the contents of the script and run that manually. For new releases - since we're doing them manually we'd have to just run the script. For CI / CDs it will be ran on each commit.

docs/src/main/asciidoc/ghpages.sh
marcingrzejszczak: That would solve the version - we would need also that for the property - I don't know how to do that with Maven :/ But I didn't check that either

ryanjbaxter:
I am not sure I understand everything the script does, but from what I see it looks OK.  If we have to run this script as part of the release process, I just ask that we update the documentation I put together on that process with the information on how to run the script.

Seems like we will have to apply these changes to every project now right?  If so would it be worth trying to centralize `ghpages.sh` somewhere so we don't have to go through this process again?

docs/src/main/asciidoc/ghpages.sh
dsyer: Why do we need the version anyway? I don't want to accumulate a new directory per version particularly. I prefer to choose which ones are archived and not keep old versions that are obsolete (e.g. use "1.0.x" instead of "1.0.0", "1.0.1", "1.0.2" etc.). Maybe we could just use the branch name, and keep a whitelist of branch names that we want to save the docs for (so that feature branches and pull requests don't pollute the tree).

marcingrzejszczak: Hmm actually the docs can differ between minor versions can't they? In this case for sleuth we would have snapshots docs for 1.0.x and master branches. + we would have the docs for all milestones and release versions in separate folder. I think that's the easiest to maintain. As for the whitelist - it's not necessary. We would control the docs release process from CI (execute the script only for those branches that we know are of any interest).

dsyer: > docs can differ between minor versions can't they?

Yes, but we only support the latest version of the code, so I don't see any value in having the docs in a separate directory for unsupported versions. Tags will identify the old values if we need them for reference. So for sleuth at the moment we need 1.0.x and master (which doesn't need a sub directory).

marcingrzejszczak: Ok so we don't want to version the lib in the way Spring is versioned for instance? Don't we want to be consistent?(e.g. http://docs.spring.io/spring-boot/docs/ ) . They have a folder for each release / snapshot version and that's pretty nice for someone who is using a particular version of the docs, isn't it?

If not then achieving what you say will be quite trivial. We'd have to change the code for master so that we don't publish docs to an additional folder. For the rest we don't use the version at all but choose the branch name as the name of the folder. Our Jekins would call the script for 1.0.x and master.

dsyer: I like trivial. Also, we are storing all this data in a git repo, so I don't think we have the same constraints that Spring Boot does (it's just a static web site). There's no reason to store old snapshot and milestone versions of docs in the git history and in sub-directories IMO.

marcingrzejszczak:
@ryanjbaxter yeah once I fix the script I'll apply the changes to all projects. Theoretically we could put `ghpages.sh` to spring-cloud-build and curl it from there. WDYT @dsyer ?

dsyer:
Looking good to me. I'm not sure I trust that this script will only be run on "whitelisted" branches though. We can solve that problem separately I guess.

marcingrzejszczak:
ok maybe it's better to guard against such situations. I'll add a condition for that. I'd prefer to have that property in maven build maybe? `docs.whitelisted.branches` ?

marcingrzejszczak:
Ok so with the changes the script will:

- search in the `docs` module available under `docs` profile for a property called `docs.whitelisted.branches` - a comma separated list
- if the current branch name matches one in that property then all the things that I've mentioned previously will happen ;)
- otherwise sth like this will pop up `Branch [wip_on_script] is not on the white list! Check out the Maven [docs.whitelisted.branches] property in [docs] module available under [docs] profile. Won't commit any changes to gh-pages for this branch.`

PR #379, opened 2016-08-18
Updates to Zipkin 1.7

adriancole:
Notably includes CORS config and version in the /info endpoint.

https://github.com/openzipkin/zipkin/releases/tag/1.7.0


marcingrzejszczak:
Should we update also - https://github.com/spring-cloud-samples/brewery/blob/master/zipkin-server/build.gradle#L13 ?

marcingrzejszczak:
Do we want to have it also for 1.0.x ?

adriancole:
ya for good measure we should

adriancole:
sure!

adriancole:
well, this is spring-boot 1.4, so as long as 1.0.x is ok with that

marcingrzejszczak:
Ah 1.0.x doesn't like boot 1.4 - so let's skip it

PR #386, opened 2016-08-24
Use StreamListener for coercing messages from JSON `Span`

mbogoevici:
Rely on the `contentType` header of the transported message to
tell Spring Cloud Stream how to coerce the message to `Span`.

### DIFF
--- a/spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/StreamEnvironmentPostProcessor.java
+++ b/spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/StreamEnvironmentPostProcessor.java
@@ -22,7 +22,6 @@ import java.util.Map;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.env.EnvironmentPostProcessor;
 import org.springframework.cloud.sleuth.stream.SleuthSink;
-import org.springframework.cloud.sleuth.stream.Spans;
 import org.springframework.core.env.ConfigurableEnvironment;
 import org.springframework.core.env.MapPropertySource;
 import org.springframework.core.env.MutablePropertySources;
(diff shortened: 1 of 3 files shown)


spencergibb:
@marcingrzejszczak @dsyer

marcingrzejszczak:
:+1:

spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/StreamEnvironmentPostProcessor.java
```diff
@@ -44,8 +43,7 @@ public void postProcessEnvironment(ConfigurableEnvironment environment,
 			SpringApplication application) {
 		Map<String, Object> map = new HashMap<String, Object>();
 		map.put("spring.cloud.stream.bindings." + SleuthSink.INPUT + ".content-type",
-				environment.getProperty("spring.sleuth.stream.content-type",
-						"application/x-java-object;type=" + Spans.class.getName()));
+				"");
```
adriancole: maybe add a comment why we are clearing this?

adriancole:
LGTM with one comment as it would leave me scratching my head once I forget about this PR :)

spring-cloud-sleuth-zipkin-stream/src/main/java/org/springframework/cloud/sleuth/zipkin/stream/StreamEnvironmentPostProcessor.java
mbogoevici: Will do. To be frank, I am not very happy with this approach. I'd rather remove this postprocessor completely, but it is necessary because the other `StreamEnvironmentPostProcessor` that we use for producers is also on the classpath and sets the `content-type` of the input channel to `application/json`. Maybe we could consider using separate channel names for input and output and set configurations separately.

marcingrzejszczak:
Can we merge this @mbogoevici  ?

mbogoevici:
I'll add the requested comments shortly. I'd love to hear some feedback on my comments about the input/output channel overlap.

marcingrzejszczak:
The same name for input / output channels were simplification I guess. The thing is that the change of the name would signify breaking backwards compatibility. The stream server would have to support both names, right?

mbogoevici:
Added a comment as requested by Adrian.

adriancole:
Thx

marcingrzejszczak:
So actually we can merge it now?

PR #391, opened 2016-09-05
Fixes edge-case encoding bug

adriancole:
Sleuth doesn't create I64 binary annotations, but others might. This
version of zipkin fixes an encoding bug when someone logs a binary
annotation of type I64.


marcingrzejszczak:
:shipit:

PR #401, opened 2016-09-13
Project Reactor Instrumentation

marcingrzejszczak:
Added support for Project Reactor.

@smaldini - do we need to instrument all other schedulers as well ? Or the single one is enough? If not can you help me with writing some tests to the other schedulers?

cc @dsyer @adriancole @spencergibb @ryanjbaxter


adriancole:
Heh someone's been busy :)

marcingrzejszczak:
We've pair-programmed with @smaldini for 2 hours and that's the result :)

marcingrzejszczak:
After the review of @smaldini we have come to the following conclusions:

REACTOR:

- TracePublisher should be in core
- TraceFlux and TraceMono would do the same what Publisher does
- TraceFlux and TraceMono should have static builder methods
- TraceScheduler should have a static create
- TracePublisher / TraceFlux / TraceMono - in the static methods should pass a span name
- Wrap all schedulers - all but timing stuff
   - Timed Scheduler - if it didn’t exist don’t do it

RXJAVA:

- remove the old approach
- instrument through the schedulers

DOCS:

REACTOR

- Flow -> TracePublisher / TraceFlux / TraceMono - point of entry
- ExecutionModel -> TraceScheduler - is done automatically to carry on between threads

spring-cloud-sleuth-reactor/src/main/java/org/springframework/cloud/sleuth/instrument/reactive/TraceReactorAutoConfiguration.java
```diff
@@ -0,0 +1,71 @@
...
+	static class ReactorHookRegistrar {
+
+		private static final Log log = LogFactory.getLog(MethodHandles.lookup().lookupClass());
+
+		ReactorHookRegistrar(final Tracer tracer, final TraceKeys traceKeys) {
+			if (log.isDebugEnabled()) {
+				log.debug("Registering the Trace Scheduler Single Factory");
+			}
+			Schedulers.setFactory(new Schedulers.Factory() {
+				@Override public Scheduler newSingle(ThreadFactory threadFactory) {
```
simonbasle: I'm not sure what your intent was here. Do you only want this span manipulation to occur through the `single()` Scheduler? What about a `Flux` that would use the `elastic()` Scheduler, or a `timer()`-based operator?

marcingrzejszczak: Yeah that's the problem. Since my knowledge of Reactor is *limited* to say the least I thought that this covers Fluxes and all the publishers. Obviously it doesn't ;) Is there a way I can hook in into creation of any kind of publisher?

marcingrzejszczak: Yeah I have just read my own comments and I see that I was supposed to implement those :D I absolutely don't remember why we weren't to instrument `TimedScheduler` - @smaldini ?

marcingrzejszczak:
Notes to myself to make the https://github.com/marcingrzejszczak/spring-reactive-university with `sleuth` branch work:

[] - Instead of a `Servlet` `Filter` we need to use the `WebFilter` from spring-web

marcingrzejszczak:
Closing the PR since 1.0.x branch was deprecated

Issue #403, opened 2016-09-13
DNS lookup performance bottleneck when DiscoveryClient returns nonexistent host


s4s0l:
I was experiencing huge performance problems when using spring-cloud-consul-discovery + feign clients with seluth publising to zipkin. After some investigation I've narrowed it to DNS lookup made by org.springframework.cloud.sleuth.zipkin.DiscoveryClientEndpointLocator. In my case consul was running in docker, and my app was running on host. DiscoveryClient in getLocalServiceInstance is returning host as 'consul'. I believe it is hostname of docker container running consul. Then in DiscoveryClientEndpointLocator there is:

```
private int getIpAddress(ServiceInstance instance) {
        try {
            return InetUtils.getIpAddressAsInt(instance.getHost());
        }
        catch (Exception e) {
            return 0;
        }
    }
```

(see: https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/DiscoveryClientEndpointLocator.java)

This is called for EVERY span. When DNS is cached it's somewhat ok. If DNS lookup is performed, not so ok. If DNS timeouts/nonexistent host it has terrible impact on performance! In my case name 'consul' was not resolved... Another story is why consul-discovery is returning consul agent name as host, but i think it is irrelevant, this can always happen.

marcingrzejszczak:
Thanks a lot for the feedback. I guess caching the result would enough to fix it, wouldn't it?

s4s0l:
Actually it is cached, just figured it out...i think. In FallbackHavingEndpointLocator there is:
```
@Override
	public Endpoint local() {
		this.cachedEndpoint.compareAndSet(null, endpoint());
		return this.cachedEndpoint.get();
	}
```
But if i understand it correctly endpoint() method is called every time anyway?

Answering: I'ts hard for me to talk about it as i just started my adventure with zipkin, but first of all why zipkin needs an Ip address? Would it be possible to just leave hostname? 

If hostname will be cached wouldn't it mess with docker containers being migrated? Or what will happen when ip is changed? I would cache based on unconverted host name + port + serviceId - Until it changes I would leave cachedEndpoint alone....But then what if DNS resolution was only temporary unavailable during first resolution...ok i'm overcomplicating it.

s4s0l:
[...]

PR #404, opened 2016-09-13
Issue 403 - DNS lookups may impact performance.

s4s0l:
When spring.zipkin.localEndpointCachingEnabled is set to true (defaults to false), zipkin.Endpoints will be cached to prevent DNS lookups for converting host name to ip, when it is obtained from DiscoveryClient.

Fixes gh-403


spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/EndpointCache.java
```diff
@@ -0,0 +1,28 @@
...
+	 * When first called will create new Endpoint with <code>factory</code> and store it
+	 * for later usage along with provided <code>keys</code>. Upon next calls will check
+	 * if <code>keys</code> are equal to previously stored. If they are previous instance
+	 * of Endpoint will be returned, otherwise it will recreate Endpoint using
+	 * <code>factory</code>.
+	 * @param factory
+	 * @param keys
+	 * @return
+	 */
+	Endpoint getEndpoint(EndpointCacheImpl.EndpointFactory factory, Object... keys);
```
spencergibb: Did you mean `EndpointCache.EndpointFactory`?

spring-cloud-sleuth-zipkin/src/test/java/org/springframework/cloud/sleuth/zipkin/DiscoveryClientEndpointLocatorTest.java
```diff
@@ -31,26 +34,46 @@
...
  */
 @RunWith(MockitoJUnitRunner.class)
 public class DiscoveryClientEndpointLocatorTest {
 
 	@Mock DiscoveryClient discoveryClient;
+	@Mock EndpointCache endpointCache;
+
 	@InjectMocks DiscoveryClientEndpointLocator discoveryClientEndpointLocator;
 
+    @Before
```
spencergibb: Formatting, we use tabs. Looks like this method slipped through :-)

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/DiscoveryClientEndpointLocator.java
```diff
@@ -43,10 +46,22 @@ public Endpoint local() {
...
 		}
-		return Endpoint.create(instance.getServiceId(), getIpAddress(instance), instance.getPort());
+		return this.endpointCache.getEndpoint(createEndpointFactory(instance),
+				getHost(instance), instance.getPort(), instance.getServiceId());
 	}
 
-	private int getIpAddress(ServiceInstance instance) {
+	private EndpointCache.EndpointFactory createEndpointFactory(
+			final ServiceInstance instance) {
+		return new EndpointCacheImpl.EndpointFactory() {
```
spencergibb: again `EndpointCache.EndpointFactory`?

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinAutoConfiguration.java
```diff
@@ -123,7 +126,7 @@ public EndpointLocator zipkinEndpointLocator() {
 
 		private DiscoveryClientEndpointLocator discoveryClientEndpointLocator() {
 			if (this.client!=null) {
-				return new DiscoveryClientEndpointLocator(this.client);
+				return new DiscoveryClientEndpointLocator(this.client, new EndpointCacheImpl(this.zipkinProperties));
```
spencergibb: `EndpointCacheImpl` should be a bean that can have other implementations, otherwise, drop the interface.

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/DiscoveryClientEndpointLocator.java
spencergibb: Declared in it's own `protected static` config class with a `@ConditionalOnProperty` then autowired (required=false).

marcingrzejszczak:
Also I think we should do this for branch 1.0.x . I'll take a closer look at this tomorrow

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/EndpointCache.java
s4s0l: jup

spring-cloud-sleuth-zipkin/src/test/java/org/springframework/cloud/sleuth/zipkin/DiscoveryClientEndpointLocatorTest.java
s4s0l: I started using eclipse formatting plugin 4 idea at the end, it seem validation does not include tests:)

marcingrzejszczak:
One more thing cause maybe I don't get sth... Why not just do this?

Pseudocode:

```
public class DiscoveryClientEndpointLocator implements EndpointLocator {

	private DiscoveryClient client;
        private static final Map<ServiceInstance, Integer> CACHE = new ConcurrentHashMap<>();

	public DiscoveryClientEndpointLocator(DiscoveryClient client) {
		this.client = client;
	}

	@Override
	public Endpoint local() {
		ServiceInstance instance = this.client.getLocalServiceInstance();
		if (instance == null) {
			throw new NoServiceInstanceAvailableException();
		}
		return Endpoint.create(instance.getServiceId(), getIpAddress(instance), instance.getPort());
	}

	private int getIpAddress(ServiceInstance instance) {
		try {
                     Integer ip = CACHE.get(instance);
                      if (ip != null) {
                          return ip;
                      } 
			ip = InetUtils.getIpAddressAsInt(instance.getHost());
                        CACHE.put(instance, ip);
                        return ip;
		}
		catch (Exception e) {
			return 0;
		}
	}

	static class NoServiceInstanceAvailableException extends RuntimeException { }
}
```

Will it not be easier, faster and simpler? Or am I missing sth?

s4s0l:
Hmm... you depend on implementation of ServiceInstance (eg EurekaServiceInstance has no equals(), EurekaDiscoveryClient returns an anonymous class.. - so you will end with OOM pretty soon).

As @spencergibb noted It can be simplified by removing the EndpointCache and putting this code into DiscoveryClientEndpointLocator it should be then a few lines, but I wanted to leave the possibility to change that in case someone wants some different caching strategy (like DNS down on first hit). Also as @spencergibb noticed I've forgot to leave that possibility:)

Anyway if it's overdesigned I'll collapse it to sth lesss bloated.

marcingrzejszczak:
Yeah you're right with the OOM. I think it's a little bit overdesigned that's why I suggested sth less complex (even though wrong ;) ). That means that my time is over for today and we'll revisit what you do tomorrow ;)

s4s0l:
Ok. I'll simplify it. Mainly because when I look at what I've done in last commit DiscoveryClientEndpointLocator actually has almost zero responsibility. If someone wants to change caching strategy he can replace the whole thing. Tomorrow in the morning I'll push it again..

Sorry for fuss. My first PR ever. They say if u wanna learn, learn from the best, so I've picked you guys for my gh learning exercise;)

marcingrzejszczak:
We're extremely happy that you're spending your private time on fixing this. It means a lot to us! Thanks a lot and let's make this code awesome :)

spencergibb:
👍  to what @marcingrzejszczak said. We appreciate your efforts.

marcingrzejszczak:
I had one last look before going to sleep. I wonder if we really need all of those factories / beans etc. I'll come back for a sec to what I did. Since I was typing from my mobile I didn't have access to the IDE. Looking at the https://github.com/spring-cloud/spring-cloud-commons/blob/master/spring-cloud-commons/src/main/java/org/springframework/cloud/client/ServiceInstance.java interface, as a key to the ConcurrentHashMap we could use a `serviceInstanceId` cause basically that differntiates one `ServiceInstance` from another doesn't it @spencergibb ? So since the `DiscoveryClientEndpointLocator ` is a bottle neck we could easilly fix that without introducing any new classes:

Pseudocode

```
public class DiscoveryClientEndpointLocator implements EndpointLocator {

    private DiscoveryClient client;
        private static final Map<String, Integer> CACHE = new ConcurrentHashMap<>();

    public DiscoveryClientEndpointLocator(DiscoveryClient client) {
        this.client = client;
    }

    @Override
    public Endpoint local() {
        ServiceInstance instance = this.client.getLocalServiceInstance();
        if (instance == null) {
            throw new NoServiceInstanceAvailableException();
        }
        return Endpoint.create(instance.getServiceId(), getIpAddress(instance), instance.getPort());
    }

    private int getIpAddress(ServiceInstance instance) {
        try {
                     Integer ip = CACHE.get(instance.getServiceId());
                      if (ip != null) {
                          return ip;
                      } 
            ip = InetUtils.getIpAddressAsInt(instance.getHost());
                        CACHE.put(instance.getServiceId(), ip);
                        return ip;
        }
        catch (Exception e) {
            return 0;
        }
    }

    static class NoServiceInstanceAvailableException extends RuntimeException { }
}
```

WDYT @spencergibb @s4s0l  ?

s4s0l:
the key should be a hostname. It can change over time (at least i can imagine that), while serviceId - unlikely. also putting things into cache isn't atomic.

marcingrzejszczak:
But didn't you say that resolving a hostname takes a lot of time abd is a bottleneck? As for atomicity that's true, to make things even simpler we can use synchronized and lock on the cache (regards of whether we use a map or a reference). I don't think that the performance impact would be gigantic (or bigger than the current hostname resolution)

s4s0l:
I've pushed simplified version, while circle is spinning i looked at it again, and I have 2 problems with what I've done:
1) don't know how to test it, because it is impossible now to mock DNS lookup or at least I don't know how - would you mind if I wrap InetUtils.getIpAddressAsInt in some interface?
2) I've looked again at FallbackHavingEndpointLocator - there was this construct with atomicReference - I removed it as it was always returning same thing and in the same time was always calling locators. As an effect there is no Endpoint instance caching anywhere. Which may cause some memory problems. Am I right? To fix it it would be better for DiscoveryClientEndpointLocator to cache not only the IP but whole Endpoint object. Then the same should be done for ServerPropertiesEndpointLocator.... Am i right?

s4s0l:
@marcingrzejszczak dns lookup is a bottleneck when done against invalid hostname, but doing DNS lookup once every time DiscoveryClient returns different host name is OK i think.

s4s0l:
about synchronization, this is really hot code I'd avoid synchronization. Previous implementation was not so slow, because of OS/JVM level DNS caching. So when everything was set up correctly (hostname was valid) it was fast. Problem arise only when DNS cache expires. And it hurts plenty when there is some DNS hiccup. But between theese hiccups this code was called.

Another look at this problem. Even if all this works and ok, resolving DNS during reporting is not too good anyway. Schouldn't resolving be delayed till flushing to zipkin? It is done this way in seluth-stream. Even when we create Endpoint only the first time it can have huge consequences. When DNS times out it will probably time out some circut breaker, it gets closed and we loose a lot of requests (actually it's my case)....maybe it would be better to leave the locator alone and just delay it till reporter pushes spans to zipkin, then it would not hurt too much to even do these lookups every time?

s4s0l:
There is also DiscoveryClientHostLocator in seluth-stream and it is also affected by DNS problems. Lesson learned: whed doing PR, before i do it I should go to sleep.

s4s0l:
I think it would be best to do the following:
1) no endpoint caching/DNS caching at all anywhere - the only change would be to remove caching from FallbackHavingEndpointLocator
2) Merge ZipkinSpanListener with HttpZipkinSpanReporter so that pending queue will contain only seluth spans, and when they are pushed conversion to zipkin.span will occur, and all zipkin.span will get the same instance of Endpoint so the call to locate() will occur only once per push, what will minimize DNS lookups. And will make Endpoint information consistent in all spans in one push. This will make ZipkinSpanReporter obsolete. Also putting conversion in some separate bean is a good idea as HttpZipkinSpanReporter (HttpZipkinSpanListener?) will become huge.
3) leave a warning in logs that ip resolution failed, so that users will be aware of it. 

This time I'm first asking then coding. Keep in mind ZipkinSpanReporter is a bean so above changes may break users code if they replaced it. It supposed to be 30min task but now I'm madly determined to have my first PR request accepted:)

marcingrzejszczak:
OK I read the code, I read your comments and had a chat with @dsyer . These are the conclusions:

- I shouldn't reply to issues when I'm half-asleep
- we have to move the conversion to a background thread (kudos @s4s0l for pointing it out)
- this part is slow - `InetUtils.getIpAddressAsInt(instance.getHost());` - 

Now how to achieve this...

- In `FallbackHavingEndpointLocator` the method `this.cachedEndpoint.compareAndSet(null, endpoint());` is wrong. You're right about this that the `endpoint()` will also be called. There should be a check if the endpoint has an entry.
- `ZipkinSpanListener` needs to have its `report` method executed in a separate thread
-  Notice how it looks for `StreamSpanReporter`... we're doing a different conversion (from `Span` -> `Spans`)  together with endpoint location in a separate thread. We have to reach the same in case of `ZipkinSpanListener`
- The caching of `InetUtils` inside the `DiscoveryClientEndpointLocator` - it's enough for us to cache this by means of `instance.getHost()` as a key.  
- `no endpoint caching/DNS caching at all anywhere - the only change would be to remove caching from FallbackHavingEndpointLocator` - why? Let's fix the caching in the fallback thing and let's add it in the discovery one.
- I think that `Merge ZipkinSpanListener with HttpZipkinSpanReporter` should be done in another way - let's run `ZipkinSpanListener`'s code from the `report` method in a separate thread.
- `leave a warning in logs that ip resolution failed, so that users will be aware of it.` - makes sense

CC @adriancole and @dsyer

adriancole:
On the separate thread thing, zipkin-reporter-java's AsycReporter was made
for this. I can raise a PR for that.

marcingrzejszczak:
That would be fantastic! :)

s4s0l:
Great i like it. 

I believe caching endpoints is different topic than asynch reporter so If you let me I'll change according to what was said and leave asynchronous reporting for @adriancole PR. (btw zipkin-reporter-java's AsycReporter won't fit as it pushes List<byte[]> to the sender, but i like it anyway).

But 3 concerns:
* About caching in FallbackHavingEndpointLocator - this class IMHO is unable to decide if endpoint needs caching, it does not know if Endpoint is changed or not, I believe that if DiscoveryClientEndpointLocator is caching then it should be enough. But then ServerPropertiesEndpointLocator would need to cache Endpoint instance as well, because when it is used it would end up in up to 4 instances of Endpoint per zipkin.Span, quite unnecessarily.
* during convert local() is being called multiple times per span, if hostname gets changed during that it would create span with inconsistent addressing information, I'd call it once
* If we agree caching host address is a good idea it should also be a good idea for seluth-stream's DiscoveryClientHostLocator, where other concerns do not apply, so the change is local to this one class.

marcingrzejszczak:
> About caching in FallbackHavingEndpointLocator - this class IMHO is unable to decide if endpoint needs caching, it does not know if Endpoint is changed or not, I believe that if DiscoveryClientEndpointLocator is caching then it should be enough. But then ServerPropertiesEndpointLocator would need to cache Endpoint instance as well, because when it is used it would end up in up to 4 instances of Endpoint per zipkin.Span, quite unnecessarily.

When you take a look at `ServerPropertiesEndpointLocator` it's calling `return InetUtils.getIpAddressAsInt(this.serverProperties.getAddress().getHostAddress());` which in turn is calling `inetAddress = InetAddress.getByName(host);`. <-- this is the slow bit. So each time if discovery client fails to find the `Endpoint` it will call this. I guess we want to get rid of this problem, don't we?

> during convert local() is being called multiple times per span, if hostname gets changed during that it would create span with inconsistent addressing information, I'd call it once

Very good idea!

> If we agree caching host address is a good idea it should also be a good idea for seluth-stream's DiscoveryClientHostLocator, where other concerns do not apply, so the change is local to this one class.

Agreed. Here we also have this `InetAddress address = InetAddress.getByName(instance.getHost());` which is slow.

s4s0l:
regarding ServerPropertiesEndpointLocator we misunderstood somehow, anyway my proposal is to cache Endpoint here no matter what is the reason.

take a peek 
https://github.com/s4s0l/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ServerPropertiesEndpointLocator.java

I've used there something copied from my later projects:
https://github.com/s4s0l/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/DclSingleton.java

I'd put it into sleuth-core because pattern for ServerPropertiesHostLocator could be similar. 

BTW there should be a warning not to run runAcceptanceTests.sh if you are using docker on your machine:)

marcingrzejszczak:
@s4s0l thanks a lot for the feedback!

For me the DCL is hardcore :D What you did is what I proposed - just synchronize the ConcurrentHashMap or as we spoke the reference. I'd inline it in the class. As you see we already have some duplications between the two modules.

> BTW there should be a warning not to run runAcceptanceTests.sh if you are using docker on your machine:)

OOOOPS... I think it might have brute forced killed all your docker containers :D

marcingrzejszczak:
oh and one more thing

```
if (result == null) {
			synchronized (this) {
				result = this.value;
				if (result == null) {
					this.value = result = factory.newInstance();
				}
			}
		}
```

This is wrong I think. Synchronized should be at the beginning, or even the whole method should be synchronized cause `result` could have been changed before reaching the synchronization block, right?

s4s0l:
Nope. It's OK. :) that's the whole point of dcl not having to visit synchronized. Notice the volatile keyword. But you are right. If someone invalidates value after result is assigned it will return 'old' value. But does it matter that we will get old port address in this 1Pico second time? Next call to get the ip will be fine. It would happen anyway even with synchronize on whole method.

 The whole point of this dcl singleton is to create instance only once and not synchronize when it's not necessary. Invalidation is something I've added to support dynamic port changes. It's an overkill for this use case as usually in my code, but as I had it ready why not use it.

adriancole:
here's my 2p

In every other tracer I have seen, most don't worry about the "local" ip address changing while the process is up. Finagle only starts a timing trace "after" it has discovered the ip address of the local host, but it isn't that expensive as they don't attempt DNS (iirc).

Incidentally, zipkin has no way to search by ip anyway. In polling, actually no-one has been interested even enough to raise an issue to support search by ip. The IP address matters when doing clock-skew correction. If it is inconsistently reported, that won't work. Otherwise, it is just more metadata.

In most cases, you don't need DNS to understand your current IP address, so this is not expensive. I'm not sure why we need discovery client to lookup a hostname to determine what our local IP address is. If we have to do a naming query, we might get it wrong (ex if there are multiple answers!) Let's assume there's a good reason to do a DNS lookup to figure the IP address we are listening on..

The most straightforward way to solve the problem is via a Lazy. For those who want to make sure they don't incur a performance hit, they eagerly reference that lazy. That solves the problem as it is defined (a performance problem), without widening it to a new scope (tracking dynamic ips)

marcingrzejszczak:
After having a longer discussion with @adriancole we came to the following conclusions.

The whole complexity introduced here is not worth the problem that we're solving. What is that problem? Resolving IP and putting it to the Endpoint. But actually as @adriancole says Finagle doesn't do that and neither do other traces. We could use CA / SA annotations to set addresses and Zipkin will take care of the rest. 

The problem is that the in a couple of places we're using `InetUtils` that underneath are doing this call `InetAddress.getByName(host);`. Which is slow and might be inaccurate. We have to replace those calls with NetworkInterface iteration. The current resolution via `DiscoveryClientEndpointLocator` is slow, inaccurate and doesn't give us a lot of benefits - until we update the `DiscoveryClient`'s API we should stop using it.

Summing it up:
- we remove `DiscoveryClientEndpointLocator` as a default registered bean
- we allow people to register that bean as default if they want to
- the current `ServerPropertiesEndpointLocator` instead of trying to resolve an ip from host (which under the hood is slow) will go through network interfaces and take out the IP
- we file a feature request to `DiscoveryClient` to provide a method (sth like `InetAddress getHostAddress(boolean refresh);`) - Sleuth shouldn't be solving these issues - it should profit from the `DiscoveryClient`'s konwledge of the IP (for Eureka we already have this info in the metadata)

Anything to add? @adriancole @dsyer @spencergibb @ryanjbaxter ?
WDYT @s4s0l ? (BTW as you can see we're very agile ;) )

spencergibb:
@marcingrzejszczak 👍

s4s0l:
When what I do is overdesigning what we do seems to be overthinking:)

In my opinion making the reporting asynchronous is solving this performance problem in first place, after that whatever is done is just a small improvement, not worth this conversation:)  

Currently when someone is not using discovery, and is not using any container - i think sleuth is pushing 127.0.0.1, so its another reason for +1 for @marcingrzejszczak.

marcingrzejszczak:
Sweet - we've all agreed :) @s4s0l care to continue with the PR? :)

s4s0l:
yes, of corse

s4s0l:
So i got rid of INetUtils. Is @adriancole still wishing to make SpanListener asynchronous or may I?

marcingrzejszczak:
It's already Async - https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/HttpZipkinSpanReporter.java#L32

marcingrzejszczak:
Ah you're talking about different thing (too much on my head). Yeah go ahead he was making the reporter async

adriancole:
I wasn't planning to make the SpanListener async. My thoughts were that the
value of the only expensive thing (the endpoint) could be published
asynchronously or lazily, leaving SpanListener alone sync.

Those concerned about performance would eagerly reference the value of that
endpoint during bootstrap.

>
Let's be careful about making threads that call threads that call threads
without measuring anything.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/util/LocalAdressResolver.java
```diff
@@ -0,0 +1,129 @@
...
+import java.util.Enumeration;
+
+import org.apache.commons.logging.Log;
+
+/**
+ * Utility class for getting local Ip address from network interfaces.
+ * Picks up first non loopback and up address.
+ * @author Matcin Wielgus
+ */
+public class LocalAdressResolver {
```
adriancole: do we want this to be a public type? In most tracers, the other option besides detect is to supply an ip.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/util/LocalAdressResolver.java
```diff
@@ -0,0 +1,129 @@
...
+
+import java.net.*;
+import java.nio.ByteBuffer;
+import java.util.Enumeration;
+
+import org.apache.commons.logging.Log;
+
+/**
+ * Utility class for getting local Ip address from network interfaces.
+ * Picks up first non loopback and up address.
```
adriancole: this code is similar to a lot of things, but it is substantial enough to warrant a question. Was this inspired by another project's code? if so, note the class it was derived from as it usually helps explain choices.

spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/ServerPropertiesHostLocator.java
```diff
@@ -80,7 +83,7 @@ private String getAddress() {
 			address = this.serverProperties.getAddress().getHostAddress();
```
adriancole: ex here I guess we are getting this from properties, so curious why we need to make a public type for local address resolver (as opposed to package private utility). If it were a static utility method, we wouldn't need to affect and break the public ctor of ServerPropertiesHostLocator

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/util/LocalAdressResolver.java
dsyer: Why aren't we using InetUtils from spring-cloud-commons (that's what it is for, and most of this code is duplicated)?

s4s0l: i was googling like crazy could not find, but was hoping for exactly this reaction:) I'll change it.

spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/ServerPropertiesHostLocator.java
s4s0l: It was because this way its mockable/testable

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/util/LocalAdressResolver.java
s4s0l: its similar cause its hard to write it any other way:)

spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/ServerPropertiesHostLocator.java
```diff
@@ -80,7 +83,7 @@ private String getAddress() {
...
 		}
 		else {
-			address = "127.0.0.1";
+			address = this.localAdressResolver.getLocalIp4AddressAsString();
```
adriancole: how about this instead. if you are looking to test fallback with mocks, make a package protected method here called `getLocalIp4AddressAsString()` or similar.

Since this class isn't final, in your test you can either mock that method or simply extend the class and supply a constant.

ex.
```java
testThing = new ServerPropertiesHostLocator(..){
  String getLocalIp4AddressAsString() {
    return "123.4.5";
  }
};
```

s4s0l:
@adriancole So I got lost somewhere...too many opinions in this thread:) I agree one thread is enough. At the beginning I was just thinking the Listener should be asynch not reporter in order to push as much as possible from application thread. But ok now i get it.

spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/ServerPropertiesHostLocator.java
s4s0l: done

spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/ServerPropertiesHostLocator.java
```diff
@@ -38,6 +42,7 @@
 
 	private final ServerProperties serverProperties; // Nullable
 	private final String appName;
+	private final InetUtils inetUtils = new InetUtils(new InetUtilsProperties());
```
spencergibb: `InetUtils` is a bean and should be injected

spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/ServerPropertiesHostLocator.java
```diff
@@ -38,12 +42,14 @@
 
 	private final ServerProperties serverProperties; // Nullable
 	private final String appName;
+	private final InetUtils inetUtils;
 	private Integer port; // Lazy assigned
 
-	public ServerPropertiesHostLocator(ServerProperties serverProperties,
-			String appName) {
+	public ServerPropertiesHostLocator(ServerProperties serverProperties, String appName,
```
adriancole: I'd delegate the old constructor to this one, so it isn't an api break. Ex. by newing up InetUtils when the old constructor is used.

spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/ServerPropertiesHostLocator.java
```diff
@@ -18,17 +18,21 @@
...
 import org.springframework.context.event.EventListener;
 import org.springframework.util.Assert;
 
+import java.net.InetAddress;
+
 /**
  * A {@link HostLocator} that retrieves:
  *
  * <ul>
- *     <li><b>service name</b> - either from {@link Span#getProcessId()} or current application name</li>
```
adriancole: revert formatting

spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/ServerPropertiesHostLocator.java
```diff
@@ -18,17 +18,21 @@
 
 import org.springframework.boot.autoconfigure.web.ServerProperties;
 import org.springframework.boot.context.embedded.EmbeddedServletContainerInitializedEvent;
+import org.springframework.cloud.commons.util.InetUtils;
 import org.springframework.cloud.sleuth.Span;
 import org.springframework.context.event.EventListener;
 import org.springframework.util.Assert;
 
+import java.net.InetAddress;
```
adriancole: I don't think this import is used anymore. revert if so

spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/SleuthStreamAutoConfiguration.java
```diff
@@ -44,11 +44,10 @@
 
 /**
  * {@link org.springframework.boot.autoconfigure.EnableAutoConfiguration
- * Auto-configuration} for sending spans over Spring Cloud Stream. This is for
- * the producer (via {@link SleuthSource}). A consumer can enable binding to
- * {@link SleuthSink} and receive the messages coming from the source (they have
- * the same channel name so there is no additional configuration to do by
- * default).
+ * Auto-configuration} for sending spans over Spring Cloud Stream. This is for the
```
adriancole: revert formatting

spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/SleuthStreamAutoConfiguration.java
```diff
@@ -69,7 +68,8 @@ public Sampler defaultTraceSampler(SamplerProperties config) {
 
 	@Bean
 	@GlobalChannelInterceptor(patterns = SleuthSource.OUTPUT, order = Ordered.HIGHEST_PRECEDENCE)
-	public ChannelInterceptor zipkinChannelInterceptor(SpanMetricReporter spanMetricReporter) {
+	public ChannelInterceptor zipkinChannelInterceptor(
```
adriancole: revert formatting

spring-cloud-sleuth-stream/pom.xml
```diff
@@ -59,6 +58,11 @@
 			<artifactId>spring-boot-starter-test</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>org.aspectj</groupId>
```
adriancole: is this dependency used?

spring-cloud-sleuth-stream/src/test/java/org/springframework/cloud/sleuth/stream/ServerPropertiesHostLocatorTests.java
```diff
@@ -16,55 +16,67 @@
...
 import java.util.Collections;
 
 import org.junit.Test;
+import org.mockito.Mockito;
 import org.springframework.boot.autoconfigure.web.ServerProperties;
+import org.springframework.cloud.commons.util.InetUtils;
+import org.springframework.cloud.commons.util.InetUtilsProperties;
 import org.springframework.cloud.sleuth.Span;
 
-import static org.assertj.core.api.Assertions.assertThat;
```
adriancole: revert import re-org

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/FallbackHavingEndpointLocator.java
```diff
@@ -1,36 +1,33 @@
 package org.springframework.cloud.sleuth.zipkin;
 
-import java.util.concurrent.atomic.AtomicReference;
-
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+
```
adriancole: revert import and javadoc reformat

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/FallbackHavingEndpointLocator.java
```diff
@@ -1,36 +1,33 @@
...
 public class FallbackHavingEndpointLocator implements EndpointLocator {
 
-	private final AtomicReference<Endpoint> cachedEndpoint = new AtomicReference<>();
-
 	private static final Log log = LogFactory.getLog(FallbackHavingEndpointLocator.class);
 
 	private final DiscoveryClientEndpointLocator discoveryClientEndpointLocator;
 	private final ServerPropertiesEndpointLocator serverPropertiesEndpointLocator;
 
-	public FallbackHavingEndpointLocator(DiscoveryClientEndpointLocator discoveryClientEndpointLocator,
```
adriancole: revert formatting

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/FallbackHavingEndpointLocator.java
```diff
@@ -39,8 +36,11 @@ private Endpoint endpoint() {
 		}
 		try {
 			return this.discoveryClientEndpointLocator.local();
-		} catch (Exception e) {
```
adriancole: revert formatting

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ServerPropertiesEndpointLocator.java
```diff
@@ -27,8 +30,8 @@
  * {@link EndpointLocator} implementation that:
  *
  * <ul>
- *     <li><b>address</b> - from {@link ServerProperties}</li>
```
adriancole: revert javadoc formatting

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ServerPropertiesEndpointLocator.java
```diff
@@ -73,11 +78,16 @@ private Integer getPort() {
...
+			return ByteBuffer.wrap(this.serverProperties.getAddress().getAddress())
+					.getInt();
 		}
 		else {
-			return 127 << 24 | 1;
+			return ByteBuffer.wrap(getFirstNonLoopbackAddress().getAddress()).getInt();
 		}
 	}
+
+	private InetAddress getFirstNonLoopbackAddress() {
```
adriancole: inline this as it isn't more clear than calling the method findFirstNonLoopbackAddress directly

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinAutoConfiguration.java
```diff
@@ -37,13 +37,13 @@
 import org.springframework.web.client.RestTemplate;
 
 /**
- * {@link org.springframework.boot.autoconfigure.EnableAutoConfiguration Auto-configuration}
```
adriancole: revert javadoc formatting

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinAutoConfiguration.java
```diff
@@ -53,24 +53,26 @@
...
 @Configuration
-@EnableConfigurationProperties({ZipkinProperties.class, SamplerProperties.class})
+@EnableConfigurationProperties({ ZipkinProperties.class, SamplerProperties.class })
 @ConditionalOnProperty(value = "spring.zipkin.enabled", matchIfMissing = true)
 @AutoConfigureBefore(TraceAutoConfiguration.class)
 public class ZipkinAutoConfiguration {
 
 	@Bean
 	@ConditionalOnMissingBean
-	public ZipkinSpanReporter reporter(SpanMetricReporter spanMetricReporter, ZipkinProperties zipkin,
```
adriancole: revert every change to this method.. it is distracting as it is unrelated to this request

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinAutoConfiguration.java
```diff
@@ -53,24 +53,26 @@
...
 		zipkinRestTemplateCustomizer.customize(restTemplate);
-		return new HttpZipkinSpanReporter(restTemplate, zipkin.getBaseUrl(), zipkin.getFlushInterval(),
-				spanMetricReporter);
+		return new HttpZipkinSpanReporter(restTemplate, zipkin.getBaseUrl(),
+				zipkin.getFlushInterval(), spanMetricReporter);
 	}
 
 	@Bean
 	@ConditionalOnMissingBean
-	public ZipkinRestTemplateCustomizer zipkinRestTemplateCustomizer(ZipkinProperties zipkinProperties) {
```
adriancole: revert

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/ZipkinAutoConfiguration.java
```diff
@@ -81,48 +83,54 @@ public Sampler defaultTraceSampler(SamplerProperties config) {
 	}
 
 	@Bean
-	public SpanReporter zipkinSpanListener(ZipkinSpanReporter reporter, EndpointLocator endpointLocator) {
```
adriancole: revert

spring-cloud-sleuth-zipkin/src/test/java/org/springframework/cloud/sleuth/zipkin/HttpZipkinSpanReporterTest.java
```diff
@@ -20,22 +26,18 @@
 import zipkin.junit.HttpFailure;
 import zipkin.junit.ZipkinRule;
 
-import static java.util.Arrays.asList;
```
adriancole: revert re-org of imports as well all of the other formatting changes to this file

spring-cloud-sleuth-stream/pom.xml
s4s0l: In order to use spring-cloud-commons' InetUtils as bean, I added UtilAutoConfiguration to spring context in tests. For some reason aspectj is needed then.

s4s0l:
rebasing..

s4s0l:
rebased i think.

marcingrzejszczak:
I see one big problem - you're doing this against master? We could think about this in terms of 1.0.x branch too. Could you either file a PR towards 1.0.x or at least squash all of your commits into one?

s4s0l:
I'll file a PR against 1.0.x

s4s0l:
Filed again: https://github.com/spring-cloud/spring-cloud-sleuth/pull/490

PR #407, opened 2016-09-15
Reroutes HttpZipkinSpanReporter to use zipkin.reporter.AsyncReporter

adriancole:
AsyncReporter is a more robust version of what we were doing before.
Notably, it can give a memory threshold instead of span count for the
backlog. This change ports to use AsyncReporter internally.

See https://github.com/openzipkin/zipkin-reporter-java#asyncreporter

### DIFF
--- a/spring-cloud-sleuth-dependencies/pom.xml
+++ b/spring-cloud-sleuth-dependencies/pom.xml
@@ -16,7 +16,8 @@
 	<properties>
 		<spring-cloud-netflix.version>1.2.0.BUILD-SNAPSHOT</spring-cloud-netflix.version>
 		<aspectj.version>1.8.4</aspectj.version>
-		<zipkin.version>1.8.4</zipkin.version>
+		<zipkin.version>1.11.1</zipkin.version>
+		<zipkin-reporter.version>0.4.3</zipkin-reporter.version>
 	</properties>
 	<dependencyManagement>
 		<dependencies>
@@ -91,6 +92,11 @@
 				<groupId>io.zipkin.java</groupId>
 				<artifactId>zipkin-junit</artifactId>
 				<version>${zipkin.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>io.zipkin.reporter</groupId>
+				<artifactId>zipkin-reporter</artifactId>
+				<version>${zipkin-reporter.version}</version>
 			</dependency>
 		</dependencies>
 	</dependencyManagement>
--- a/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/RestTemplateSender.java
+++ b/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/RestTemplateSender.java
@@ -1,0 +1,75 @@
+package org.springframework.cloud.sleuth.zipkin;
+
+import java.net.URI;
+import java.util.List;
+
+import org.springframework.http.HttpHeaders;
+import org.springframework.http.HttpMethod;
+import org.springframework.http.MediaType;
+import org.springframework.http.RequestEntity;
+import org.springframework.web.client.RestTemplate;
+
+import zipkin.reporter.BytesMessageEncoder;
+import zipkin.reporter.Callback;
+import zipkin.reporter.Encoding;
+import zipkin.reporter.Sender;
+
+final class RestTemplateSender implements Sender {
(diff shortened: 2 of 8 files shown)


spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/RestTemplateSender.java
```diff
@@ -0,0 +1,74 @@
...
+		this.restTemplate = restTemplate;
+		this.url = baseUrl + (baseUrl.endsWith("/") ? "" : "/") + "api/v1/spans";
+	}
+
+	@Override public Encoding encoding() {
+		return Encoding.JSON;
+	}
+
+	@Override public int messageMaxBytes() {
+		return 5 * 1024 * 1024;
```
marcingrzejszczak: Any particular reason for that amount of bytes?

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/RestTemplateSender.java
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	@Override public int messageSizeInBytes(List<byte[]> spans) {
+		return encoding().listSizeInBytes(spans);
+	}
+
+	/** close is typically called from a different thread */
+	transient boolean closeCalled;
+
+	@Override public void sendSpans(List<byte[]> encodedSpans, Callback callback) {
```
marcingrzejszczak: I'm blind - where is this called? In the `delegate.flush` ?

spring-cloud-sleuth-dependencies/pom.xml
```diff
@@ -92,6 +93,11 @@
 				<artifactId>zipkin-junit</artifactId>
 				<version>${zipkin.version}</version>
 			</dependency>
+			<dependency>
+				<groupId>io.zipkin.reporter</groupId>
+				<artifactId>zipkin-sender-urlconnection</artifactId>
```
adriancole: oops.. wrong dep. we don't actually use this. will fix

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/RestTemplateSender.java
adriancole: arbitrary, but tends to be the max. in unrealistic profiling, 512KiB is the sweet spot. we could use that!

marcingrzejszczak: would it be a problem to add it as a comment?

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/RestTemplateSender.java
adriancole: AsyncReporter calls this.. you can't see it here. https://github.com/openzipkin/zipkin-reporter-java/blob/master/core/src/main/java/zipkin/reporter/AsyncReporter.java#L222

marcingrzejszczak: ah ok - good to know

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin/RestTemplateSender.java
```diff
@@ -0,0 +1,75 @@
...
+		this.restTemplate = restTemplate;
+		this.url = baseUrl + (baseUrl.endsWith("/") ? "" : "/") + "api/v1/spans";
+	}
+
+	@Override public Encoding encoding() {
+		return Encoding.JSON;
+	}
+
+	@Override public int messageMaxBytes() {
+		// This will drop a span larger than 5MiB. Note: values like 512KiB benchmark better.
```
adriancole: @marcingrzejszczak how's this comment?

marcingrzejszczak: 👍 sweet

PR #411, opened 2016-09-19
Updates to adjust for reporter metrics drift

adriancole:
while we don't send this info, AsyncSpanReporter has backlog metrics now.

aside: we could make an actuate module to expose all of these, if we wanted


adriancole:
picked into 1.0.x too

PR #417, opened 2016-10-06
Updates to latest Zipkin and avoids Endpoint factory method

adriancole:
Now that Endpoint has ipv6, the factory method is even worse than it
was before. This switches to a builder instead.

Note: Endpoint now has a nice toString which should help debugging.


marcingrzejszczak:
Do we want it also in 1.0.x ?

adriancole:
yeah, there's a fix to CORS config in zipkin in the latest version, which
might make it worthwhile putting into 1.0.x

marcingrzejszczak:
Great! Thanks a lot!

adriancole:
cherry-picked into 1.0.x

PR #419, opened 2016-10-10
Ensure message headers are still mutable after interceptor is called

dsyer:
Some components in Spring assume (perhaps wrongly) that a channel
interceptor will not convert a mutable message into an immutable one, and
they continue to modify the headers dowstream. We can dosge the issue of
whether this is right or wrong by keeping the message headers mutable, just
in case.


dsyer:
Wrong base branch (should be 1.0.x)

Issue #237, opened 2016-03-29
Feature Request: Supporting custom headers to be traced along with X-Trace-Id & X-Span-Id


shivangshah:
Currently, tracing X-trace-id and other related headers get traced just fine between applications (provided all are sleuth enabled). It'd be nice to have the ability to add custom headers in application.properties such that they can be traced at the entry and the exit point of a given application. This ability will help under scenarios where a spring sleuth enabled application interacts with legacy applications which manually track some specific headers and the spring sleuth enabled application could track those same headers without really passing them through across all the method calls manually.

marcingrzejszczak:
Related to https://github.com/spring-cloud/spring-cloud-sleuth/issues/143

adriancole:
This is sometimes referred to as baggage. Basically propagating other
values in the same context.

dsyer:
Isn't this already supported via `spring.sleuth.keys.*.headers`?

adriancole:
good question.. @shivangshah are you asking to have custom headers show up in the zipkin ui?

shivangshah:
@adriancole actually no. I haven't actually used zipkin ui yet. What we needed is a way to somehow specify custom headers so that they can be traced through entry/exit points (via Client Interceptors and Servlet Filters) by Sleuth. These headers can be dynamically picked up based on what the specific application configures (probably in their application.properties) to be traced through it's entry/exit points. 
@dsyer I am not sure I know how this supported (or currently works). Is there an example around?

shivangshah:
Also wanted to add. It'd be good to trace and see them through UI as well (but not necessary). Essentially, these parameters (X-trace-Id and other custom headers) get logged (per MDC) and then eventually can be tracked in Splunk or ELK for debugging purposes. What we do have right now is legacy applications (not using sleuth) manually tracking some specific headers, and we would like to not do that manually if we are already using sleuth, and can probably piggyback on sleuth to provide this and log them.

dsyer:
I think I need a more detailed example of what the legacy apps are doing and how you want to use the data the provide. I don't really follow the use case yet (the existing support lets you change the names of the `X-Span_id` (etc.) and also log additional headers in the Span (but those don't get propagated since they aren't needed downstream). A legacy app knows nothing about sleuth so how would you use any features that we provide anyway?

shivangshah:
Sure. So let's say we have legacy Application A, sleuth enabled Application B and legacy Application C. 
[...]

PR #439, opened 2016-11-03
Text map introduction

marcingrzejszczak:
part of #237


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/MessagingSpanTextMapInjector.java
```diff
@@ -0,0 +1,14 @@
+package org.springframework.cloud.sleuth.instrument.messaging;
+
+import org.springframework.cloud.sleuth.SpanInjector;
+import org.springframework.cloud.sleuth.SpanTextMap;
+
+/**
+ * Contract for extracting tracing headers from a {@link SpanTextMap}
```
marcingrzejszczak: injecting

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TraceChannelInterceptor.java
```diff
@@ -40,22 +42,27 @@
...
 		super(tracer, traceKeys, spanExtractor, spanInjector);
+		this.objectMapper = objectMapper;
 	}
 
 	@Override
 	public void afterSendCompletion(Message<?> message, MessageChannel channel, boolean sent, Exception ex) {
 		Span spanFromHeader = getSpanFromHeader(message);
-		if (containsServerReceived(spanFromHeader)) {
-			spanFromHeader.logEvent(Span.SERVER_SEND);
+		getTracer().continueSpan(spanFromHeader);
```
marcingrzejszczak: That's because we send Span headers in form a JSON

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/HeaderBasedMessagingInjector.java
```diff
@@ -0,0 +1,123 @@
...
+				addHeader(textMap, TraceMessageHeaders.PARENT_ID_NAME, Span.idToHex(parentId));
+			}
+			addHeader(textMap, TraceMessageHeaders.SPAN_NAME_NAME, span.getName());
+			addHeader(textMap, TraceMessageHeaders.PROCESS_ID_NAME, span.getProcessId());
+			addHeader(textMap, TraceMessageHeaders.SAMPLED_NAME, Span.SPAN_SAMPLED);
+		}
+		else {
+			addHeader(textMap, TraceMessageHeaders.SAMPLED_NAME, Span.SPAN_NOT_SAMPLED);
+		}
+		addHeader(textMap, TraceMessageHeaders.SPAN_HEADER, storeSpanAsJson(span));
```
adriancole: eh? we aren't literally sending the span in headers, right? that's basically the opposite of the intent of the in-band/reporting duo of dapper systems

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/HeaderBasedMessagingInjector.java
```diff
@@ -0,0 +1,123 @@
...
+		if (StringUtils.hasText(value)) {
+			textMap.put(name, value);
+		}
+	}
+
+	private Long getFirst(List<Long> parents) {
+		return parents.isEmpty() ? null : parents.get(0);
+	}
+
+	// TODO: Seems to be faster than iterating with iterator each time
```
adriancole: a single loop would be faster.. I think you are referring to it being more conventional? PS be careful about case sensitivity since you are now converting it to a map. ex you can use case insensitive tree map if you want to do it like this.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TraceChannelInterceptor.java
adriancole: wow not sure how I missed us serializing the span itself in message headers!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/ZipkinHttpSpanInjector.java
```diff
@@ -0,0 +1,53 @@
...
+ * Default implementation of {@link HttpSpanInjector}, compatible with Zipkin propagation.
+ *
+ * @author Marcin Grzejszczak
+ * @since 1.2.0
+ */
+public class ZipkinHttpSpanInjector implements HttpSpanInjector {
+
+	@Override
+	public void inject(Span span, SpanTextMap carrier) {
+		setIdHeader(carrier, Span.TRACE_ID_NAME, span.getTraceId());
```
adriancole: why wouldn't you do a loop or switch here instead of changing to a map? seems similar amount of code or possibly less

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/HttpRequestTextMap.java
```diff
@@ -0,0 +1,61 @@
...
+
+	private final HttpRequest delegate;
+
+	HttpRequestTextMap(HttpRequest delegate) {
+		this.delegate = delegate;
+	}
+
+	@Override
+	public Iterator<Map.Entry<String, String>> iterator() {
+		Map<String, String> map = new HashMap<>();
```
adriancole: Personally, I'd just make a wrapping iterator which "flat maps" the values for a key into key,value pairs

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TraceChannelInterceptor.java
marcingrzejszczak: Yeah... I can give it another thought on how to do it better.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/HeaderBasedMessagingInjector.java
marcingrzejszczak: That's unfortunately true

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/HeaderBasedMessagingInjector.java
marcingrzejszczak: I thought that maybe it will be faster and easier to work with to return a map, so that you can reference the keys via get instead of looping each time and matching the key.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/ZipkinHttpSpanInjector.java
marcingrzejszczak: Where do I change it to a map here?

PR #440, opened 2016-11-04
Added baggage support

### LINKED ISSUE #237: Feature Request: Supporting custom headers to be traced along with X-Trace-Id & X-Span-Id

marcingrzejszczak:
without this change there is no support for context propagation
with this change whenever you pass the `baggage-...` for http or `baggage_` for messaging headers then such a value will be propagated through your system
                                                                         
fixes #237


README.adoc
```diff
@@ -275,6 +279,34 @@ Below you can find an example of a Logback configuration (file named `https://gi
...
+ header is prefixed with `baggage-` and for messaging it starts with `baggage_`.
+
+IMPORTANT: Since you can attach as many baggage elements as you want and it's propagated through the network
+then if it's weighing too much or you have too many elements then you it may decrease system throughput or increase RPC latencies.
+
+Example of setting span:
+
+[source,java]
+----
+Unresolved directive in intro.adoc - include::http://raw.github.com/spring-cloud/spring-cloud-sleuth/master/spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/multiple/MultipleHopsIntegrationTests.java[tags=baggage,indent=0]
```
marcingrzejszczak: This will get fixed once I merge this ;)

README.adoc
```diff
@@ -275,6 +279,34 @@ Below you can find an example of a Logback configuration (file named `https://gi
...
+----
+
+====== Baggage vs. Span Tags
+
+Baggage travels with the trace (i.e. every child span contains the baggage of its parent). Zipkin has no knowledge of
+baggage and will not even receive that information.
+
+Tags are not passed from a parent to a child, however they are sent to Zipkin and will get presented in the
```
adriancole: If you want to be able to lookup a span based on baggage, you should add corresponding entry as a tag in the root span. (not sure if you want to doc this or not)

README.adoc
```diff
@@ -275,6 +279,34 @@ Below you can find an example of a Logback configuration (file named `https://gi
...
+Zipkin UI for every span.
```
adriancole: tags are attached to a specific span, so they aren't presented "for every span", rather the span they are recorded in. that said you can search by tag...

README.adoc
```diff
@@ -290,7 +322,7 @@ the `spring-cloud-starter-sleuth` module to your project.
               <dependency>
                   <groupId>org.springframework.cloud</groupId>
                   <artifactId>spring-cloud-dependencies</artifactId>
-                  <version>Brixton.RELEASE</version>
+                  <version>Dalston.BUILD-SNAPSHOT</version>
```
adriancole: PS is the dalston change here bc you need a dependency, or just to make sure the change doesn't affect brixton?

README.adoc
```diff
@@ -556,6 +573,9 @@ rest template, scheduled actions, message channels, zuul filters, feign client).
 works via Zipkin-compatible request headers. This propagation logic is defined and customized via
 `SpanInjector` and `SpanExtractor` implementations.
 
+* Sleuth gives you the possibility to propagate context (also known as baggage) between processes. That means that if you set on a Span
+a baggage element then it will be sent either via HTTP or messaging to other processes.
```
adriancole: maybe clarify "sent" as "sent downstream" .. maybe not

docs/src/main/asciidoc/features.adoc
```diff
@@ -34,6 +34,9 @@ rest template, scheduled actions, message channels, zuul filters, feign client).
 works via Zipkin-compatible request headers. This propagation logic is defined and customized via
 `SpanInjector` and `SpanExtractor` implementations.
 
+* Sleuth gives you the possibility to propagate context (also known as baggage) between processes. That means that if you set on a Span
+a baggage element then it will be sent either via HTTP or messaging to other processes.
```
adriancole: same coment as above

docs/src/main/asciidoc/intro.adoc
```diff
@@ -164,6 +164,34 @@ include::https://raw.githubusercontent.com/spring-cloud-samples/sleuth-documenta
 NOTE: If you're using a custom `logback-spring.xml` then you have to pass the `spring.application.name` in
 `bootstrap` instead of `application` property file. Otherwise your custom logback file won't read the property properly.
 
+===== Propagating Span Context
+
+The span context is the state that must get propagated to any child Spans across process boundaries.
```
adriancole: here and above, you can clarify like this. "...across process boundaries. For example, the trace and span IDs are a required part of the span context. Baggage is an optional part."

docs/src/main/asciidoc/intro.adoc
```diff
@@ -164,6 +164,34 @@ include::https://raw.githubusercontent.com/spring-cloud-samples/sleuth-documenta
...
+Part of the Span Context is the Baggage.
+
+Baggage is a set of key:value pairs stored in the span context. Baggage travels together with the trace
+and is attached to every span. Spring Cloud Sleuth will understand that a header is baggage related if the HTTP
+ header is prefixed with `baggage-` and for messaging it starts with `baggage_`.
+
+IMPORTANT: Since you can attach as many baggage elements as you want and it's propagated through the network
```
adriancole: There's currently no limitation of the count or size of baggage items. However, keep in mind that too many can decrease system throughput or increase RPC latency. In extreme cases, it could crash the app due to exceeding transport-level message or header capacity.

docs/src/main/asciidoc/intro.adoc
```diff
@@ -164,6 +164,34 @@ include::https://raw.githubusercontent.com/spring-cloud-samples/sleuth-documenta
...
+Example of setting span:
+
+[source,java]
+----
+include::{github-raw}/spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/multiple/MultipleHopsIntegrationTests.java[tags=baggage,indent=0]
+----
+
+====== Baggage vs. Span Tags
+
+Baggage travels with the trace (i.e. every child span contains the baggage of its parent). Zipkin has no knowledge of
```
adriancole: maybe a see link? noticing this is copied thrice..

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/SpanContext.java
```diff
@@ -0,0 +1,28 @@
...
+ * boundaries and (2) any Tracer-implementation-specific fields that are needed to identify or otherwise contextualize
+ * the associated Span instance (e.g., a <trace_id, span_id, sampled> tuple).
+ *
+ * @see Span#setBaggageItem(String, String)
+ * @see Span#getBaggageItem(String)
+ *
+ * @author Marcin Grzejszczak
+ * @since 1.2.0
+ */
+public interface SpanContext {
```
adriancole: noticing this doesn't include the span ids or sampled flag. ex SpanId in brave

README.adoc
marcingrzejszczak: It's just to have the readme up to date with the given branch

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/SpanContext.java
marcingrzejszczak: I took the interface as it was from OpenTracing. I don't think there were more methods there.

PR #441, opened 2016-11-05
Updates to zipkin 1.14.4

adriancole:
* Support for MySQL 5.7 with SQL_MODE=ONLY_FULL_GROUP_BY
* Includes support to store 128-bit trace ID (not yet implemented here)

See https://github.com/openzipkin/b3-propagation/issues/6


marcingrzejszczak:
I guess we'll need that for `1.1.x` and `1.0.x` too

PR #446, opened 2016-11-10
Disable Feign Retry

ryanjbaxter:
This is needed to make the changes in spring-cloud/spring-cloud-netflix#1457 work when Sleuth is on the classpath.  Do not merge until that pull request is merged into Netflix.

Also changes will need to be made in the 1.1.x branch.  But for that branch we can use the `NEVER_RETRY` class from OpenFeign since a more recent version is in that branch.


marcingrzejszczak:
So what do we do about this?

ryanjbaxter:
Still waiting on the Netflix PR.  Hopefully it will be merged this week.

PR #452, opened 2016-11-17
Propagates logs and reports 128-bit trace IDs

adriancole:
This supports 128-bit traces via a new field traceIdHigh, which matches
other zipkin implementations. In encoded form, the trace ID is simply
twice as long (32 hex characters).

With this change in, a 128-bit trace propagated will not be downgraded
to 64-bits when sending downstream, reporting to Zipkin or adding to
the logging context.

This will be followed by a change to support initiating 128-bit traces.

See https://github.com/openzipkin/b3-propagation/issues/6

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
@@ -138,6 +138,7 @@ public class Span implements SpanContext {
 	private final long begin;
 	private long end = 0;
 	private final String name;
+	private final long traceIdHigh;
 	private final long traceId;
 	private List<Long> parents = new ArrayList<>();
 	private final long spanId;
@@ -169,6 +170,7 @@ public class Span implements SpanContext {
 		this.begin = current.getBegin();
 		this.end = current.getEnd();
 		this.name = current.getName();
+		this.traceIdHigh = current.getTraceIdHigh();
 		this.traceId = current.getTraceId();
 		this.parents = current.getParents();
 		this.spanId = current.getSpanId();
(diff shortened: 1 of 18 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
```diff
@@ -183,37 +185,63 @@ public Span(Span current, Span savedSpan) {
...
+	 */
+	@Deprecated
 	public Span(long begin, long end, String name, long traceId, List<Long> parents,
 			long spanId, boolean remote, boolean exportable, String processId) {
 		this(begin, end, name, traceId, parents, spanId, remote, exportable, processId,
 				null);
 	}
 
+	/**
+	 * @deprecated please use {@link SpanBuilder}
```
adriancole: continuing to overload public ctors would become rediculous

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
```diff
@@ -397,7 +425,30 @@ public long getSpanId() {
...
+	 * <p>Ex. to convert the two fields to a 128bit opaque id array, you'd use code like below.
+	 * <pre>{@code
+	 * ByteBuffer traceId128 = ByteBuffer.allocate(16);
+	 * traceId128.putLong(span.getTraceIdHigh());
+	 * traceId128.putLong(span.getTraceId());
+	 * traceBytes = traceId128.array();
+	 * }</pre>
+	 *
+	 * @see #traceIdString()
+	 * @since 1.2.0
```
adriancole: I think it is worthwhile to backport this to 1.0 and 1.1, even though it would be annoying. The reason is that these versions will be around a long time, and truncate trace ids perpetually unless we do..

marcingrzejszczak: Agreed

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/zuul/SpanInjectingRibbonRequestCustomizer.java
```diff
@@ -68,7 +68,7 @@ public void inject(Span span, SpanTextMap carrier) {
 		}
 		carrier.put(Span.SAMPLED_NAME, span.isExportable() ?
 				Span.SPAN_SAMPLED : Span.SPAN_NOT_SAMPLED);
-		carrier.put(Span.TRACE_ID_NAME, Span.idToHex(span.getTraceId()));
+		carrier.put(Span.TRACE_ID_NAME, span.traceIdString());
```
marcingrzejszczak: I like the `traceIdString` method :)

adriancole:
changed since thing to 1.0.11

PR #453, opened 2016-11-18
Propagates logs and reports 128-bit trace IDs (1.1.x)

adriancole:
This supports 128-bit traces via a new field traceIdHigh, which matches
other zipkin implementations. In encoded form, the trace ID is simply
twice as long (32 hex characters).

With this change in, a 128-bit trace propagated will not be downgraded
to 64-bits when sending downstream, reporting to Zipkin or adding to
the logging context.

This will be followed by a change to support initiating 128-bit traces.


adriancole:
backport of #452

adriancole:
picked into 1.0.x

PR #454, opened 2016-11-18
Updates to latest Zipkin 1.16.2

adriancole:
Here are the features relevant to sleuth

* MySQL support of 128-bit group-by
* don't stack overflow on json write bug
* more forgiving of IPv4-mapped addresses in json


marcingrzejszczak:
Should we also update this - https://github.com/spring-cloud-samples/brewery/blob/master/zipkin-server/build.gradle#L13 ?

adriancole:
yeah. there are a few UI fixes since then, and always nicer to have the
versions aligned

PR #455, opened 2016-11-18
Starts 128-bit traces when spring.sleuth.traceId128=true

adriancole:
This adds autoconfiguration to create 128-bit traces when
`spring.sleuth.traceId128=true`.


adriancole:
not sure where to document this, so open to ideas!

PR #456, opened 2016-11-18
Starts 128-bit traces when spring.sleuth.traceId128=true (1.0.x backport)

adriancole:
This adds autoconfiguration to create 128-bit traces when
`spring.sleuth.traceId128=true`.

See #455


adriancole:
raised as pull request only to make sure I didn't break the branch (already cherry-picked into 1.1.x)

Issue #469, opened 2016-12-10
Trace does not close properly when using ViewController


gavlyukovskiy:
I have single index.jsp inside webapp and next configuration:
```
@SpringBootApplication
public class AggregationServiceApplication extends WebMvcConfigurerAdapter {

    public static void main(String[] args) {
        SpringApplication.run(AggregationServiceApplication.class, args);
    }

    @Override
    public void addViewControllers(ViewControllerRegistry registry) {
        registry.addViewController("/index").setViewName("index");
    }
}
```
application.yml
```
spring.mvc.view.prefix: /
spring.mvc.view.suffix: .jsp
```

When I'm trying to access `/index` (`/index.jsp` is ok) sleuth warns:

> Tried to close span but it is not the current span: [Trace: bdb52aad2860fb80, Span: bdb52aad2860fb80, Parent: null, exportable:false].  You may have forgotten to close or detach [Trace: bdb52aad2860fb80, Span: 334110a624d20d76, Parent: bdb52aad2860fb80, exportable:false]

Request flow:
```
Request
    -> TraceFilter
        -> spanFromRequest = null
        -> createSpan
            -> SpanContextHolder.push([Trace: bdb52aad2860fb80, Span: bdb52aad2860fb80, Parent: null, exportable:false])
        -> filterChain.doFilter
            -> DispatcherServlet
                -> TraceHandlerInterceptor.preHandle
                    -> createSpan
                        -> SpanContextHolder.push([Trace: bdb52aad2860fb80, Span: 334110a624d20d76, Parent: bdb52aad2860fb80, exportable:false])
                -> TraceFilter
                    -> spanFromRequest = [Trace: bdb52aad2860fb80, Span: bdb52aad2860fb80, Parent: null, exportable:false]
                    -> continueSpan
                        -> SpanContextHolder.push([Trace: bdb52aad2860fb80, Span: bdb52aad2860fb80, Parent: null, exportable:false])
                    -> createSpan
                        -> return spanFromRequest
                    -> filterChain.doFilter
                    -> detachOrCloseSpans
                        -> SpanContextHolder.close([Trace: bdb52aad2860fb80, Span: bdb52aad2860fb80, Parent: null, exportable:false])
                            -> current = [Trace: bdb52aad2860fb80, Span: bdb52aad2860fb80, Parent: null, exportable:false]
[...]

Issue #427, opened 2016-10-20
Warnings when used with jsp


semeiser:
Hello,

I get the following warnings in the log when using sleuth with jsps, seems to be one line per included jsp, per request:

```
2016-10-20 10:44:28.288  WARN [-,,,] 72523 --- [nio-8080-exec-1] o.s.cloud.sleuth.util.ExceptionUtils     : Tried to close span but it is not the current span: [Trace: 4638aa2540b6c27b, Span: 4638aa2540b6c27b, Parent: null, exportable:false].  You may have forgotten to close or detach [Trace: 4638aa2540b6c27b, Span: a55e2f71e64e73e4, Parent: 4638aa2540b6c27b, exportable:false]
2016-10-20 10:44:28.291  WARN [-,,,] 72523 --- [nio-8080-exec-1] o.s.cloud.sleuth.util.ExceptionUtils     : Tried to close span but it is not the current span: [Trace: 4638aa2540b6c27b, Span: 4638aa2540b6c27b, Parent: null, exportable:false].  You may have forgotten to close or detach [Trace: 4638aa2540b6c27b, Span: a55e2f71e64e73e4, Parent: 4638aa2540b6c27b, exportable:false]
2016-10-20 10:44:28.306  WARN [-,,,] 72523 --- [nio-8080-exec-1] o.s.cloud.sleuth.util.ExceptionUtils     : Tried to close span but it is not the current span: [Trace: 4638aa2540b6c27b, Span: 4638aa2540b6c27b, Parent: null, exportable:false].  You may have forgotten to close or detach [Trace: 4638aa2540b6c27b, Span: a55e2f71e64e73e4, Parent: 4638aa2540b6c27b, exportable:false]
```

A minimal demo producing the above output (to be run in tomcat container) can be found at https://github.com/semeiser/sleuthjspdemo .

Is there something I can configure to mitigate this issue?

Thank you for any help :)

gavlyukovskiy:
Hi, faced same issue https://github.com/spring-cloud/spring-cloud-sleuth/issues/469. 
There are exmplanation of why it happens, but I don't know how it should be fixed and seems like there are no workaround for it.

marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-sleuth/commit/ee8c73df53478b031e7cd080b328a7b2131b391f

PR #474, opened 2016-12-14
Spans are continued in Handler Interceptors

### LINKED ISSUE #471: Continue spans in TraceHandlerInterceptor
Currently we're creating a new span. We can consider not creating a new span for this but add all the tags to the previous one. 

Example:

image

We would have 1 span `http/foo` that would have all the tags of `service2-method-in-controller`. 

PROS:
- we're not flooding apps with spans

CONS:
- we're not that explicit in terms of span names

@dsyer @adriancole

marcingrzejszczak:
without this change an explicit new span is created on the server side. Its name is equal to the method name of the controller. It introduces some nice advantages in terms of readability of trace.

with this change we're continuing a previous span on the server side. We're attaching the tags and logs to that span with information about controller class and controller name. Also events related to start and finish of the controller are there.

fixes #471 #469 #427

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/TraceKeys.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/TraceKeys.java
@@ -459,6 +459,19 @@ public class TraceKeys {
 	public static class Mvc {
 
 		/**
+		 * The timestamp used to provide a concrete moment in time when request processing
+		 * reached the controller.
+		 */
+		private String controllerStart = "mvc.controller.start";
+
+
+		/**
+		 * The timestamp used to provide a concrete moment in time when request processing
+		 * ended on the controller side.
+		 */
+		private String controllerFinish = "mvc.controller.finish";
+
+		/**
 		 * The lower case, hyphen delimited name of the class that processes the request.
 		 * Ex. class named "BookController" will result in "book-controller" tag value.
 		 */
@@ -469,6 +482,22 @@ public class TraceKeys {
 		 * Ex. method named "listOfBooks" will result in "list-of-books" tag value.
 		 */
 		private String controllerMethod = "mvc.controller.method";
+
+		public String getControllerStart() {
+			return this.controllerStart;
+		}
+
+		public void setControllerStart(String controllerStart) {
+			this.controllerStart = controllerStart;
+		}
+
+		public String getControllerFinish() {
+			return this.controllerFinish;
+		}
+
+		public void setControllerFinish(String controllerFinish) {
+			this.controllerFinish = controllerFinish;
+		}
 
 		public String getControllerClass() {
 			return this.controllerClass;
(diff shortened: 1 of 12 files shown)


marcingrzejszczak:
Here is an example of the new trace. BTW @adriancole the controller start and finish - they were to be events not tags right?

image

dsyer:
LGTM

adriancole:
@marcingrzejszczak why does a controller start and finish twice?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/TraceKeys.java
```diff
@@ -459,6 +459,19 @@ public void setMethodNameKey(String methodNameKey) {
 	public static class Mvc {
 
 		/**
+		 * The timestamp used to provide a concrete moment in time when request processing
```
adriancole: looks good but curious why this would happen multiple times. if it is happening multiple times, it may be confusing (at least it is confusing to me..)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHandlerInterceptor.java
```diff
@@ -62,26 +62,19 @@ public TraceHandlerInterceptor(BeanFactory beanFactory) {
...
-		}
-		if (isSpanContinued(request)) {
-			if (log.isDebugEnabled()) {
-				log.debug("Skipping creation of a span since the span is continued");
-			}
-			return true;
-		}
 		String spanName = spanName(handler);
-		Span span = getTracer().createSpan(spanName);
+		boolean reuseSpan = getTracer().isTracing();
```
adriancole: I think we use the terms join or continue? which are we using more often? maybe do that vs a 3rd way (reuse)

marcingrzejszczak:
AFAIR it's going to error processing to another controller (cause a 404 is thrown or sth).

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/TraceKeys.java
marcingrzejszczak: Yeah... the problem is that `HandlerInterceptor` is used for all types of dispatch (e.g. normal, error, etc.). So I can enter the HandlerInterceptor multiple times through request processing. That's actually a very technical detail and it indeed can confuse people :/

adriancole:
>
> AFAIR it's going to error processing to another controller (cause a 404 is
> thrown or sth).
>
so this means the controller tag is the "primary controller" then?

marcingrzejszczak:
Yeah exactly. It's the first controller that started processing the request.

adriancole:
well since it can hop controllers, is it possible to rename the annotation
to the controller name start/stop? that might make it more clear what's
going on.

adriancole:
ps I think there's a lot of detail we are trying to cram into a single
span.. personally, if there are multiple controllers being called, I'd
make them local spans (and not copy any irrelevant tags to them such
as http tags)

PR #482, opened 2016-12-22
Add tip in case of using spring-cloud-sleuth-stream and spring-cloud-netflix-hystrix-stream

making:
https://twitter.com/making/status/811860408415850496

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-sleuth.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-sleuth.adoc
@@ -534,7 +534,7 @@ If you want to skip Span creation for some `@Scheduled` annotated classes you ca
 `spring.sleuth.scheduled.skipPattern` with a regular expression that will match the fully qualified name of the
 `@Scheduled` annotated class.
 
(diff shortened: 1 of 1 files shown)


marcingrzejszczak:
Actually come to think of it maybe that makes sense to be added as a default value? Is there any reason someone would want to include spans from that scheduled class?

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
```diff
@@ -534,6 +534,8 @@ If you want to skip Span creation for some `@Scheduled` annotated classes you ca
 `spring.sleuth.scheduled.skipPattern` with a regular expression that will match the fully qualified name of the
 `@Scheduled` annotated class.
 
+TIP: If you are using `spring-cloud-sleuth-stream` and `spring-cloud-netflix-hystrix-stream` (which comes with https://docs.pivotal.io/spring-cloud-services/client-dependencies.html[`spring-cloud-services-starter-circuit-breaker`]) together, Span will be created for each Hystrix metrics and sent to Zipkin. This may be annoying. You can prevent this by setting `spring.sleuth.scheduled.skipPattern=org.springframework.cloud.netflix.hystrix.stream.HystrixStreamTask`
```
spencergibb: I wouldn't link to SCS, but to OSS

making: Removed link to SCS

Issue #477, opened 2016-12-16
Misspeling in metrics configuration


gavlyukovskiy:
`TraceMetricsAutoConfiguration` enabled by using `spring.sleuth.metrics.enabled=true`, but the `SleuthMetricProperties` annotated by `@ConfigurationProperties("spring.sleuth.metric")`.

Difference in metric/metrics is confusing, I think they should have same prefix.

And also I'd like to move all such properties under `@ConfigurationProperties` classes or put them into `additional-spring-configuration-metadata.json`.

marcingrzejszczak:
Let's split this into 2 issues. One will be for misspelling and the other for `@ConfigurationProperties`. It would be awesome if you could file a PR for the latter :D

PR #485, opened 2016-12-29
Created @ConfigurationProperties classes and updated additional metadata

gavlyukovskiy:
As we discussed in https://github.com/spring-cloud/spring-cloud-sleuth/issues/477 and gitter I created configuration properties for all properties used in spring-sleuth-core. 
I have used additional-spring-configuration-metadata.json for single `enabled` properties.

In case of camel case property name it resolves both `skip-pattern` and `skipPattern`, but I'd rather update documentation to use conventional property names.
Also `spring.sleuth.rxjava.schedulers.ignoredthreads` looks strange without word separator, we can add new one called `ignored-threads` and deprecate old one, what do you say?`

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
@@ -80,6 +80,12 @@ public class TraceFilter extends GenericFilterBean {
 	protected static final String TRACE_ERROR_HANDLED_REQUEST_ATTR = TraceFilter.class.getName()
 			+ ".ERROR_HANDLED";
 
+	/**
+	 * @deprecated please use {@link SleuthWebProperties#DEFAULT_SKIP_PATTERN}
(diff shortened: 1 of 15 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
```diff
@@ -80,9 +80,6 @@
 	protected static final String TRACE_ERROR_HANDLED_REQUEST_ATTR = TraceFilter.class.getName()
 			+ ".ERROR_HANDLED";
 
-	public static final String DEFAULT_SKIP_PATTERN =
```
marcingrzejszczak: I'd leave this field, deprecate it and point it to the `SleuthWebProperties .DEFAULT_SKIP_PATTERN`. If anyone is using this field it will still be applicable

gavlyukovskiy: done

PR #487, opened 2016-12-30
Added property for overriding the service name in Zipkin

### LINKED ISSUE #324: Add a property for the zipkin service name
@brockovercash

> Is there a way to configure the name that is reported without modifying spring.application.name? I want spring.application.name to be environment agnostic but report up name-dev or similar so that I can differentiate within zipkin.

@marcingrzejszczak

> Theoretically you can write your own EndpointLocator and register it with @Primary
> Check ZipkinAutoConfiguration or SleuthStreamAutoConfiguration

@shakuzen

> Might be nice to have a configuration property for that that defaults to spring.application.name

@adriancole

> good idea
> probably a property with zipkin and service in it, to make it obvious

marcingrzejszczak:
without this only either service discovery service id / `spring.application.name` can be chosen as a service name for zipkin
with this change you can pass `spring.zipkin.service.name` property to change override that both for HTTP and Stream collectors

fixes #324

My only small concern is that I'm duplicating the `ZipkinProperties`. On the other hand those two modules (HTTP / Stream collectors) are kind of similar but not really from the very beginning ;) 

cc @adriancole @gavlyukovskiy

### DIFF
--- a/spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/DiscoveryClientHostLocator.java
+++ b/spring-cloud-sleuth-stream/src/main/java/org/springframework/cloud/sleuth/stream/DiscoveryClientHostLocator.java
@@ -22,27 +22,39 @@ import org.springframework.cloud.client.ServiceInstance;
 import org.springframework.cloud.client.discovery.DiscoveryClient;
 import org.springframework.cloud.sleuth.Span;
 import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 
 /**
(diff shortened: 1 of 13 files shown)


marcingrzejszczak:
THat's a very good question @gavlyukovskiy ;) No deadlines on the release yet. I guess we'll discuss that next week. Happy new year BTW :D

gavlyukovskiy:
Happy new year :)

marcingrzejszczak:
Done in https://github.com/spring-cloud/spring-cloud-sleuth/commit/225a2d4cbf40237303ad775e6010e76e7257ad3f

PR #488, opened 2017-01-02
Adding the client/server.instanceid tag

### LINKED ISSUE #369: Add instanceid to the tags
When doing RPC call (if there's CS / SR / SS / CR ) - ensure that the tag with `spring.application.instanceid` is added for both client / server. E.g. `client.instanceid`, `server.instanceid`.

The rationale behind is that without it in CF it's difficult to assess which particular instance has issues.

marcingrzejszczak:
- without this tag it's impossible to discern from which server was the given span originated
- with this change we're adding a tag in which we're passing the contents of the `spring.application.instanceid` value to define the id of the given instance

fixes #369 

cc @dsyer @adriancole

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
@@ -134,6 +134,18 @@ public class Span {
 	 */
 	public static final String SPAN_PEER_SERVICE_TAG_NAME = "peer.service";
 
+	/**
+	 * ID of the instance from which the span was originated. This tag is related to the
+	 * client side communication of an RPC call.
+	 */
+	public static final String SPAN_CLIENT_INSTANCEID = "client.instanceid";
+
+	/**
+	 * ID of the instance from which the span was originated. This tag is related to the
+	 * server side communication of an RPC call.
+	 */
+	public static final String SPAN_SERVER_INSTANCEID = "server.instanceid";
+
 	private final long begin;
 	private long end = 0;
 	private final String name;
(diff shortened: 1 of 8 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
```diff
@@ -134,6 +134,18 @@
 	 */
 	public static final String SPAN_PEER_SERVICE_TAG_NAME = "peer.service";
 
+	/**
+	 * ID of the instance from which the span was originated. This tag is related to the
+	 * client side communication of an RPC call.
+	 */
+	public static final String SPAN_CLIENT_INSTANCEID = "client.instanceid";
```
adriancole: I don't think you need to add this twice, if you think about how it is used.

For example, are you going to search for a client whose instanceid is this? or anything whose instanceid is this? In the normal dual-host span model, the span detail screen would show the servicename and IP address who reported the instanceid (which would also match any other client-related tags). In any future "single-host span" model, this clarifier will become even more redundant. In other words, unless there's a very good reason to qualify this as client/server I'd just leave it as a normal tag like "spring.application.instanceid" or shorter like "spring.instanceid" id there's not likely an instanceid that isn't an application

marcingrzejszczak: I see one issue with this (but maybe actually I'm wrong then there's no issue at all ;) ). If both the client and the server put the same tag with 2 different values, won't it be broken on the Zipkin UI in the RPC span?

adriancole: it will show up twice, once for each value (for each tracer)

marcingrzejszczak: Awesome! So indeed I'll prefer to simplify things 👍

marcingrzejszczak: Non Cloud Foundry execution:

image

Cloud Foundry
image

marcingrzejszczak:
fixed in https://github.com/spring-cloud/spring-cloud-sleuth/commit/5b442e952673d7a9893370528b9cb5e52e13500c

PR #490, opened 2017-01-07
Issue 403 - DNS lookups may impact performance for 1.0.x

s4s0l:
Changes done according to discussion in https://github.com/spring-cloud/spring-cloud-sleuth/pull/404

Due to possible performance problems when DiscoveryClient gives nonexistent DNS name as local address Local endpoint is created from server properties by default.

Local Ip determined from spring.commons InetUtils.
To enable local endpoint resolution via service discovery, DiscoveryClient must be present and 'spring.zipkin.discoveryLocalEndpointLocator' should be set to true.
FallbackHavingEndpointLocator was caching and recreating EndpointLocator in the same time - removed the caching part.

Fixes gh-403

### DIFF
--- a/spring-cloud-sleuth-stream/pom.xml
+++ b/spring-cloud-sleuth-stream/pom.xml
@@ -27,7 +27,6 @@
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-commons</artifactId>
-			<optional>true</optional>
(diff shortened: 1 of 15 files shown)


marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-sleuth/commit/9785bcae15a4b44b114bc9b5ae6ae7fe36f0342d

PR #497, opened 2017-01-17
Add an option to use the X-B3-Flags header to override any sampling d…

### LINKED ISSUE #496: Add an option to use the X-B3-Flags header to override any sampling decision
without this change it's pretty much impossible to enforce sampling for certain traces
with this change setting the `X-B3-Flags` to `1` for HTTP messages / `spanFlags` to `1` for messaging will override any sampling decisions

marcingrzejszczak:
…ecision

without this change it's pretty much impossible to enforce sampling for certain traces
with this change setting the X-B3-Flags to 1 for HTTP messages / spanFlags to 1 for messaging will override any sampling decisions

fixes #496

cc @adriancole

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/HeaderBasedMessagingExtractor.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/HeaderBasedMessagingExtractor.java
@@ -31,8 +31,13 @@ public class HeaderBasedMessagingExtractor implements MessagingSpanTextMapExtrac
 				.traceIdHigh(traceId.length() == 32 ? Span.hexToId(traceId, 0) : 0)
 				.traceId(Span.hexToId(traceId))
 				.spanId(Span.hexToId(carrier.get(TraceMessageHeaders.SPAN_ID_NAME)));
-		spanBuilder.exportable(
+		String flags = carrier.get(TraceMessageHeaders.SPAN_FLAGS_NAME);
+		if (Span.SPAN_SAMPLED.equals(flags)) {
+			spanBuilder.exportable(true);
+		} else {
+			spanBuilder.exportable(
 				Span.SPAN_SAMPLED.equals(carrier.get(TraceMessageHeaders.SAMPLED_NAME)));
+		}
 		String processId = carrier.get(TraceMessageHeaders.PROCESS_ID_NAME);
 		String spanName = carrier.get(TraceMessageHeaders.SPAN_NAME_NAME);
 		if (spanName != null) {
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/HeaderBasedMessagingExtractor.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/HeaderBasedMessagingExtractor.java
@@ -19,8 +19,10 @@ public class HeaderBasedMessagingExtractor implements MessagingSpanTextMapExtrac
 	public Span joinTrace(SpanTextMap textMap) {
 		Map<String, String> carrier = TextMapUtil.asMap(textMap);
 		if (Span.SPAN_SAMPLED.equals(carrier.get(TraceMessageHeaders.SPAN_FLAGS_NAME))) {
-			generateIdIfMissing(carrier, TraceMessageHeaders.SPAN_ID_NAME);
-			generateIdIfMissing(carrier, TraceMessageHeaders.TRACE_ID_NAME);
+			String traceId = generateTraceIdIfMissing(carrier);
+			if (!carrier.containsKey(TraceMessageHeaders.SPAN_ID_NAME)) {
+				carrier.put(TraceMessageHeaders.SPAN_ID_NAME, traceId);
+			}
 		} else if (!hasHeader(carrier, TraceMessageHeaders.SPAN_ID_NAME)
 				|| !hasHeader(carrier, TraceMessageHeaders.TRACE_ID_NAME)) {
 			return null;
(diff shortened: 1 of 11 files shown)


spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/messaging/HeaderBasedMessagingExtractorTests.java
```diff
@@ -0,0 +1,101 @@
...
+
+import static org.springframework.cloud.sleuth.assertions.SleuthAssertions.then;
+
+/**
+ * @author Marcin Grzejszczak
+ */
+public class HeaderBasedMessagingExtractorTests {
+
+	@Test
+	public void overridesTheSampleFlagWithSpanFlagForSampledScenario() {
```
adriancole: maybe add a test for when only the flags header is present? Here's the code from finagle..

```
    } else if (request.headers.contains(Header.Flags)) {
      // even if there are no id headers we want to get the debug flag
      // this is to allow developers to just set the debug flag to ensure their
      // trace is collected
      Some(Trace.nextId.copy(flags = getFlags(request)))
```

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/TraceFilterTests.java
```diff
@@ -382,6 +382,38 @@ public void returns200IfSpanParentIsMalformedAndCreatesANewSpan() throws Excepti
...
+		TraceFilter filter = new TraceFilter(this.tracer, this.traceKeys, this.spanReporter,
+				this.spanExtractor, this.httpTraceKeysInjector);
+
+		filter.doFilter(this.request, this.response, this.filterChain);
+
+		then(new ListOfSpans(this.spanReporter.getSpans())).allSpansAreExportable();
+		then(TestSpanContextHolder.getCurrentSpan()).isNull();
+		then(ExceptionUtils.getLastException()).isNull();
+	}
+
```
adriancole: there's a subtle side-effect on debug not in this change. that is that it sets Span.debug=true. This makes sure the collector doesn't throw the span away either.. (note this only shows up when the server has collector sampling as a rate less than 1.0)

marcingrzejszczak: ah so you're saying that we need to update the collectors to check for the presence of that flag?

adriancole:
>
> ah so you're saying that we need to update the collectors to check for the
> presence of that flag?
>
yep or add a TODO and wait for someone to ask :P

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/TraceFilterTests.java
```diff
@@ -414,6 +414,42 @@ public void doesNotOverrideTheSampledFlagWhenXB3FlagIsSetToOtherValueThan1() thr
...
+				.allSpansAreExportable().hasSize(2).hasASpanWithSpanId(Span.hexToId("10"));
+		then(TestSpanContextHolder.getCurrentSpan()).isNull();
+		then(ExceptionUtils.getLastException()).isNull();
+	}
+
+	@Test
+	public void samplesWhenDebugFlagIsSetTo1AndOnlyTraceIdIsSet() throws Exception {
+		this.request = builder()
+				.header(Span.SPAN_FLAGS, 1)
+				.header(Span.TRACE_ID_NAME, 10L)
```
adriancole: heh mild lie :) trace id name is also set!

marcingrzejszczak: lol - indeed ;)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/HeaderBasedMessagingExtractor.java
```diff
@@ -17,14 +18,23 @@
 	@Override
 	public Span joinTrace(SpanTextMap textMap) {
 		Map<String, String> carrier = TextMapUtil.asMap(textMap);
-		if (!hasHeader(carrier, TraceMessageHeaders.SPAN_ID_NAME)
+		if (Span.SPAN_SAMPLED.equals(carrier.get(TraceMessageHeaders.SPAN_FLAGS_NAME))) {
+			generateIdIfMissing(carrier, TraceMessageHeaders.SPAN_ID_NAME);
```
adriancole: this is fine, though you could make it simpler by using the trace == span id when absent

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/ZipkinHttpSpanExtractor.java
```diff
@@ -33,7 +34,10 @@ public ZipkinHttpSpanExtractor(Pattern skipPattern) {
 	@Override
 	public Span joinTrace(SpanTextMap textMap) {
 		Map<String, String> carrier = TextMapUtil.asMap(textMap);
-		if (carrier.get(Span.TRACE_ID_NAME) == null) {
+		boolean debug = Span.SPAN_SAMPLED.equals(carrier.get(Span.SPAN_FLAGS));
+		if (debug) {
+			generateIdIfMissing(carrier, Span.TRACE_ID_NAME);
```
adriancole: does this need to also set span id? if not probably add a comment why

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/HeaderBasedMessagingExtractor.java
marcingrzejszczak: True!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/ZipkinHttpSpanExtractor.java
marcingrzejszczak: TraceId is enough cause later we're generating a span if it's missing. I'll add the comment

marcingrzejszczak:
done via https://github.com/spring-cloud/spring-cloud-sleuth/commit/c347f724cdb9cba96f89dbc0c828f99d614ebdfc

PR #501, opened 2017-01-24
Add a new log method to span that allows for setting the time

devinsba:
I have a use case where we have the time where an event took place but not the ability to create the log as it happens. This method will allow us to still log those events to sleuth.

This also makes the OpenTracing adapter I am working on more feature complete so that is good too.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
@@ -344,6 +344,13 @@ public class Span implements SpanContext {
 	}
 
 	/**
+	 * Add a {@link Log#event event} to a specific point in the timeline associated with this span.
+	 */
+	public void logEvent(long timestamp, String event) {
+		this.logs.add(new Log(timestamp, event));
+	}
+
+	/**
 	 * Sets a baggage item in the Span (and its SpanContext) as a key/value pair.
 	 *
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@devinsba Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=501)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=501) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/faq) for frequently asked questions.

pivotal-issuemaster:
@devinsba Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=501)!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
```diff
@@ -344,6 +344,13 @@ public void logEvent(String event) {
 	}
 
 	/**
+	 * Add a {@link Log#event event} to a specific point in the timeline associated with this span.
```
adriancole: probably makes sense to indicate the unit of the timestamp

marcingrzejszczak:
Congrats on the PR @devinsba :D

Issue #431, opened 2016-10-27
Zipkin annotations - cr(ClientReceive) before ss(ServerSend)


nklmish:
I have two services developed using spring cloud (tried both Camden.SR1 and Camden.RELEASE) and spring-cloud-starter-zipkin (plain vanilla - no hystrix, rxjava, **async** etc). I am running both services on a single machine, with zipkin **(1.13.1)** running inside a docker container and using and in memory database. 

Service A invokes Service B using `org.springframework.web.client.RestTemplate`. 

It seems that sometime **cr** (client receive) annotation is calculated before **ss**(server send). However, as per[ docs](http://zipkin.io/pages/instrumenting.html)
-  **cr** :  _the client has received the response from the server. This sets the end of the span_ 
- **ss**:  _that the server has completed processing and has sent the request back to the client_. 

A client should receive response only after it was sent by the server.  Below, you can see that for an identical request, I am receiving annotations in different order. 

You can also find the **minimal demo illustrating this problem**  [here](https://github.com/nklmish/issue-cr-ss)

![trace-overview](https://cloud.githubusercontent.com/assets/13149921/19765265/b144951c-9c47-11e6-890b-ac0aac529877.png)

<img width="1156" alt="screen shot 2016-10-27 at 02 56 34" src="https://cloud.githubusercontent.com/assets/13149921/19750644/2c44856a-9bf3-11e6-8a35-6b2b06fcb22a.png"> correct order
<img width="1157" alt="screen shot 2016-10-27 at 02 56 27" src="https://cloud.githubusercontent.com/assets/13149921/19750645/2c6dceca-9bf3-11e6-971d-bf43713f5b7c.png"> cr before ss
<img width="950" alt="screen shot 2016-10-27 at 02 50 37" src="https://cloud.githubusercontent.com/assets/13149921/19751280/60cc57e0-9bf8-11e6-86bb-860383cf851c.png"> cr before ss

[JsonForBadTrace.txt](https://github.com/spring-cloud/spring-cloud-sleuth/files/555532/JsonForBadTrace.txt)

adriancole:
just guessing, but the cr before ss smells like the server-side of the span
is closing late, or it is being added only after the response is fully read
by the client.

nklmish:
After debugging, It looks like a problem is on the side of spring cloud sleuth. I am summarising my findings below :

[...]

PR #503, opened 2017-01-30
Fixing the moment when SS is set

### LINKED ISSUE #492: Add integration tests to ensure that the order of spans is maintained
and the timestamps are correct for the RPC spans. CS first, SR then, SS next, CR at the end, obviously :)

marcingrzejszczak:
without this change there's a problem with the time when the SS is set on a span. Currently it's done  in `TraceFilter`'s finally block. The problem is that this code is executed after the response has been sent back to the client. Thus CR sometimes was set faster than SS (it doesn't make any sense from the logical point of view).

with this change we move setting the SS on the span to the `TraceHandlerInterceptor`.  We leave the mechanism of `TraceFilter` stamping the span SS as a fallback one. That means that if for some reason the `TraceHandlerInterceptor` wasn't executed then we need to try to save the situation. 

also this change fixes ensures that the `TraceHandlerInterceptor` is available for Zuul. Until now even though we've added it to the list of interceptors it was never executed at runtime. That's why we're wrapping the `ZuulHandlerMapping` with our own wrapper and we're initializing the interceptors for it.

fixes #492 #431

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/SpanCloser.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/SpanCloser.java
@@ -1,0 +1,150 @@
+/*
+ * Copyright 2013-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.sleuth.instrument.web;
+
+import java.lang.invoke.MethodHandles;
+import javax.servlet.http.HttpServletRequest;
+import javax.servlet.http.HttpServletResponse;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.springframework.beans.factory.BeanFactory;
+import org.springframework.cloud.sleuth.Span;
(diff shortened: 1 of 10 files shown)


spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/assertions/ListOfSpansAssert.java
```diff
@@ -155,4 +164,51 @@ protected void failWithMessage(String errorMessage, Object... arguments) {
...
+				break;
+			default:
+				break;
+			}
+		}));
+		return rpcLogKeeper;
+	}
+}
+
+class RpcLogKeeper {
```
adriancole: neat

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/SpanCloser.java
```diff
@@ -0,0 +1,150 @@
...
+import org.springframework.cloud.sleuth.SpanReporter;
+import org.springframework.cloud.sleuth.Tracer;
+import org.springframework.http.HttpStatus;
+
+/**
+ * A helper class to close spans
+ *
+ * @author Marcin Grzejszczak
+ */
+class SpanCloser {
```
adriancole: cool thanks for not exposing this

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/SpanCloser.java
```diff
@@ -0,0 +1,150 @@
...
+		this.beanFactory = beanFactory;
+	}
+
+	void detachOrCloseSpans(HttpServletRequest request,
+			HttpServletResponse response, Span spanFromRequest, Throwable exception) {
+		Span span = spanFromRequest;
+		if (log.isDebugEnabled()) {
+			log.debug("Will try to detach or close " + span + "");
+		}
+		if (span != null) {
```
adriancole: I know this code is mostly moved, but if you early return when span is null, it saves a level of indent

marcingrzejszczak:
Actually I have to close this cause it's not working ;)

PR #506, opened 2017-01-31
Fixing the moment when SS is set

### LINKED ISSUE #492: Add integration tests to ensure that the order of spans is maintained

marcingrzejszczak:
without this change there's a problem with the time when the SS is set on a span. Currently it's done in TraceFilter's finally block. The problem is that this code is executed after the response has been sent back to the client. Thus CR sometimes was set faster than SS (it doesn't make any sense from the logical point of view).

with this change we're introducing wrappers over the HttpServletResponse where we annotate the span with SS just after the response gets sent to the recipient.

fixes #492 #431


marcingrzejszczak:
Done via - https://github.com/spring-cloud/spring-cloud-sleuth/commit/b996de50eafa26c9e5dd091a5ee0dadbb2018048

Issue #276, opened 2016-05-13
Exception when connect to a Websocket


rafaelhz:
I'm using Spring Cloud Sleuth in some projects and everything is working well. But in a project where I have a websocket, an exception is thrown when connected to the websocket.

To simulate the error, clone the spring websocket demo app: 
https://github.com/spring-guides/gs-messaging-stomp-websocket

After just include the dependency: (Without the dependency everything works)
```
        <dependency>
            <groupId>org.springframework.cloud</groupId>
            <artifactId>spring-cloud-starter-sleuth</artifactId>
            <version>1.0.0.RELEASE</version>
        </dependency>
```

After run the application, connect to the websocket, you will see the exception:
```
Exception in thread "clientInboundChannel-2" org.springframework.messaging.MessageDeliveryException: Failed to handle GenericMessage [...]; nested exception is java.lang.IllegalStateException: No header accessor (not using the SimpMessagingTemplate?): GenericMessage [...]

Caused by: java.lang.IllegalStateException: No header accessor (not using the SimpMessagingTemplate?): GenericMessage [...]
at org.springframework.messaging.simp.broker.SimpleBrokerMessageHandler.handleMessageInternal(SimpleBrokerMessageHandler.java:228)
	at org.springframework.messaging.simp.broker.AbstractBrokerMessageHandler.handleMessage(AbstractBrokerMessageHandler.java:238)
	at org.springframework.messaging.support.ExecutorSubscribableChannel$SendTask.run(ExecutorSubscribableChannel.java:135)
	... 3 more
```

gauravrmazra:
@dsyer this issue is same what I raised few months back. This is related to sending stomp based messages over websocet. If you remember few months back you removed some Stomp related interceptor and i claim it won't work. I will attach more  detail wrn have access to Laptop.

dsyer:
I'm interested in why it doesn't work in Spring Websocket land. See here: https://jira.spring.io/browse/SPR-14279.

dsyer:
Should be fixed in Spring 4.2.7 according the the Spring ticket above.

marcingrzejszczak:
@gauravrmazra , @rafaelhz does this work for you?

gauravrmazra:
@marcingrzejszczak  We are not using this feature(STOMP messages) anymore.

ArthurasJ:
I guess I have the same problem: 
[...]

PR #521, opened 2017-02-17
Makes message serializable

### LINKED ISSUE #520: TraceChannelInterceptor makes Message NonSerializable
The code:
```
MessageHeaderAccessor headers = MessageHeaderAccessor.getMutableAccessor(message);
headers.copyHeaders(messageBuilder.build().getHeaders());
return new GenericMessage<Object>(message.getPayload(), headers.getMessageHeaders());
```
returns the `GenericMessage` with the `MutableMessageHeaders` which has a reference to the outer `MessageHeaderAccessor` instance which isn't `Serializable`.

Would be better to use `MessageHeaderAccessor.toMessageHeaders()`. At least for now.
Although it clashes with the fix for #276...

For more info: https://jira.spring.io/browse/SPR-15262

marcingrzejszczak:
without this change exceptions where thrown due to the fact that we were using a wrong method to retrieve headers
with this change that's changed

fixes #520

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TraceChannelInterceptor.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TraceChannelInterceptor.java
@@ -90,7 +90,7 @@ public class TraceChannelInterceptor extends AbstractTraceChannelInterceptor {
 		getSpanInjector().inject(span, messageBuilder);
 		MessageHeaderAccessor headers = MessageHeaderAccessor.getMutableAccessor(message);
 		headers.copyHeaders(messageBuilder.build().getHeaders());
-		return new GenericMessage<Object>(message.getPayload(), headers.getMessageHeaders());
+		return new GenericMessage<Object>(message.getPayload(), headers.toMessageHeaders());
 	}
 
 	private Span startSpan(Span span, String name, Message<?> message) {
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TraceChannelInterceptor.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TraceChannelInterceptor.java
@@ -90,7 +90,7 @@ public class TraceChannelInterceptor extends AbstractTraceChannelInterceptor {
 		getSpanInjector().inject(span, messageBuilder);
 		MessageHeaderAccessor headers = MessageHeaderAccessor.getMutableAccessor(message);
 		headers.copyHeaders(messageBuilder.build().getHeaders());
-		return new GenericMessage<Object>(message.getPayload(), headers.toMessageHeaders());
+		return new GenericMessage<Object>(message.getPayload(), headers.getMessageHeaders());
 	}
 
 	private Span startSpan(Span span, String name, Message<?> message) {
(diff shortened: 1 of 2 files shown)


marcingrzejszczak:
cc @dsyer I had to remove a test that checks for mutability of headers but as far as I understand @artembilan stuff was fixed in Spring Framework and we no longer need to have that mutability? We will not introduce regression with this change (https://github.com/spring-cloud/spring-cloud-sleuth/issues/276) ? The only test that we had for that issue is the one that I'm removing so I'd really prefer to have some new tests (but currently I'd need support how to write one)

artembilan:
See https://jira.spring.io/browse/SPR-15262. So, you should place a message to your interceptor and perform that serialization hacks.

marcingrzejszczak:
@dsyer @artembilan this will fail due to this test failing

```


	@Test
	public void nonExportableSpanCreation() {
		this.tracedChannel.send(MessageBuilder.withPayload("hi")
				.setHeader(TraceMessageHeaders.SAMPLED_NAME, Span.SPAN_NOT_SAMPLED).build());
		assertNotNull("message was null", this.message);

		String spanId = this.message.getHeaders().get(TraceMessageHeaders.SPAN_ID_NAME, String.class);
		then(spanId).isNotNull();
		then(TestSpanContextHolder.getCurrentSpan()).isNull();
		then(this.span.isExportable()).isFalse();
	}
```

Should remove this test due to the fact it got fixed somewhere? I didn't get the outcome of the discussions on Jira with whether it got fixed or not in Spring Core.

marcingrzejszczak:
Ah yeah true. Check this commit - https://github.com/spring-cloud/spring-cloud-sleuth/pull/521/commits/85ba1904cadcb5b6f1c1fc985bff7e51e4c509a8 . Over there I made the span serializable and the added test passes but the `nonExportableSpanCreation` doesn't . With the current changes the `messageHeadersStillMutable` passes but the `nonExportableSpanCreation` doesn't. So the only wayis  to do the change from the commit and remove `messageHeadersStillMutable`. But it means that the headers are not mutable.

artembilan:
@marcingrzejszczak , @dsyer , would you mind taking a look into your release plans on the matter?

Look, SF JIRA https://jira.spring.io/browse/SPR-15262 is for `4.3.7`, and its plan is Fabruary 28.

My idea to revert the change you did in the `TraceChannelInterceptor` and rely on the fact that it is fixed in the SF its own way. That serialization test will pass any way. Plus your `messageHeadersStillMutable` should work as well.

WDYT?

marcingrzejszczak:
So you say that I should return to the `getHeaders()` and it will work OOB? Sounds perfect for me.

artembilan:
Yes, ` headers.getMessageHeaders()` will work, but with the latest SF `4.3.7`. Right now only in `BUILD-SNAPSHOT`

marcingrzejszczak:
Hmm but we take the version from Boot so we'd need to have the version updated from Boot and get the new Boot version. Or change the SF version only for Sleuth. WDYT @dsyer ?

artembilan:
Spring Boot `1.5.2` is planned for March 2d.
Now everything depends of your release plans.

I think there is simple workaround like:

```
public class ImmutableMessageInterceptor extends ChannelInterceptorAdapter {

	@Override
	public Message<?> preSend(Message<?> message, MessageChannel channel) {
		MessageHeaderAccessor headers = MessageHeaderAccessor.getMutableAccessor(message);
		return  new GenericMessage<Object>(message.getPayload(), headers.toMessageHeaders());
	}

}
```
And we have to place this interceptor just after `TraceChannelInterceptor`.

So, this your fix (actually just a new serialization test) may really wait until the proper Spring Boot version.

artembilan:
@marcingrzejszczak ,

Here you can find the workaround we have talked about: https://github.com/artembilan/spring-cloud-sleuth/commit/91e9b801b6fb205314f2dcdecf9adb19f571f486

I guess you can pull my commit to your PR.
Or... I can just apply it locally and merge! 😄

marcingrzejszczak:
Done via 7caf522

PR #526, opened 2017-02-21
Managing spans with annotations

### LINKED ISSUE #182: Library aspect for tracing
I am experimenting with Sleuth at the moment and am a little bit wondering about what's the best way to create Spans at points not catched by the automatic instrumentation.
For myself i created an annotation ("SleuthInstrumented") and an AOP Advice which matches this is on a method or a class. The advice then creates a new Span either with a name defined in the annoation or simply "CLASSNAME/METHODNAME" as name. Additionally you can name method parameters in the annotation which should be added as tags to span.

Do you think this is a good way to handle this?
If it is: Do you think it would be something i should contribute?
If it is not: What would you do?

PS:  Very nice work you are making here! Works really good :)

marcingrzejszczak:
## Rationale

The main arguments for this features are

* api-agnostic means to collaborate with a span
    - use of annotations allows users to add to a span with no library dependency on a span api.
    This allows Sleuth to change its core api less impact to user code.
* reduced surface area for basic span operations.
    - without this feature one has to use the span api, which has lifecycle commands that
    could be used incorrectly. By only exposing scope, tag and log functionality, users can
    collaborate without accidentally breaking span lifecycle.
* collaboration with runtime generated code
    - with libraries such as Spring Data / Feign the implementations of interfaces are generated
    at runtime thus span wrapping of objects was tedious. Now you can provide annotations
     over interfaces and arguments of those interfaces

## Creating new spans

If you really don’t want to take care of creating local spans manually you can profit from the `@NewSpan` annotation. Also we give you the `@SpanTag` annotation to add tags in an automated fashion.

Let’s look at some examples of usage.

```
@NewSpan
void testMethod();
```

Annotating the method without any parameter will lead to a creation of a new span whose name will be equal to annotated method name.

```
@NewSpan("customNameOnTestMethod4")
void testMethod4();
```

If you provide the value in the annotation (either directly or via the name parameter) then the created span will have the name as the provided value.

```
// method declaration
@NewSpan(name = "customNameOnTestMethod5")
void testMethod5(@SpanTag("testTag") String param);

// and method execution
this.testBean.testMethod5("test");
```

You can combine both the name and a tag. Let’s focus on the latter. In this case whatever the value of the annotated method’s parameter runtime value will be - that will be the value of the tag. In our sample the tag key will be `testTag` and the tag value will be `test`.

```
@NewSpan(name = "customNameOnTestMethod3")
@Override
public void testMethod3() {
}
```

You can place the `@NewSpan` annotation on both the class and an interface. If you override the interface’s method and provide a different value of the `@NewSpan` annotation then the most concrete one wins (in this case `customNameOnTestMethod3` will be set).

## Continuing spans

If you want to just add tags and annotations to an existing span it’s enough to use the `@ContinueSpan` annotation as presented below:

```
// method declaration
@ContinueSpan(log = "testMethod11")
void testMethod11(@SpanTag("testTag11") String param);

// method execution
this.testBean.testMethod11("test");
```

That way the span will get continued and: 
- logs with name `testMethod11.before` and `testMethod11.after` will be created 
- if an exception will be thrown a log `testMethod11.afterFailure` will also be created
- tag with key `testTag11` and value `test` will be created

## More advanced tag setting

There are 3 different ways to add tags to a span. All of them are controlled by the SpanTag annotation. Precedence is:

- try with the bean of `TagValueResolver` type and provided name
- if one hasn't provided the bean name, try to evaluate an expression. We're searching for a `TagValueExpressionResolver` bean.
The default implementation uses SPEL expression resolution.
- if one hasn't provided any expression to evaluate just return a `toString()` value of the parameter

if there’s no SPEL expression just return a `toString()` value of the parameter

### Custom extractor

The value of the tag for following method will be computed by an implementation of `TagValueResolver` interface. Its class name has to be passed as the value of the `resolver` attribute.

Having such an annotated method:

```
@NewSpan
public void getAnnotationForTagValueResolver(@SpanTag(key = "test", resolver = TagValueResolver.class) String test) {
}
```

and such a `TagValueResolver` bean:

```
@Bean(name = "myCustomTagValueResolver")
public SleuthTagValueResolver tagValueResolver() {
	return parameter -> "Value from myCustomTagValueResolver";
}
```

Will lead to setting of a tag value equal to `Value from myCustomTagValueResolver`.

### Using SpEL expression for value

Having such an annotated method:

```
@NewSpan
public void getAnnotationForTagValueExpression(@SpanTag(key = "test", expression = "length() + ' characters'") String test) {
}
```

and no custom implementation of a TagValueExpressionResolver will lead to evaluation of the SPEL expression and a tag with value 4 characters will be set on the span. If you want to use some other expression resolution mechanism you can create your own implementation of the bean.

### Using toString method

Having such an annotated method:

```
@NewSpan
public void getAnnotationForArgumentToString(@SpanTag("test") Long param) {
}
```

if executed with a value of `15` will lead to setting of a tag with a String value of `"15"`.

This PR is an adoption of @Koizumi85 work started here - https://github.com/Koizumi85/spring-cloud-sleuth-annotation

fixes #182

cc @dsyer @adriancole @ryanjbaxter @spencergibb


marcingrzejszczak:
## `@NewSpan` and `@SpanTag` annotations

If you really don’t want to take care of local spans creation you can profit from the @NewSpan annotation. Also we give you the @SpanTag annotation to add tags in an automated fashion.

Let’s look at some examples of usage.

```
@NewSpan
void testMethod();
```

Annotating the method without any parameter will lead to a creation of a new span whose name will be equal to annotated method name.

```
@NewSpan("customNameOnTestMethod4")
void testMethod4();
```

If you provide the value in the annotation (either directly or via the name parameter) then the created span will have the name as the provided value.

```
// method declaration
@NewSpan(name = "customNameOnTestMethod5")
void testMethod5(@SpanTag("testTag") String param);

// and method execution
this.testBean.testMethod5("test");
```

You can combine both the name and a tag. Let’s focus on the latter. In this case whatever the value of the annotated method’s parameter runtime value will be - that will be the value of the tag. In our sample the tag key will be testTag and the tag value will be test.

```
@NewSpan(name = "customNameOnTestMethod3")
@Override
public void testMethod3() {
}
```

You can place the `@NewSpan` annotation on both the class and an interface. If you override the interface’s method and provide a different value of the `@NewSpan` annotation then the most concrete one wins (in this case `customNameOnTestMethod3` will be set).

###  More advanced tag setting

There are 3 different ways to add tags to a span. All of them are controlled by the SpanTag annotation. Precedence is:

- try with the SleuthTagValueResolver bean
- if there is no bean, try to evaluate a SPEL expression
- if there’s no SPEL expression just return a toString() value of the parameter

#### Custom SleuthTagValueResolver Bean

The value of the tag for following method will be computed by an implementation of SleuthTagValueResolver interface.

Having such an annotated method:

```
@NewSpan
public void getAnnotationForTagValueResolver(@SpanTag(value = "test", tagValueResolverBeanName = "myCustomTagValueResolver") String test) {
}
```

and such a `SleuthTagValueResolver` bean implementation

```
@Bean(name = "myCustomTagValueResolver")
public SleuthTagValueResolver tagValueResolver() {
	return parameter -> "Value from myCustomTagValueResolver";
}
```

Will lead to setting of a tag value equal to Value from myCustomTagValueResolver.

#### Using SpEL expression for value

Having such an annotated method:

```
@NewSpan
public void getAnnotationForTagValueExpression(@SpanTag(value = "test", tagValueExpression = "length() + ' characters'") String test) {
}
```

Will lead to evaluation of the SPEL expression and a tag with value `4 characters` will be set on the span.

####Using toString method

Having such an annotated method:

```
@NewSpan
public void getAnnotationForArgumentToString(@SpanTag(value = "test") Long param) {
}
```

if executed with a value of 15 will lead to setting of a tag with a String value of `"15"`.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/DefaultSleuthSpanCreator.java
```diff
@@ -0,0 +1,56 @@
...
+
+	@Override public Span createSpan(JoinPoint pjp, NewSpan newSpanAnnotation) {
+		String key = StringUtils.isEmpty(newSpanAnnotation.name()) ?
+				pjp.getSignature().getName() : newSpanAnnotation.name();
+		Span span = createSpan(key);
+		this.annotationSpanUtil.addAnnotatedParameters(pjp);
+		return span;
+	}
+
+	private Span createSpan(String key) {
```
marcingrzejszczak: @Koizumi85 I've changed this to ALWAYS create spans. That's a change in comparison to your impl (you were creating spans only if tracing was there).

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/annotation/SleuthSpanCreatorAspectNegativeTests.java
```diff
@@ -0,0 +1,156 @@
...
+		then(this.accumulator.getSpans()).isEmpty();
+		then(ExceptionUtils.getLastException()).isNull();
+	}
+
+	@Test
+	public void shouldCallAdviceForAnnotatedBean() throws Throwable {
+		this.annotatedTestBean.testMethod();
+
+		List<Span> spans = new ArrayList<>(this.accumulator.getSpans());
+		then(new ListOfSpans(spans)).hasSize(1).hasASpanWithName("testMethod");
```
marcingrzejszczak: @Koizumi85 I've also change testing of the implementation vs checking the results. We don't care if some methods got executed. We care that a span with a given name was created.

Koizumi85: Hm yes. Had no idea how to get the created spans there :D

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/DefaultSleuthSpanCreator.java
Koizumi85: @marcingrzejszczak Yes. I remember we talked about this in the past. Wasn't sure if a span is useful if there is no tracing. :)

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/annotation/SleuthSpanCreatorAspectNegativeTests.java
marcingrzejszczak: I've forgotten about this discussion. I came to the conclusion that if you do `@NewSpan` then it means that you want a new span. Not creating one would be surprising.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/DefaultSleuthSpanCreator.java
Koizumi85: Yes. You are right. The span should be created regardless of tracing or not.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthSpanCreateBeanPostProcessor.java
```diff
@@ -0,0 +1,89 @@
...
+/**
+ * This bean post processor analyzes whether the bean is eligible to
+ * be wrapped with an aspect. That will be the case only if there are
+ * {@link NewSpan} annotations either on the class or the interface
+ * methods.
+ *
+ * @author Christian Schwerdtfeger
+ * @since 1.2.0
+ */
+class SleuthSpanCreateBeanPostProcessor implements BeanPostProcessor {
```
gavlyukovskiy: Aplying aspects through `BeanPostProcessor` is not good approach, in case of circular dependency `BeanCurrentlyInCreationException` will be thrown:
```
@Service
public class Service2 {

    @Autowired
    private Service1 service1;

    @NewSpan
    public void serve2() {
    }
}

@Service
public class Service1 {

    @Autowired
    private Service2 service2;

    @NewSpan
    public void serve1() {
    }
}
```

leads to 
> BeanCurrentlyInCreationException: Error creating bean with name 'service1': Bean with name 'service1' has been injected into other beans [service2] in its raw version as part of a circular reference, but has eventually been wrapped. This means that said other beans do not use the final version of the bean. This is often the result of over-eager type matching - consider using 'getBeanNamesOfType' with the 'allowEagerInit' flag turned off, for example.

Creation aspect bean in this case much better. Also I tried to place annotation on the class and it doesn't working.
I didn't understand the issue with pointcut for interfaces with annotated methods, but I tried to register aspect with following signature:
```
@Pointcut("execution(public * *(..)) && @annotation(newSpan)")
public void annotatedMethod(NewSpan newSpan) {
}

@Pointcut("execution(public * *(..)) && @within(newSpan)")
public void annotatedClass(NewSpan newSpan) {
}

@Around("annotatedMethod(newSpan) || annotatedClass(newSpan)")
public Object instrumentOnMethodAnnotation(ProceedingJoinPoint pjp, NewSpan newSpan) throws Throwable {
	Method method = getMethod(pjp);
        Assert.notNull(method, "method can't be null inside aspecting pointcut");
	Span span = null;
	try {
		span = this.spanCreator.createSpan(pjp, newSpan);
		return pjp.proceed();
	} finally {
		if (span != null) {
			this.tracer.close(span);
		}
	}
}
```
It works fine for both classes and methods, also we don't need `SleuthAnnotationUtils` to find annotation.

gavlyukovskiy:
Also please consider using events instead of spans, I have suggested to have boolean flag to change spans to the events, it requres to change annotation name. Anyway it's more conventional to have annotation named Spannable or Traceable.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthSpanCreateBeanPostProcessor.java
gavlyukovskiy: Also it may be useful to catch an exception on pjp.proceed() and add message as tag `<spanName>.error`

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAnnotationAutoConfiguration.java
```diff
@@ -0,0 +1,57 @@
...
+ * {@link NewSpan} annotation. You can annotate classes or just methods.
+ * You can also apply this annotation to an interface.
+ *
+ * @author Christian Schwerdtfeger
+ * @since 1.2.0
+ */
+@Configuration
+@ConditionalOnBean(Tracer.class)
+@ConditionalOnProperty(name = "spring.sleuth.annotation.enabled", matchIfMissing = true)
+public class SleuthAnnotationAutoConfiguration {
```
gavlyukovskiy: And in case of using @AutoConfigureAfter(TraceAutoConfiguration.class) we always have Tracer bean, so this condition can be removed.

marcingrzejszczak: That's not necessary. If Sleuth is disabled things work fine. Boot will automatically pick the proper order of auto configuration. In all the tests that we have we don't define a custom `Tracer` and it's working like a charm

gavlyukovskiy: I tried case when it's enabled. Simple application without any customizations, autoconfiguration report:

> SleuthAnnotationAutoConfiguration:
>   Did not match:
>      - @ConditionalOnBean (types: org.springframework.cloud.sleuth.Tracer; SearchStrategy: all) did not find any beans (OnBeanCondition)
>   Matched:
>      - @ConditionalOnProperty (spring.sleuth.annotation.enabled) matched (OnPropertyCondition)

marcingrzejszczak: Ok I'll add it - it's not a problem to ensure it's in proper order

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
```diff
@@ -225,28 +238,45 @@ public void doWith(Method method) throws IllegalArgumentException,
...
+				}
+				if (newSpan != null) {
+					tracer().close(span);
 				}
-				tracer().close(span);
 			}
 		}
 	}
 
+	private String log(NewSpan annotation, ContinueSpan continueSpan) {
```
gavlyukovskiy: So what the meaning of `log` inside `@NewSpan`?
since it creates new span and all timings are present in binary annotations

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
```diff
@@ -225,28 +238,45 @@ public void doWith(Method method) throws IllegalArgumentException,
...
-				span.logEvent(annotation.log() + ".start");
+				spanTagAnnotationHandler().addAnnotatedParameters(invocation);
+				span.logEvent(log + ".start");
 			}
 			return invocation.proceed();
 		} finally {
 			if (span != null) {
 				if (hasLog) {
-					span.logEvent(annotation.log() + ".end");
+					span.logEvent(log + ".end");
```
gavlyukovskiy: If event name suggested to be an accomplished act e.g. `igredientsReceived` it would be strange to have events `igredientsReceived.start` and `igredientsReceived.end`. 
Otherwise `log` may be optional and equal to method name by default `@ContinueSpan ingridientService.receiveIngridients()` would produce events `receiveIngridients.start` and `receiveIngridients.end`

marcingrzejszczak: In that case I wouldn't use the feature of `log` and would just manually create the log myself using `tracer`.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/ContinueSpan.java
```diff
@@ -0,0 +1,42 @@
+/*
```
devinsba: Wondering if these annotations should be in a separate maven module so that other libraries can include them without all of sleuth being required

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
gavlyukovskiy: In which one?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/ContinueSpan.java
marcingrzejszczak: I'd leave them in Sleuth so that we see that they are of any value. If more people will use it we can consider pushing it somewhere else. WDYT?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
marcingrzejszczak: I'd the following

```
public void startIngredientsAggregation() {
this.tracer.getCurrentSpan().log("ingredientsReceived");
}
```

gavlyukovskiy: make sense, so in case of using annotation it should be like 
```
@ContinueSpan(log = "ingredientsAggregation")
public void startIngredientsAggregation() {
}
```
and we can took just method name by default, right?

marcingrzejszczak: > and we can took just method name by default, right?

don't follow?

in this case you will have 2 logs `ingredientsAggregation.start` and `ingredientsAggregation.end`

gavlyukovskiy: I mean that 
```
@ContinueSpan
public void startIngredientsAggregation() {
}
```
is enough in this case (without `log`)

marcingrzejszczak: No cause I don't want to explicitly log this method.

gavlyukovskiy: oh, got it, I can use it if I want to add something to the current span using `@SpanTag` without event, right?

marcingrzejszczak: yes, like this:

```
// method declaration
@ContinueSpan
void testMethod11(@SpanTag("testTag11") String param);

// method execution
this.testBean.testMethod11("test");
```

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/ContinueSpan.java
devinsba: Makes sense, just wondering if the spring-data folks would add tags/spans on things in the main data modules if it meant they didn't have to bring in all of sleuth

marcingrzejszczak: I think that the users of Spring Data that want to have some stuff traced will be able to extend the interfaces and add the Sleuth repositories. Why would someone want to use these annotations without Sleuth?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAnnotationAutoConfiguration.java
```diff
@@ -0,0 +1,61 @@
...
+ * Auto-configuration} that allows creating spans by means of a
+ * {@link NewSpan} annotation. You can annotate classes or just methods.
+ * You can also apply this annotation to an interface.
+ *
+ * @author Christian Schwerdtfeger
+ * @since 1.2.0
+ */
+@Configuration
+@ConditionalOnBean(Tracer.class)
+@ConditionalOnProperty(name = "spring.sleuth.annotation.enabled", matchIfMissing = true)
```
gavlyukovskiy: It would be great to add metadata for this property in `additional-spring-configuration-metadata.json`

marcingrzejszczak: Fixed

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAnnotatedParameter.java
```diff
@@ -0,0 +1,44 @@
...
+ * of the annotated method argument.
+ *
+ * @author Christian Schwerdtfeger
+ * @since 1.2.0
+ */
+class SleuthAnnotatedParameter {
+
+	int parameterIndex;
+
+	Object annotation;
```
gavlyukovskiy: As I understood after removing `@SpanLog` there is no reasons to have `Object`

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAnnotationAutoConfiguration.java
```diff
@@ -0,0 +1,63 @@
...
+	@Autowired private Tracer tracer;
+	
+	@Bean SpanTagAnnotationHandler spanTagAnnotationHandler(ApplicationContext context) {
+		return new SpanTagAnnotationHandler(context, this.tracer);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean(SpanCreator.class)
+	SpanCreator spanCreator(ApplicationContext context) {
+		return new DefaultSpanCreator(this.tracer, spanTagAnnotationHandler(context));
```
gavlyukovskiy: As for I know using `spanTagAnnotationHandler(context)` leads to creation of another object, not using bean (even when configuration became proxy since it is call through `this`). Why don't you take `SpanTagAnnotationHandler` in arguments?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/DefaultSpanCreator.java
```diff
@@ -0,0 +1,56 @@
...
+ */
+class DefaultSpanCreator implements SpanCreator {
+
+	private final Tracer tracer;
+	private final SpanTagAnnotationHandler annotationSpanUtil;
+
+	DefaultSpanCreator(Tracer tracer,
+			SpanTagAnnotationHandler annotationSpanUtil) {
+		this.tracer = tracer;
+		this.annotationSpanUtil = annotationSpanUtil;
```
gavlyukovskiy: After `SpanTagAnnotationHandler` make sense to rename variable as well.

marcingrzejszczak:
I think that the `log` on a `@NewSpan` makes no sense. We're duplicating the beginning and the end. That's really pointless IMO.

As for log on `@ContinueSpan` that theoretically makes sense cause you're automatically annotating a beginning and end of some method. Since we're not creating a new span then this information is quite interesting to know when something measurable has started and finished. If an event is a concrete action that is interesting for us (not when it has started and finished) then we can add a log manually as described in previous discussions.

WDYT @adriancole @gavlyukovskiy @devinsba

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/NewSpan.java
```diff
@@ -0,0 +1,59 @@
...
+	 * The name of the span which will be created. Default is the annotated method's name.
+	 */
+	@AliasFor("name")
+	String value() default "";
+
+	/**
+	 * The value passed to the annotation will be used and the framework
+	 * will create two events with the {@code .start} and {@code .end} suffixes
+	 */
+	String log() default "";
```
gavlyukovskiy: You've missed to remove `log` here.

marcingrzejszczak: Great catch - thanks!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/ContinueSpan.java
adriancole: It could be of value since sleuth has an opinion inherited by spring boot etc. spring data may be compatible with those opinions but others may not be.

regardless, we can wait until more request an "api jar" as that can be spun out at any time. There's also a possibility to allow any annotation!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/NewSpan.java
```diff
@@ -0,0 +1,54 @@
...
+ * @author Christian Schwerdtfeger
+ * @since 1.2.0
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Inherited
+@Target(value = { ElementType.METHOD, ElementType.TYPE })
+public @interface NewSpan {
+
+	/**
+	 * The name of the span which will be created. Default is the annotated method's name.
```
adriancole: anything about case etc here? for example, I think we usually encourage lower-hyphen...

marcingrzejszczak: Ah true - I've forgotten about this!

marcingrzejszczak: Fixed ^^

adriancole:
if the key part of SpanTag is "key" then why not "valueExpression" vs "tagValueExpression"?

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
```diff
@@ -201,6 +201,125 @@ include::../../../../spring-cloud-sleuth-core/src/test/java/org/springframework/
 
 will lead in creating a span named `calculateTax`.
 
+== Managing spans with annotations
```
adriancole: this should be the summary name of the PR and the release note

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
```diff
@@ -201,6 +201,125 @@ include::../../../../spring-cloud-sleuth-core/src/test/java/org/springframework/
...
+----
+include::../../../../spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/annotation/SleuthSpanCreatorAspectTests.java[tags=annotated_method,indent=0]
+----
+
+Annotating the method without any parameter will lead to a creation of a new span whose name
+will be equal to annotated method name.
+
+[source,java]
+----
+include::../../../../spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/annotation/SleuthSpanCreatorAspectTests.java[tags=custom_name_on_annotated_method,indent=0]
```
adriancole: I notice that "View" doesn't render this in github. maybe should be a github issue as otherwise we need to assume it works..

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/DefaultSpanCreator.java
```diff
@@ -0,0 +1,63 @@
...
+import org.springframework.cloud.sleuth.Span;
+import org.springframework.cloud.sleuth.Tracer;
+import org.springframework.cloud.sleuth.util.SpanNameUtil;
+import org.springframework.util.StringUtils;
+
+/**
+ * Default implementation of the {@link SpanCreator} that creates
+ * a new span around the annotated method.
+ *
+ * @author Christian Schwerdtfeger
```
adriancole: careful with those author tags!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/NewSpan.java
```diff
@@ -0,0 +1,54 @@
...
+import org.springframework.core.annotation.AliasFor;
+
+/**
+ * Allows to create a new span around a public method or a class.
+ * <p>
+ * For each public method in an annotated class, or self annotated method,
+ * a new {@link org.springframework.cloud.sleuth.Span} will be created.
+ * Method parameters can be annotated with {@link SpanTag}, which will
+ * in adding the parameter value as a tag to the span.
+ *
```
adriancole: maybe a note here about being a child if a trace is already in progress

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
```diff
@@ -0,0 +1,306 @@
...
+
+		private Class<? extends Annotation> annotationType;
+
+		public AnnotationMethodsResolver(Class<? extends Annotation> annotationType) {
+			this.annotationType = annotationType;
+		}
+
+		public boolean hasAnnotatedMethods(Class<?> clazz) {
+			final AtomicBoolean found = new AtomicBoolean(false);
+			ReflectionUtils.doWithMethods(clazz,
```
adriancole: one day.. retrolambda :)

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
spencergibb: Yeah, it's part of the docs build.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAnnotationAutoConfiguration.java
adriancole: where is this fixed? (maybe link to the line.. I am curious as can't find it)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SpanTag.java
```diff
@@ -0,0 +1,62 @@
...
+	/**
+	 * The name of the key of the tag which should be created
+	 */
+	@AliasFor("value")
+	String key() default "";
+
+	/**
+	 * Execute this SPEL expression to calculate the tag value
+	 */
+	String tagValueExpression() default "";
```
adriancole: is this only used when the default resolver is?

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/annotation/SleuthSpanCreatorAspectTests.java
```diff
@@ -0,0 +1,345 @@
...
+		this.testBean.testMethod3();
+
+		List<Span> spans = new ArrayList<>(this.accumulator.getSpans());
+		then(new ListOfSpans(spans)).hasSize(1).hasASpanWithName("custom-name-on-test-method3");
+		then(ExceptionUtils.getLastException()).isNull();
+	}
+	
+	@Test
+	public void shouldCreateSpanWithCustomNameWhenAnnotationOnInterfaceMethod() {
+		this.testBean.testMethod4();
```
adriancole: is it possible to declare method local classes? that might make what the tests are proving easier to see without navigation.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAnnotationAutoConfiguration.java
marcingrzejszczak: Over here - https://github.com/spring-cloud/spring-cloud-sleuth/pull/526/files#diff-4dee63ad879a9394168dd3c9c3a36b6dR42 . By adding `SleuthAnnotationProperties` class and enabling this with the `EnableConfigProps` annotation the boot annotation processor parses all of these classes and creates a JSON that the IDEs can understand.

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
marcingrzejszczak: updated

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
marcingrzejszczak: `./mvnw clean install -Pdocs -pl docs`

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/DefaultSpanCreator.java
marcingrzejszczak: What's wrong with this one? @Koizumi85 was the author (I've polished this class)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SpanTag.java
gavlyukovskiy: Look at this one https://github.com/spring-cloud/spring-cloud-sleuth/blob/4246207844e80dc324c84184de1052da0c4faa67/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAnnotationProperties.java
`@ConfigurationProperties("spring.sleuth.annotation")` does the magic ;)

gavlyukovskiy: Look at this one https://github.com/spring-cloud/spring-cloud-sleuth/blob/4246207844e80dc324c84184de1052da0c4faa67/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAnnotationProperties.java
`@ConfigurationProperties("spring.sleuth.annotation")` does the magic ;)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAnnotationAutoConfiguration.java
gavlyukovskiy: Look at this one https://github.com/spring-cloud/spring-cloud-sleuth/blob/4246207844e80dc324c84184de1052da0c4faa67/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAnnotationProperties.java
`@ConfigurationProperties("spring.sleuth.annotation")` does the magic ;)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SpanTag.java
marcingrzejszczak: It's written in the docs but I'll update the javadocs here.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/annotation/SleuthSpanCreatorAspectTests.java
marcingrzejszczak: No - they need to be beans so I don't this would work.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/NewSpan.java
marcingrzejszczak: Done

marcingrzejszczak:
> Overall, nice job. For release notes, might want to note what was impossible before this change. Here's a couple that come to mind.

updated the docs / PR notes on this

marcingrzejszczak:
> There was a question about eventually making this pluggable (ex request for an api/annotation jar), so the only concern I might have is about the assumption of SPEL support which has a lot of assumptions in implementation.

We could allow a combination of a bean name + expression. Then the `SleuthTagValueResolver` could have another method for evaluating expressions. The default implementation would be SPEL but you could easily override it.

marcingrzejszczak:
I've polished the names and updated the docs as discussed on Gitter. I guess it looks much nicer now ;)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/ContinueSpan.java
```diff
@@ -0,0 +1,42 @@
...
+@Retention(RetentionPolicy.RUNTIME)
+@Inherited
+@Target(value = { ElementType.METHOD, ElementType.TYPE })
+public @interface ContinueSpan {
+
+	/**
+	 * The value passed to the annotation will be used and the framework
+	 * will create two events with the {@code .start} and {@code .end} suffixes
+	 */
+	String log() default "";
```
adriancole: hmm.. I kindof prefer start/end, as that allows you to emulate existing stuff. ex in finagle there are 3 parameters to a trace filter

https://github.com/twitter/finagle/blob/develop/finagle-core/src/main/scala/com/twitter/finagle/tracing/TraceInitializerFilter.scala#L132
```
before: Annotation <- added at the beginning of a task
after: Annotation <- added unconditionally after a task
afterFailure: Annotation + ": $error.class: $error.message" <- additionally added when the task failed
```

marcingrzejszczak: so you're suggesting to add afterFailure too ? or do you want to also rename the event names?

adriancole:
basically there should be a side-effect on the span when there's an
exception thrown. In finagle, it is an annotation not a tag (as a nested
failure might not be critical to the span)

marcingrzejszczak:
It's merged! Congratulations to the whole community for discussions, PRs and eager participation in the final design of the solution. There's nothing more rewarding to us library developers than having such a fabulous community!!! Kudos to @Koizumi85 @gavlyukovskiy @devinsba @adriancole !

Koizumi85:
@marcingrzejszczak Question: My name is mentioned in all the javadoc comments. Should I sign the CLA then?

PR #567, opened 2017-04-18
Span Adjuster

### LINKED ISSUE #553: Add ability to adjust spans before reporting them
Adding a functional interface to adjust the Zipkin's `SpanBuilder` before the Span is built and reported could allow to highly customize the built span according to one's needs without adding anything to the architecture.

For example for this issue https://github.com/spring-cloud/spring-cloud-sleuth/issues/500#issuecomment-290980875 one could apply a customizer to verify the the tags to retrieve the URL and then rename the span accordingly.

Similar to this: https://github.com/openzipkin/zipkin-sparkstreaming#adjuster

cc @adriancole @ouaibsky @s-spindler

marcingrzejszczak:
with this change we're adding a possibility to fully customize the span before it gets reported to e.g. Zipkin 

fixes #553

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/SpanAdjuster.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/SpanAdjuster.java
@@ -1,0 +1,31 @@
+/*
+ * Copyright 2013-2016 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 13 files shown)


marcingrzejszczak:
Obiously docs need to be added too

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/SpanAdjuster.java
```diff
@@ -0,0 +1,31 @@
...
+/**
+ * Adds ability to adjust a span before reporting it.
+ *
+ * @author Marcin Grzejszczak
+ * @since 1.1.4
+ */
+public interface SpanAdjuster {
+	/**
+	 * You can adjust the {@link Span} by creating a new one using the {@link Span.SpanBuilder}
+	 * before reporting it. That way you can e.g. change the name of the span.
```
adriancole: make a code example javadoc of the same.

ex s -> span.toBuilder().name(scrub(span.name)).build();

PR #572, opened 2017-04-22
Synchronizes on zipkin which uses latest Spring Boot 1.4

adriancole:
This updates to a Zipkin version synchronized with out spring boot
config.


adriancole:
will be slightly different for master, as that's on spring boot 1.5

PR #590, opened 2017-05-09
Added the ErrorParser interface

### LINKED ISSUE #576: Attaching the Exception/Throwable to the Span
I wanted to report the `Span`s to logs instead of Zipkin, see: #575.
I would like to report more details about the Exception than the [current error message in the tags](
https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java#L95).

For this, it would be great if the `Span` could contain the Exception/Throwable, not just the error message as a tag.

marcingrzejszczak:
without this change there's no way to alter the exception message that gets attached as a tag to the span
with this change you can register your own ErrorParser bean to alter that behaviour

fixes #576

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/ErrorParser.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/ErrorParser.java
@@ -1,8 +1,9 @@
 package org.springframework.cloud.sleuth;
 
 /**
- * Contract for parsing errors. For the given exception will produce a message
- * that will be then added as a {@link Span#SPAN_ERROR_TAG_NAME} tag to the Span.
+ * Contract for hooking into process of adding error response tags.
+ * This interface is only called when an exception is thrown upon receiving a response.
+ * (e.g. a response of 500 may not be an exception).
  *
  * @author Marcin Grzejszczak
  * @since 1.2.1
@@ -10,7 +11,10 @@ package org.springframework.cloud.sleuth;
 public interface ErrorParser {
 
 	/**
-	 * Returns a {@link String} representation of a {@link Throwable}
+	 * Allows setting of tags when an exception was thrown when the response was received.
+	 *
+	 * @param span - current span in context
+	 * @param error - error that was thrown upon receiving a response
 	 */
-	String parseError(Throwable error);
+	void parseErrorTags(Span span, Throwable error);
 }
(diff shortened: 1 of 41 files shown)


jonatan-ivanov:
How about returning a list of key-value pairs (e.g.: a map) instead of a String in the error parser and add all of them to the tags?
So that if somebody wants to implement a custom `ErrorParser`, they will be able to do not only this:
```
error=custom Something from the exception
```
but this as well:
```
errorClass=ExternalException
error=Hulk sad
```

marcingrzejszczak:
Ok another try... I've changed the `ErrorParser` to 

`void parseErrorTags(Span span, Throwable error);`

That way you can basically do whatever you want with the span and the error. The deafult impl does this:

```
/**
 * {@link ErrorParser} that sets the error tag for an exportable span.
 *
 * @author Marcin Grzejszczak
 * @since 1.2.1
 */
public class ExceptionMessageErrorParser implements ErrorParser {

	private static final org.apache.commons.logging.Log log = LogFactory.getLog(MethodHandles.lookup().lookupClass());

	@Override
	public void parseErrorTags(Span span, Throwable error) {
		if (span.isExportable()) {
			String errorMsg = ExceptionUtils.getExceptionMessage(error);
			if (log.isDebugEnabled()) {
				log.debug("Adding an error tag [" + errorMsg + "] to span " + span);
			}
			span.tag(Span.SPAN_ERROR_TAG_NAME, errorMsg);
		}
	}
}
```

marcingrzejszczak:
If there are no objections I'll merge this some time soon :P

marcingrzejszczak:
Thanks @adriancole . I guess that once someone requests sth similar we can then try to introduce this. Until now I see that mostly people go with the defaults and don't require vast customizations. But that's good to know that Brave has sth like this.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/ErrorParser.java
```diff
@@ -0,0 +1,20 @@
...
+ */
+public interface ErrorParser {
+
+	/**
+	 * Allows setting of tags when an exception was thrown when the response was received.
+	 *
+	 * @param span - current span in context
+	 * @param error - error that was thrown upon receiving a response
+	 */
+	void parseErrorTags(Span span, Throwable error);
```
adriancole: I think mentioned a while back, but Span allows many more things than setting tags. One way to address this is in docs. For example, you could mention explicitly that implementations should only call span.tag and not any other methods. I'm ok on this regardless.

marcingrzejszczak: Ok I'll add the docs and merge this then

marcingrzejszczak:
Done in https://github.com/spring-cloud/spring-cloud-sleuth/commit/f834e5222b19f24e6421bcfd52d46e9c810f0530

PR #622, opened 2017-06-23
Add missing implementation of shutdown method on LazyTraceThreadPoolTaskExecutor

tai6651:
If we call shutdown method on LazyTraceThreadPoolTaskExecutor instance, It will invoke shutdown from ExecutorConfigurationSupport class which ExecutorService is null (as we never initialize executor from LazyTraceThreadPoolTaskExecutor  but use delegate instead). This resulting in null pointer exception.

This fix is to override shutdown method to call delegate.shutdown.

ps. I also add missing override annotation on destroy method :)


marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-sleuth/commit/c902680244ca79ddd743f3474ea026e02bff8df0 (I needed to apply this to 1.2.x then merge 1.2.x to master and then merge master to 2.0.x)

PR #624, opened 2017-06-26
Edgware changes - removed deprecated API usage

marcingrzejszczak:
Also I've added `SpanBuilder.buildContinuedSpan()` - the function differs from `build()` in such a way that it does not copy tags / logs. It's allowing their reusage.


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
```diff
@@ -191,21 +194,13 @@ public Span(Span current, Span savedSpan) {
 		this.savedSpan = savedSpan;
 	}
 
-	/**
-	 * @deprecated please use {@link SpanBuilder}
-	 */
-	@Deprecated
-	public Span(long begin, long end, String name, long traceId, List<Long> parents,
```
marcingrzejszczak: yup - that's why it's for `1.3.0` release.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/trace/DefaultTracer.java
```diff
@@ -242,7 +228,7 @@ private Span createContinuedSpan(Span span, Span saved) {
 		if (saved == null && span.getSavedSpan() != null) {
 			saved = span.getSavedSpan();
 		}
-		return new Span(span, saved);
```
adriancole: so looking at the code and toying around, I think the best is to continue use of the old constructor here and not add a new public method of buildContinuedSpan. In sleuth 2, we will almost surely change this part of the code, so it is helpful to avoid having to deprecate another thing between now and then

marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-sleuth/commit/2f6a8df5fe620527720951b925d7ca3570517604

PR #629, opened 2017-07-06
Updates to latest zipkin stuff

adriancole:
* zipkin core library has Endpoint.Builder.parseIp now
* reporter being 1.0 is just announcing we won't break compat


adriancole:
@marcingrzejszczak you were just toying with Endpoint recently, the new parse thing is nice because it can safely parse text form of IP address and also extract the correct IP from InetAddress https://github.com/openzipkin/zipkin/pull/1580

PR #698, opened 2017-09-14
Introduced a "shared" field in Span

### LINKED ISSUE #696: Add Span.shared
Using B3 propagation, it is most typical to share the same span ID across client and the server. This has backend implications like who owns the timestamp (hint the client does). When a SpanReporter receives a completed span, it should know if it is shared or not.

* The only one to set shared flag is a server, as the client will never know if the span ID will be shared (for example, some servers always fork a child)
* The easiest place to set this is during span extraction (ex when a span is taken from headers, set the shared bit)
* If you have a shared span ID, but immediately fork a child, make sure that child doesn't have shared bit set (as it is exclusive at this point)

Here are some details: http://zipkin.io/zipkin-api/#/default/post_spans

When you are reporting to zipkin, just check span.shared, then don't report timestamp or duration because you don't own it. For example, something like this:

```java
    // don't report server-side timestamp on shared server spans
    if (Boolean.TRUE.equals(in.shared()) && sr != null) {
      result.timestamp(null).duration(null);
    }
```

marcingrzejszczak:
whenever
- a span or trace id get generated when a request / message arrives
- child span is created

we set the field to false

whenever we find ids in the incoming request / message we set the field to true

fixes #696

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
@@ -512,6 +512,9 @@ public class Span implements SpanContext {
 
 	/**
 	 * Span and trace id got extracted from a carrier?
+	 * We are adding data to the same span created by a remote client2
+	 *
+	 * @since 1.3.0
 	 */
 	public boolean isShared() {
 		return this.shared;
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
@@ -16,12 +16,6 @@
 
 package org.springframework.cloud.sleuth;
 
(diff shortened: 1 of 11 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/Span.java
```diff
@@ -500,6 +511,13 @@ public boolean isExportable() {
 	}
 
 	/**
+	 * Span and trace id got extracted from a carrier?
```
adriancole: We are adding data to the same span created by a remote client

devinsba:
@adriancole what is the rationale for dropping the duration from the server span? I think this is valuable information to throw away. Timestamp makes sense to me though

adriancole:
it is actually a historical change around duration ranking in shared spans

https://github.com/openzipkin/openzipkin.github.io/issues/49

sleuth already drops it, but this is using a more specific flag to handle
the signal. as soon as this is in, we'll add zipkin v2 support which can
retain the server timestamp at least in v2 storage systems like ES where we
can filter properly

Issue #695, opened 2017-09-13
Sleuth and Spring JMS


Diagoras:
Hey,

I noticed that Sleuth doesn't seem to automatically work with @JmsListener annotated methods. [A previous issue](https://github.com/spring-cloud/spring-cloud-sleuth/issues/151) along these lines was closed because Spring Cloud Stream should be used instead. However, Spring Cloud Stream doesn't appear to support JMS yet, and I haven't been able to find a clear timeline for when it will.

Is the advice still to move to Spring Cloud Stream, with no plans for supporting the annotation-driven JMS integration that currently exists? If so, is there a particular reason why Spring Sleuth isn't supporting that transport? And do you forsee any difficulties with a third party providing support instead, with some kind of spring-boot-sleuth-jms-starter autoconfiguration that wraps JMS producers and listeners in spans and passes the span through the JMS headers?

Thanks for your time.

shakuzen:
Have you seen this project: https://github.com/spring-cloud/spring-cloud-stream-binder-jms? I don't see any recent commit activity, unfortunately, and there is a warning about it being considered "alpha". Also from the issue you linked, there was a Framework issue opened that needs to be fixed before JMS can be instrumented "directly": [SPR-14683](https://jira.spring.io/browse/SPR-14683).

Diagoras:
> Have you seen this project: https://github.com/spring-cloud/spring-cloud-stream-binder-jms? I don't see any recent commit activity, unfortunately, and there is a warning about it being considered "alpha".

Yeah, it doesn't really seem to be an actively developed binder. The Spring Cloud Stream README also says that only the Kafka and RabbitMQ binders are supported and actively worked on, which is why I was surprised that "use Spring Cloud Stream" was the standing advice for those wanting to propagate spans over JMS.

> Also from the issue you linked, there was a Framework issue opened that needs to be fixed before JMS can be instrumented "directly": SPR-14683.

Ah, I see. That's unfortunate. If that's the reason why JMS is not currently supported, can we leave this issue open and linked to that framework problem to make clear what's blocking progress and what needs to change before JMS can be directly instrumented? That might prevent future users from making the mistake I did, since it's clear that it's a technical issue preventing this change instead of JMS integration being judged out of scope.

[...]

PR #717, opened 2017-09-28
Add instrumenting for jms

Stefku:
Fixes gh-695


marcingrzejszczak:
I closed the PR against `1.2.x`. 

Will this functionality even work? I mean it creates a new span each time a `@JmsListener` method is called. But you're not retrieving any tracing information from the message itself. What if I received a message that already contains span information?

Also I don't think we want an explicit dependency on `activemq`. `Activemq` can be there in the tests but we'd need something more generic in optional compile scope.

Stefku:
Thanks for the review.

Yes, you are right. In our case, the calling system never transports span information thus we always have to create a new span when @JmsListener method is called. It works for our case but I see that's not the general case. I have to think about a more generic way to solve both use cases.

adriancole:
I think a future change would work a bit differently, creating a messaging span. Which version is this going for?

marcingrzejszczak:
A more generic solution is related to this issue https://jira.spring.io/browse/SPR-14683. Until that one is fixed we can't do much about it. I'll close this one

PR #720, opened 2017-10-03
Adds Zipkin v2 support

adriancole:
This adds the ability to use zipkin v2 endpoints natively. It also
allows wiring of standard zipkin reporters and senders.


spencergibb:
Should this be targeted at the 2.0.x branch?

adriancole:
@spencergibb well also 2.0, unless we aren't doing new features in 1.3. Would be nice to give a way for folks to switch to the new format and use senders even if they aren't on spring boot 2.0 yet.

adriancole:
can re-target to 2.0 and backport to 1.3 if that makes better sense.. lemme know!

adriancole:
added auto-config tests, specifically showing how to use the v1 endpoint.

Note: there's a deprecation warning on this:
```
/Users/acole/oss/spring-cloud-sleuth/spring-cloud-sleuth-zipkin2/src/main/java/org/springframework/cloud/sleuth/zipkin2/DiscoveryClientEndpointLocator.java:[53,55] getLocalServiceInstance() in org.springframework.cloud.client.discovery.DiscoveryClient has been deprecated
```
what should we do instead? cycle through the service endpoints and pick one?

adriancole:
added a commit about the deprecated thing, moving to what it said (which doesn't have host/port info, but maybe that's less important?)

adriancole:
fyi added 1.2.x backport https://github.com/spring-cloud/spring-cloud-sleuth/pull/721

Issue #459, opened 2016-12-01
Support AWS X-Ray


jramsdale:
Much like Zipkin, Amazon now provides AWS X-Ray, their hosted tracing service: https://aws.amazon.com/xray/

Adding X-Ray support to Spring Cloud Sleuth would be a nice addition for users preferring to use AWS managed services.

adriancole:
Here are some links of interest

description of data that can be sent: http://docs.aws.amazon.com/xray/latest/devguide/xray-api.html#xray-api-segments
the Api model: https://github.com/aws/aws-sdk-java/blob/master/aws-java-sdk-models/src/main/resources/models/xray-2016-04-12-model.json
doesn't seem the servlet filter mentioned in [docs](http://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-java.html) is on github https://github.com/search?utf8=%E2%9C%93&q=AWSXRayServletFilter
here's an example of using their sdk http://docs.aws.amazon.com/xray/latest/devguide/xray-sdk-java-subsegments.html

Here's some notes @ryangardner toook at reinvent 

>A few tidbits from talking to the amazon people about their implementation. One reason they went with their own stuff was because they felt they needed to put the timing stuff in the header to make it easier to scale it up
They shuttle the data over UDP to a daemon that runs on localhost and that daemon sends it to the collector. It’s a well-known endpoint that everyone in AWS will share for a given region. The data you see is based on the account you send the data in from. If you want to trace data across multiple aws accounts you have to use a cross-account linking and assuming the role of the central “x-ray” account for your org. In the future they may be able to make this easier by leveraging metadata made available by the “orgs” feature that amazon just announced

adriancole:
so I think next step would be either get a hold of the instrumentation SDK code and/or a spec for how headers propagate, etc. Particularly, if this is related to and/or compatible with the [recent announcement on ALB](https://aws.amazon.com/blogs/aws/application-performance-percentiles-and-request-tracing-for-aws-application-load-balancer/), which seems to imply "cookie crub" trace id header. Ex ids are tacked on.

nicmunroe:
I took a quick look at this yesterday for Wingtips. The docs are sort of broken with how to pull in libs, where classes are, etc. But I did find the `AWSXRayServletFilter` class. It's in this dependency: `com.amazonaws:aws-xray-recorder-sdk-core:1.0.0-beta`. The source jar had no java code in it though, so it's basically fun with decompiled code.

[...]

PR #724, opened 2017-10-05
Encodes epoch seconds into first 32bits of a 128-bit trace ID

adriancole:
Amazon will throw out trace IDs that aren't associated with a recent
timestamp. This encodes the current epoch seconds into the first 32
of a 128-bit trace ID to support conversion to an Amazon Root ID.

See #459


adriancole:
screen shot 2017-10-05 at 9 35 50 pm

This proves that amazon X-Ray accepts the IDs

Issue #395, opened 2016-09-07
Add integration support for Google Stackdriver Trace


nyilmaz:
Referencing https://cloud.google.com/trace/api/?hl=en_US,

Would you consider adding pushing trace-span data to Google's Stackdriver trace? They have a Java API [here](https://github.com/GoogleCloudPlatform/cloud-trace-java), but as far as I experienced samples have dependency issues with grpc. Since spring have [oauth2](http://projects.spring.io/spring-security-oauth/docs/oauth2.html) support, REST API may be used as a beginning.

What do you think?

adriancole:
@nyilmaz we're going to move our transport layer to https://github.com/openzipkin/zipkin-reporter-java in the near future. This will include a Sender component which by default will go to zipkin. A sender that does oauth etc will make sense when going to stackdriver, although a zipkin-> stackdriver proxy is also an option.

While you can pad out in the mean time, we need to address 128 bit trace ids (not span ids, just trace ids) for this to work well. Please upvote this change, which I'll be working on shortly.

adriancole:
uhh this change: https://github.com/openzipkin/zipkin/issues/1262

cemo:
@adriancole Is there any progress on this issue? I believe that this is a quick win for `zipkin`. Introducing `zipkin` features into a library without using a backend sounds really cool.

mtwo:
Hey Cemo, I'm the PM for Stackdriver Trace @ Google! We're actually working on something that'll achieve this, which we'll be announcing soon. I'll reply back when we're able to show more to a broader audience.

cemo:
Hey @mtwo! We will be looking forward your news. Thanks for the heads up.

mtwo:
@cemo check [this](https://github.com/GoogleCloudPlatform/stackdriver-zipkin) out. It's a Zipkin collector implementation that forwards trace data to Stackdriver Trace. We'll be publishing the official announcement shortly, but feel free to try it out and let myself and @kevinmdavis and I know what you think.

nyilmaz:
@mtwo now i'm trying `stackdriver-zipkin` adapter but i could not find how to set `ZipkinStackdriverStorageProperties#apiHost`.

kevinmdavis:
[...]

PR #725, opened 2017-10-06
Adds spring.sleuth.supports-join to disable span ID sharing

adriancole:
This adds the ability to opt out of span sharing between the client and
server side of an RPC. This is important when reporting to systems that
do not share span IDs, such as Google Stackdriver and Amazon X-Ray.

Note: this still defaults to true, in favor of existing Zipkin
implementations. See https://github.com/openzipkin/zipkin/issues/1480

See #395
See #459


adriancole:
Tested with amazon

screen shot 2017-10-06 at 1 48 19 pm

adriancole:
got virtual approval from marcin (or someone who hacked his gitter, which works for me)

PR #738, opened 2017-10-13
Adds auto-configuration for RabbitMQ and Kafka transports

adriancole:
cc @shakuzen @llinder I still need to figure out how to get the autoconfig to stop matching in unit tests (ex rabbit is configuring during the discovery (http) test)

### DIFF
--- a/spring-cloud-sleuth-dependencies/pom.xml
+++ b/spring-cloud-sleuth-dependencies/pom.xml
@@ -14,9 +14,9 @@
 	<name>spring-cloud-sleuth-dependencies</name>
 	<description>Spring Cloud Sleuth Dependencies</description>
 	<properties>
-		<zipkin.version>2.2.0</zipkin.version>
-		<zipkin-reporter.version>1.1.1</zipkin-reporter.version>
-		<zipkin-reporter2.version>2.1.1</zipkin-reporter2.version>
+		<zipkin.version>2.2.1</zipkin.version>
+		<zipkin-reporter.version>1.1.2</zipkin-reporter.version>
+		<zipkin-reporter2.version>2.1.3</zipkin-reporter2.version>
 	</properties>
 	<dependencyManagement>
 		<dependencies>
@@ -104,6 +104,30 @@
 				<groupId>io.zipkin.reporter2</groupId>
 				<artifactId>zipkin-reporter</artifactId>
 				<version>${zipkin-reporter2.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>io.zipkin.reporter2</groupId>
+				<artifactId>zipkin-sender-kafka11</artifactId>
+				<version>${zipkin-reporter2.version}</version>
+				<exclusions>
+					<!-- assigned with spring-kafka -->
+					<exclusion>
+						<groupId>org.apache.kafka</groupId>
+						<artifactId>kafka-clients</artifactId>
+					</exclusion>
+				</exclusions>
+			</dependency>
+			<dependency>
+				<groupId>io.zipkin.reporter2</groupId>
+				<artifactId>zipkin-sender-amqp-client</artifactId>
+				<version>${zipkin-reporter2.version}</version>
+				<exclusions>
+					<!-- assigned with spring-rabbit -->
+					<exclusion>
+						<groupId>com.rabbitmq</groupId>
+						<artifactId>amqp-client</artifactId>
+					</exclusion>
+				</exclusions>
 			</dependency>
 		</dependencies>
 	</dependencyManagement>
--- a/spring-cloud-sleuth-samples/pom.xml
+++ b/spring-cloud-sleuth-samples/pom.xml
@@ -59,17 +59,17 @@
 			<dependency>
 				<groupId>io.zipkin.java</groupId>
 				<artifactId>zipkin</artifactId>
-				<version>2.2.0</version>
+				<version>2.2.1</version>
 			</dependency>
 			<dependency>
 				<groupId>io.zipkin.zipkin2</groupId>
 				<artifactId>zipkin</artifactId>
-				<version>2.2.0</version>
+				<version>2.2.1</version>
 			</dependency>
 			<dependency>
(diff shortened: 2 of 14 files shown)


adriancole:
tested kafka works in the zipkin2 branch of the sleuth demo by placing this in the classpath:

```diff
diff --git a/pom.xml b/pom.xml
index 7f2e6d5..742aefd 100644
--- a/pom.xml
+++ b/pom.xml
@@ -46,6 +46,15 @@
       <groupId>org.springframework.cloud</groupId>
       <artifactId>spring-cloud-sleuth-zipkin2</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.zipkin.reporter2</groupId>
+      <artifactId>zipkin-sender-kafka11</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.springframework.kafka</groupId>
+      <artifactId>spring-kafka</artifactId>
+      <version>2.0.0.RELEASE</version>
+    </dependency>
   </dependencies>
```

adriancole:
rabbit took less, maybe we already have an implicit dep on rabbit in sleuth which could mean I need another approach to detect if the user really wants it..

```diff
diff --git a/pom.xml b/pom.xml
index 7f2e6d5..f9e7222 100644
--- a/pom.xml
+++ b/pom.xml
@@ -46,6 +46,10 @@
       <groupId>org.springframework.cloud</groupId>
       <artifactId>spring-cloud-sleuth-zipkin2</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.zipkin.reporter2</groupId>
+      <artifactId>zipkin-sender-amqp-client</artifactId>
+    </dependency>
   </dependencies>
 
   <repositories>
```

adriancole:
corrected my comment about rabbit. it was my mistake.. it works sensibly

adriancole:
ok things are working... going to make auto-config tests now

adriancole:
Ok code is ready, we just need docs somewhere.. I'll add a README I guess to show which dependencies are needed.

adriancole:
I have an idea.. what if I put a hard dependency on zipkin's senders for rabbit and kafka, excluding transitives. This means we can rely on users just adding normal spring-kafka or spring-rabbit deps. The tradeoff is a couple jars which are couple kilobytes each.

If I am not mistaken this means that a user would add the scs-zipkin2 dep and then spring-rabbit (via start.spring.io etc) and be on their way.

Does this sound sensible? Cc @joshlong @dsyer

shakuzen:
If the sender dependencies are required, we should have some opinion on which sender wins in case all conditions are met. Maybe an application uses Kafka and RabbitMQ regardless of Zipkin reporting - which way are the spans sent by default? Below are some solutions for this.

**Mandatory user selection**
Spring Boot deals with this situation [using a selector for Spring Session's storage](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-session) where users must explicitly set `spring.session.store-type`.

**Opinionated selection**
On the other hand, for caching support, Spring Boot will select a provider based on what's available on the classpath and [a priority defined here](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#_supported_cache_providers). There is still an option to force the selection using `spring.cache.type`.

**Third option?**
Users could disable the senders they don't want. For example, if they have Kafka and RabbitMQ deps but want to report via HTTP they would have to set `spring.zipkin.rabbitmq.enabled` and `spring.zipkin.kafka.enabled` to `false`. I still think this leaves the default behavior perhaps undefined if Kafka and RabbitMQ dependencies are present - or will the RabbitMQ sender win because its configuration class is first in the source?

adriancole:
@shakuzen very good review of the options! I guess one more would be a scs-zipkin-rabbit module, which contains the correct version of the sender dep and also a dep on spring-rabbit (as well the autoconfig file). If someone included this, they'd surely be wanting zipkin included.

Another option is to implement autoconfig in zipkin, where zipkin publishes sender autoconfiguration. Only issue I can think of is that it might accidentally skew the spring-rabbit|kafka versions defined in sleuth or boot, especially across different versions of boot (ex if zipkin-reporter says spring-kafka 2.0, whereas some version of sleuth does not. OTOH, this is 
 a fairly normal issue, and we don't use a lot of apis anyway. An advantage of zipkin-native autoconfig is that a sleuth 3.0 could use that directly and never have to publish anything between now and then...

https://github.com/openzipkin/zipkin-reporter-java/issues/71

adriancole:
Ps I like the opinionated selection type. If no one has overridden the http endpoint, prefer rabbit then kafka.

I can spike an impl of this. I like it because it is easy for users which means less likely to result in problematic classpath setup.

@shakuzen nice job!

So once I test this, we can still debate between now and 1.3, but seems cool as for example start.spring.io would say.. add zipkin2 and add rabbit if you want to send there

adriancole:
OK all working. tested with sleuth-webmvc example cc @shakuzen @ImFlog

* with just "org.springframework.cloud:spring-cloud-sleuth-zipkin2" reports to http at v2 endpoint
* adding "org.springframework.amqp:spring-rabbit" reports to rabbit queue named zipkin
* adding "org.springframework.kafka:spring-kafka" reports to kafka topic named zipkin

When both rabbit and kafka are present, rabbit wins. You can hard-code to ensure you get what you want: "spring.zipkin.sender.type" to "rabbit" "kafka" or "web" (rest template).

Lemme know if you are ok with the names.

I took conventions from the Cache autoconfig logic, but squashed things down as it was too much.

adriancole:
weird.. I force pushed this branch, but the changes aren't showing.. I made a copy of the branch here https://github.com/spring-cloud/spring-cloud-sleuth/compare/messaging-transports2

adriancole:
This branch is messed up somehow. I will cherry-pick in https://github.com/spring-cloud/spring-cloud-sleuth/compare/messaging-transports2 Happy for feedback post-commit between now and 1.3.0 final

adriancole:
closed via 704a18fe2457fc907ac93c3575f0066196eabe4b

PR #740, opened 2017-10-13
Filter values having Zipkin compatible keys

kislitsyn:
Optimization linked with necessity to exclude unrequired values from the temporary carrier of header values

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/HttpServletRequestTextMap.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/HttpServletRequestTextMap.java
@@ -16,11 +16,12 @@
 
 package org.springframework.cloud.sleuth.instrument.web;
 
-import javax.servlet.http.HttpServletRequest;
+import java.util.AbstractMap;
 import java.util.Enumeration;
-import java.util.HashMap;
 import java.util.Iterator;
 import java.util.Map;
+
+import javax.servlet.http.HttpServletRequest;
 
 import org.springframework.cloud.sleuth.SpanTextMap;
 import org.springframework.web.util.UrlPathHelper;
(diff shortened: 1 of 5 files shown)


marcingrzejszczak:
HI! Thanks a lot for the PR. Can you elaborate more on the idea behind this change? Is there some issue that this PR fixes?

kislitsyn:
We use Sleuth for tracing processes in microservices. Each receiving hundreds and thousands requests per second. I can't control what information will be sent in the headers. Want to be sure that Sleuth working only with minimally necessary part of data from headers.
The reason for creating an PR is warning of possible problems on the high load

adriancole:
fyi the name of this change could be a bit misleading.. there are few non-standard things sleuth propagates ex. PROCESS_ID_NAME, SPAN_NAME_NAME,  URI_HEADER and then the "baggage" prefix is not B3 compatible (only works when the other side is also sleuth)

adriancole:
note: if not for the baggage prefix, it would be a fixed count of potential headers

kislitsyn:
@adriancole `ZipkinHttpSpanMapper.SPAN_FIELDS` contains only headers that using in `ZipkinHttpSpanInjector` and `ZipkinHttpSpanExtractor`. The PR about headers compatible with this Zipkin extractor and injector. Thats why I named this changes connected with Zipkin

marcingrzejszczak:
Thanks a lot for the PR and explanation! Great work!

michael-barker:
This pull request made it so that `HttpServletRequestTextMap::put` throws an `UnsupportedOperationException`.  Is this intentional? This broke a custom `SpanExtractor` of ours.

Our use case is that our custom `SpanExtractor` delegates to `ZipkinHttpSpanExtractor` but we add a couple things to the `SpanTextMap` before delegating to the Zipkin extractor.

Did `HttpServletRequestTextMap` need to change to accomplish what this pull request was intended to do?

kislitsyn:
As the javadoc on `SpanTextMap::put` says - `Puts a key:value pair into the TextMapWriter's backing store.`. 
`HttpServletRequestTextMap`should only read the attributes from request and implement `TextMapReader` functionality.

You can solve your problem in two ways:
1. delegate `SpanTextMap` in your custom `SpanExtractor::joinTrace`
2. get span by `delegate.joinTrace`, and use this span to build your own custom span by `SpanBuilder.from(span)`

I would prefer the second way, because it is more obvious

michael-barker:
The 2nd option doesn't work because I want to generate a trace ID if one wasn't provided.  `ZipkinHttpSpanExtractor` returns null if the trace ID is missing so the `SpanTextMap` already has to have a trace ID when I call `ZipkinHttpSpanExtractor::joinTrace`.

My current solution is to implement `SpanTextMap` with a mutable backing store and create that from the `SpanTextMap` passed to `SpanExtractor::joinTrace`.  I can then pass that to `ZipkinHttpSpanExtractor::joinTrace`.

kislitsyn:
Michael, it looks like you try to create new trace by `traceFilter`. This filter create new span with new traceId if request doesn't contains trace metadata, you don't need to generate traceId.

In case you need to generate new span with some custom properties from request, you can write you custom `GenericFilterBean`. In wich to implement the creation of child span or continue existing with these properties and place it after `traceFilter`

PR #756, opened 2017-10-20
Removes spring-cloud-sleuth-zipkin-stream

### LINKED ISSUE #727: Remove the Zipkin Stream server
Zipkin server is basically incompatible with Boot 2.0, we have to hack it around to make it work. Also, Zipkin's getting more support in terms of messaging reporters. What we could do is to tell the users to use the `1.5.x` version of the server and cooperate with the Zipkin project to add support for Kafka and a bridge to the Sleuth version of a span.

WDYT @spencergibb @ryanjbaxter @adriancole @dsyer @devinsba ?

adriancole:
Fixes #727


marcingrzejszczak:
Maybe we should update the docs with the info that we did this?

PR #762, opened 2017-10-28
Ensures there's a sampler when properties are present

adriancole:


adriancole:
don't merge this as making it work for master will make it break on 2.0.x due to the scope thing

```
java.lang.IllegalStateException: No Scope registered for scope name 'refresh'
	at org.springframework.cloud.sleuth.zipkin2.ZipkinAutoConfigurationTests.propertyConfiguredSampler(ZipkinAutoConfigurationTests.java:93)
```

Issue #711, opened 2017-09-25
Add support for Brave autoconfiguration


marcingrzejszczak:
If Sleuth integrates with Brave then we would start supporting quite a few options that currently Sleuth doesn't support. 

We may open issues for some stickier points:
- [x] Provide migration path for name prefixed baggage (possibly by whitelisting names)
- [ ] Make a JMS propagation strategy (ex fallback to check underscore convention; inject with underscores)
- [x] hash out how to deal with naming conventions (ex http:/foo span namer)
  
The migration guide is drafted here - https://gist.github.com/marcingrzejszczak/d3c15a0c11dda71970e42c513c9c0e09

shakuzen:
Would the semantics of this be: if Brave is on the classpath, use it as the tracer implementation; if not, use the existing homegrown Sleuth tracer?

adriancole:
right now, folks like @devinsba @jonathan-lo @jorgheymans and @ImFlog have or contribute to brave instrumentation, and this can work for example in classic spring environments or spring boot. Examples are grpc and kafka (not stream). There are also frameworks harder for us to support directly, such as dubbo.

In the case of http, if we had the ability to use brave types, we could obviate a lot of templated patterns, such as http sampling and span data policy.

So, to get highest value, there's probably a question of whether to implement a brave shim (which implies a decent amount of brave to implement it w/o brave), or whether to make sleuth a layer on brave (with the dependency locking implications of that).

There are some shortcuts we could take with a layer, which might reduce the amount of code needed, but it will still largely be copy/paste of code that is automatically fixed in brave.

So, it is a directional decision, something to ponder. Any folks have steering input? cc also @dsyer

jorgheymans:
IMO the layer makes the most sense in the spring-boot philosophy. Make it easy to quickly get up and running doing tracing using brave/zipkin, provide trace-enable *Template classes and annotations etc, And at the same time facilitate a way to access the full power of brave for more advanced scenarios. This is how most spring-boot integrations work, from my perspective at least.

devinsba:
Personally I've used both quite a bit over the last 2 years. I would love if sleuth was a layer on top of brave to enable all of the existing brave instrumentation.

For us, sleuth is an easy first step toward zipkin, we get log correlation and all of the built in goodies without needing to set up a reporter. If this was then layered on top of brave then it would allow those teams using things with brave instrumentation (grpc, dynamodb, etc..) to get these things as well.

[...]

PR #775, opened 2017-11-10
RxJava2 Autoconfiguration

lukaszguz:
Hi guys,
we developed auto configuration to RxJava2 based on RxJavaPlugins.
We have still had small problem because RxJavaPlugins' schedule handler is called when Observable, Single etc. are subscribe and when thread pool is changed.
In the first case we don't have problem but in the second RxJava's scheduler create worker and wrapped Runnable in own package scope wrapper so we don't know which type Runnable is.
The only thing we did is checked type using reflection.
http://reactivex.io/RxJava/2.x/javadoc/io/reactivex/Scheduler.html#scheduleDirect-java.lang.Runnable-


pivotal-issuemaster:
@lukaszguz Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=775)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=775) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@lukaszguz Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=775)!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/rxjava2/SleuthRxJava2SchedulersHandler.java
```diff
@@ -0,0 +1,156 @@
...
+			}
+			Runnable wrappedAction = this.delegate != null ? this.delegate.apply(action)
+					: action;
+			return new SleuthRxJava2SchedulersHandler.TraceAction(this.tracer,
+					this.traceKeys, wrappedAction, this.threadsToSample);
+		}
+
+		private boolean isTraceActionDecoratedByRxWorker(Runnable action) {
+			try {
+				if (action instanceof Disposable) {
```
adriancole: this is likely an expensive check. maybe check w/ some rx maintainers to see if there's a better way?

adriancole:
No problem.. awkward reflection is better than no progress. Keep us posted!

lukaszguz:
I agree with you :) 
I asked on RxChannel and they got me very fast answer.
https://github.com/ReactiveX/RxJava/issues/5733

lukaszguz:
I've opened PR in Rx
https://github.com/ReactiveX/RxJava/pull/5734

adriancole:
>
> I've opened PR in Rx
> ReactiveX/RxJava#5734 <https://github.com/ReactiveX/RxJava/pull/5734>
>
hero status is on the way! I love it when things come together. Much
appreciated

lukaszguz:
Hi again :)
RxJava2 had 2.1.8 release with all things needed to remove reflection from `ScheduleHandler`.

marcingrzejszczak:
Great! BTW we've just started doing this - https://github.com/spring-cloud/spring-cloud-sleuth/issues/711 . That will mean that most of your work will be obsolete :/ We have to do this big refactoring in 2.0.x cause it's the last chance for us to break the compatibility. I'm not going to close this PR yet and ask you to redo it once we're done cause maybe I'll merge it later. I just wanted to keep you posted on this.

lukaszguz:
"Life is life" :D, no problem I can wait and if you think that it will not be useful to you I have still tool to fix previous problem :)

marcingrzejszczak:
We've merged the Brave instrumentation. Care to redo your work against the new master? :P

lukaszguz:
I can try :D

lukaszguz:
Ok, I think I've rewritten it on Brave implementation.

marcingrzejszczak:
LGTM. @adriancole - maybe this should actually go to Brave? WDYT?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/rxjava2/SleuthRxJava2SchedulersHandler.java
```diff
@@ -0,0 +1,148 @@
...
+					}
+					this.actual.run();
+					return;
+				}
+			}
+
+			Span span = this.parent;
+			boolean created = false;
+			if (span != null) {
+				span = this.tracer.joinSpan(this.parent.context());
```
adriancole: not sure we want to use joinSpan like this. if you have a span, wouldn't you just use it?

adriancole:
on the upstreaming, here's my 2p.

scheduler hook to propagate thread context should be decoupled from the auto-spanning code, as I'm sure the former is reusable, not quite sure the latter would be. Concerns I have are around the span policy and decoupling of things like which thread applies etc.

If starting with just rxjava2 trace propagation, we have a few advantages:
* there's existing need https://github.com/openzipkin/brave/issues/195
* it is smaller feature, so easier to acclimate to the different review and code style culture
* it takes out the polarizing code which helps also with ^^

Dare to give it a go? If so, it would be under brave's repo instrumentation/rxjava2 and start strictly with propagation, covering the scenarios where a user or another tool can create spans that continue from whatever was the ancestor. If you go for it, avoid any dependencies rxjava2 doesn't have, and any features that aren't strictly required for propagation to work. For example, when not auto-spanning, you can safely propagate trace context regardless I think.

If you go that route, open a PR, citiing brave's issue 195 cc'ing people from that issue and also the rxjava person you were able to contact formerly. Once that's settled, we'd talk about auto-spanning things

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/rxjava2/SleuthRxJava2SchedulersHandler.java
lukaszguz: The documentation contains "Joining is re-using the same trace and span ids extracted from an incoming RPC request. This should not be used for messaging operations, as nextSpan(TraceContextOrSamplingFlags) is a better choice." of course it doesn't have to be RPC call but observer or subscriber or whatever call. SleuthRxJavaSchedulersHook.TraceAction also has the same logic (join span).
What do @adriancole think about it?

lukaszguz:
I do not know if I understood correctly. You mean remove RxJava2Autoconfiguration from sleuth-core and add new brave's repo instrumentation/rxjava2 which comprehensively solve tracing in rx2?

adriancole:
I think the key here is that it isn't remote as you literally have a
(local) reference to the span you want. For example you didnt extract the
span from headers, you dont need to double check sampling status etc. Join
is a remote special case for RPC. Definitely we can run out join as
appropriate.

More important than documentation nuance is this.. why are we taking a span
we have and the turning it back into a span. What happens if you were to
use the reference (to what is named parent) directly? I think you can.

There is probably going to be a few places where the api mapping from
sleuth to brave is slightly different. This is probably one of those
things. In brave joinSpan is only for remote. If you have a span locally
just use it. Simple as that. If you want to make a child, what you had
(nextSpan) is correct.

On 22 Jan 2018 9:27 pm, "lukaszguz" <notifications@github.com> wrote:

*@lukaszguz* commented on this pull request.
------------------------------

In spring-cloud-sleuth-core/src/main/java/org/springframework/c
loud/sleuth/instrument/rxjava2/SleuthRxJava2SchedulersHandler.java
<https://github.com/spring-cloud/spring-cloud-sleuth/pull/775#discussion_r162935204>
:

> +			for (String threadToIgnore : this.threadsToIgnore) {
+				if (threadName.matches(threadToIgnore)) {
+					if (log.isTraceEnabled()) {
+						log.trace(String.format(
+								"Thread with name [%s] matches the regex [%s]. A span will
not be created for this Thread.",
+								threadName, threadToIgnore));
+					}
+					this.actual.run();
+					return;
+				}
+			}
+
+			Span span = this.parent;
+			boolean created = false;
+			if (span != null) {
+				span = this.tracer.joinSpan(this.parent.context());

The documentation contains "Joining is re-using the same trace and span ids
extracted from an incoming RPC request. This should not be used for
messaging operations, as nextSpan(TraceContextOrSamplingFlags) is a better
choice." of course it doesn't have to be RPC call but observer or
subscriber or whatever call. SleuthRxJavaSchedulersHook.TraceAction also
has the same logic (join span).
What do @adriancole <https://github.com/adriancole> think about it?

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/spring-cloud/spring-cloud-sleuth/pull/775#discussion_r162935204>,
or mute the thread

adriancole:
Think of sleuth as focused on autoconfiguration. If you look at bravd there
are many instrumentation there some we havent autoconfigured. So apart from
the autoconfiguration part, rebuild rxjava2 in brave in a composable way,
starting with just propagation. This code would have no spring version
binding so can work in or out of spring or any version of spring. Sleuth
would add the smart defaults maybe compose things like thread conventions
etc.

Have a look at existing brave and code here that binds it. Iirc sleuth
reuses brave apach http client stuff so maybe a good thing to look at.

So yes, build comprehensive there except spring part.... and start small..
just propagation as many wont want to make spans for every transition.

lukaszguz:
Ok, I think I've got it. Your explanation about "joins" and "brave" is enough for me :) so I will remove "joins" from autoconfig.
For now I don't know if I will have enough time to brave's development but I will see.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/rxjava2/SleuthRxJava2SchedulersHandler.java
```diff
@@ -0,0 +1,156 @@
...
+			RxJavaPlugins.reset();
+			RxJavaPlugins.setScheduleHandler(
+					new ScheduleHandler(tracer, traceKeys, threadsToSample, delegate));
+		}
+		catch (Exception e) {
+			log.error("Failed to register Sleuth RxJava SchedulersHook", e);
+		}
+	}
+
+	private void logCurrentStateOfRxJavaPlugins() {
```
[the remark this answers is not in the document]
lukaszguz: Changed.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/rxjava2/SleuthRxJava2SchedulersHandler.java
lukaszguz: Join is removed.

lukaszguz:
Could @adriancole look at https://github.com/openzipkin/brave/compare/master...lukaszguz:master ?
And say what you think about it?

adriancole:
Yep will do later today or tomorrow. Shaking off jetlag

lukaszguz:
Sure :) I wish you health ;)

adriancole:
opened https://github.com/openzipkin/brave/issues/664 for a "continue-only" scheduler hook for rxjava2. This would not trace rxjava2, but it could prevent broken traces. Once this is out, we should revisit this (which requires trace propagation, but does more as it is also automatically creating spans and configuring sleuth)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/rxjava2/SleuthRxJava2SchedulersHandler.java
```diff
@@ -0,0 +1,145 @@
...
+	SleuthRxJava2SchedulersHandler(Tracer tracer, TraceKeys traceKeys,
+			List<String> threadsToSample) {
+		try {
+			Function<? super Runnable, ? extends Runnable> delegate = RxJavaPlugins
+					.getScheduleHandler();
+			if (delegate instanceof SleuthRxJava2SchedulersHandler.ScheduleHandler) {
+				return;
+			}
+			logRegisterRxJavaPlugins();
+			RxJavaPlugins.reset();
```
adriancole: just reviewing this closer while pulling out current trace context code.

ps this resets all plugins to null/default.. is this what you wanted to do? What if other plugins besides scheduler were present?

marcingrzejszczak:
Also, I wonder if the current testing suite is enough. Does the instrumentation work with flat maps and more sophisticated operators? Our initial reactor one didn't :|

adriancole:
this is a big job :P

lukaszguz:
Sorry for silent I was on holiday :)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/rxjava2/SleuthRxJava2SchedulersHandler.java
lukaszguz: I little based on plugin to RxJava1 therefore I left this line.

marcingrzejszczak:
What do we do about this? :P

adriancole:
dunno we can put it on the hopper for discussion next week. right now we
have rxjava2 propagation only https://etherpad.net/p/zipkin_fukuoka

On Wed, Aug 1, 2018 at 9:02 PM, Marcin Grzejszczak <notifications@github.com

marcingrzejszczak:
Let me close this for now. I guess we would be adding this in Brave anyways.

PR #781, opened 2017-11-15
The AbstractTraceHttpRequestInterceptor has been enhanced in order to…

team3:
Current implementation of RestTemplate interceptor allows to track the http **request** but doesn't allow to track the http **response**.
The main reason of this pull request is to extend the current implementation so that it will be possible to send the information about http response status and then to visualize it on zipkin.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/HttpTraceKeysInjector.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/HttpTraceKeysInjector.java
@@ -7,6 +7,7 @@ import java.util.Map;
 import org.springframework.cloud.sleuth.Span;
 import org.springframework.cloud.sleuth.TraceKeys;
 import org.springframework.cloud.sleuth.Tracer;
+import org.springframework.http.HttpStatus;
 import org.springframework.util.StringUtils;
 
 /**
@@ -60,6 +61,14 @@ public class HttpTraceKeysInjector {
 			Map<String, ? extends Collection<String>> headers) {
 		addRequestTags(url, host, path, method);
 		addRequestTagsFromHeaders(headers);
+	}
+
+	/**
+	 * Adds tags from the HTTP repos to the given Span
+	 *
+	 */
+	public void addResponseTags(Span span, HttpStatus httpStatus) {
+		tagSpan(span, this.traceKeys.getHttp().getStatusCode(), String.valueOf(httpStatus.value()));
 	}
 
 	/**
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/AbstractTraceHttpRequestInterceptor.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/AbstractTraceHttpRequestInterceptor.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.sleuth.instrument.web.client;
 
+import java.io.IOException;
 import java.lang.invoke.MethodHandles;
 import java.net.URI;
 
@@ -27,6 +28,8 @@ import org.springframework.cloud.sleuth.Tracer;
 import org.springframework.cloud.sleuth.instrument.web.HttpTraceKeysInjector;
 import org.springframework.cloud.sleuth.util.SpanNameUtil;
 import org.springframework.http.HttpRequest;
+import org.springframework.http.client.ClientHttpResponse;
+
 /**
  * Abstraction over classes that interact with Http requests. Allows you
  * to enrich the request headers with trace related information.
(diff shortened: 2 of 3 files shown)


pivotal-issuemaster:
@team3 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=781)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=781) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@team3 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=781)!

team3:
screenshot

marcingrzejszczak:
Can you provide more rationale behind this change? Thanks

team3:
hey @marcingrzejszczak i've updated the original pull request description.

marcingrzejszczak:
You already have this information in Zipkin, coming from the response side, for any status code that is not 200 AFAIR

team3:
It seems that this is applicable only for the general result of the execution.

Having the chain like API gateway -> service A -> service B. API gateway is called directly using an http client. Service A is called from API gateway using RestTemplate.  Service B is called from service A using RestTemplate.
In such scenario zipkin ui shows only the status code for the request from http client to API gateway. I've added the tracing of status codes for internal calls.

marcingrzejszczak:
I don't follow.

> API gateway -> service A -> service B. 

If all of these are using Sleuth then you will get in Service A and Service B the status code.

adriancole:
This seems sensible. In brave, some are adding response status, especially in error status. Some are adding content-length (via custom code as we don't support OOB) https://github.com/openzipkin/brave/blob/master/instrumentation/http/src/main/java/brave/http/HttpParser.java#L41 Note that not sending success by default is intentional, as it consumes storage for what for a lot of people is uninteresting.

I would suggest a test that shows what you are doing.. ex that you end up with a span including http status. Particularly, this can answer Marcin's q about whether you already get error status without this change. My guess is you want to log the positive status code for some reason. Again it sounds reasonable as long as it doesn't default to send success status code as that makes everyone else have to store them by default

marcingrzejszczak:
`TraceFilter` already sets the status code. You can check it here https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java#L410-L424

team3:
Let me just check some details.

team3:
Okay, the pull request works when there is an API-gateway which calls different services which have no sleuth on board (since there are no control on them).

**API-gateway -> Service 1** than **API-gateway -> Service 2**, than **API-gateway -> Service 3**

TraceFilter handles the requests and responses on/from API-gateway. All internal requests perform by RestTemplate  which has an interceptor. The interceptor handles http requests only, hence there are no response codes in zipkin UI.

adriancole:
I think one of the issues making this hard to reason with is that we are guessing based on a low-level change what the high-level result is intended to be. Can you phrase this in terms of what you want to add to the span (especially what's not there now?) Ideally you can post a simple 2-tier example as usually you don't need multiple tiers to demonstrate something.

For example, in simplest form, you could patch something like this, showing what you need to see in the span and for what reason. This will bring us to a conclusion faster, which is overall better for the open source community as we have limited time.

https://github.com/openzipkin/sleuth-webmvc-example

adriancole:
PS if API-gateway is the motivation, do mention if this implies traces are restarting. For example, you may be able to set your gateway to not trace and instead forward B3 headers so that this doesn't happen. We need to get to the point of whether we are working around infrastructure or what..

team3:
Let's take 2 cases into consideration.

`@RestController
public class SomeController {
    private final Logger logger = LogManager.getLogger();
    private final RestTemplate restTemplate;

    @Autowired
    public SomeController(RestTemplate restTemplate) {
        this.restTemplate = restTemplate;
    }

    @RequestMapping(path = "/some/action/google")
    public void handleGoogleGetRequest() {
        logger.traceEntry("handleGoogleGetRequest");

        ResponseEntity<String> forEntity = restTemplate.getForEntity("http://google.com", String.class);

        logger.traceExit("handleGoogleGetRequest");
    }

    @RequestMapping(path = "/some/action/internal")
    public void handleServiceGetRequest() {
        logger.traceEntry("handleServiceGetRequest");

        ResponseEntity<String> forEntity = restTemplate.getForEntity("http://127.0.0.1:8082/service1/test", String.class);

        logger.traceExit("handleServiceGetRequest");
    }
}

This controller is hosted under 'gateway' application. There is also 'service1' with sleuth and zipkin support which is called from gateway using RestTemplate. 

First action is `GET http://127.0.0.1:8080/some/action/google`
Second action is `GET http://127.0.0.1:8080/some/internal`.

Please find the traces from action#1 and action#2. There are no http response codes in spans.
[trace_google.txt](https://github.com/spring-cloud/spring-cloud-sleuth/files/1514791/trace_google.txt)
[trace_service1.txt](https://github.com/spring-cloud/spring-cloud-sleuth/files/1514793/trace_service1.txt)

marcingrzejszczak:
So currently what you can do is to `tracer#getCurrentSpan()` and set a tag yourself in the parent span of the RPC span. We're trying to minimize the new features in Sleuth cause we will be rewriting a lot of logic to use Brave internally and we want to focus on this.

PR #787, opened 2017-11-20
Made ZIpkin 2 default

### LINKED ISSUE #785: Consider renaming / removing Zipkin starters in Edgware / Finchley
The default for Edgware is to use Zipkin 2 - then the default starter should use `zipkin2` and should be called `starter-zipkin`. 

The legacy Zipkin 1 should be called `zipkin1` - since it's not the default.

In Finchley I guess we should remove `zipkin1` support - that's a perfect occasion to do such a change.

cc @snicoll @adriancole @dsyer

marcingrzejszczak:
- `zipkin` is `zipkin-legacy`
- `zipkin2` is `zipkin`

fixes #785


marcingrzejszczak:
We've changed the default in Sleuth so that the default `zipkin` starter is `zipkin2`. To use Zipkin1 you have to use `zipkin-legacy` module

PR #789, opened 2017-11-20
Using TraceFeignObjectWrapper#wrap in TraceFeignAspect instead of creating a TraceFeignClient

ScienJus:
We should not create a `TraceFeignClient` directly, because this will make `TraceLoadBalancerFeignClient` be wrapped at the same time, instead, we should use the `TraceFeignObjectWrapper#wrap` method to do this.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/TraceFeignAspect.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/TraceFeignAspect.java
@@ -28,6 +28,7 @@ import feign.Request;
  * Aspect for Feign clients so that you can autowire your custom components
  *
  * @author Marcin Grzejszczak
+ * @author ScienJus
  * @since 1.1.2
  */
 @Aspect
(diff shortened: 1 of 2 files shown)


marcingrzejszczak:
Please write a test that this improvement fixes

ScienJus:
Hi @marcingrzejszczak 

I created a simple test to verify that `TraceLoadBalancerFeignClient` will not be wrapped into a `TraceFeignClient` now.

But I do not know how to prove that the last version will be wrapped and create two spans, do you have any good ideas?

marcingrzejszczak:
Thanks for filing the PR. I've managed to write a simpler test and a simpler implementation. I've created the corresponding issue here https://github.com/spring-cloud/spring-cloud-sleuth/issues/791 and this is the code: https://github.com/spring-cloud/spring-cloud-sleuth/commit/b4ecb6f9576c141a099f2f71d5f472685d1ae4b2

ScienJus:
@marcingrzejszczak 

I still recommend using `TraceFeignObjectWrapper#wrap`, because this method has already defined the same logic(making a feign client traceable), which can be directly reused.

In addition, there is a difference in commit #972 from the issue which shows that http.url of span is the service name instead of ip address.

When client is a `LoadBalancerFeignClient(Client)`, `TraceFeignContext` will wrap it to `TraceLoadBalancerFeignClient(TraceFeignClient(Client))`, and `TraceFeignAspect` in your commit will wrap it to `TraceFeignClient(LoadBalancerFeignClient(Client))`. the difference between them is whether a span would be created before load balancing, Witch indicate that the `http.url` of span will be service name instead of a ip address.

The logic in `TraceFeignAspect` should be consistency with `TraceFeignContext`.

marcingrzejszczak:
Can you create a sample that proves this behavior?

ScienJus:
@marcingrzejszczak 

I think I misunderstood the case. It should not happen again. Thanks for the fix.

marcingrzejszczak:
Actually, you were right about the `TraceFeignObjectWrapper`. That will indeed a better solution. 

In general, it's always much better to start with an issue than a PR. Also, it's good to write a test / sample to replicate the problem. In your case, the screenshot was the most useful part.

PR #815, opened 2017-12-28
Add "Authorization" Header if spring.zipkin.base-url as uri have userinfo

yoker:
I have a `zipkin-server` using http collector to follow trace, I can get trace data.
after adding `spring-boot-starter-security`, I got nothing.
but I found No Authorization Header to send to `zipkin-server` from zipkin-client when I set Authorization info in `spring-zipkin-base-url`.

This PR can fix it.

### DIFF
--- a/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/DefaultZipkinRestTemplateCustomizer.java
+++ b/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/DefaultZipkinRestTemplateCustomizer.java
@@ -18,12 +18,15 @@ package org.springframework.cloud.sleuth.zipkin2;
 
 import java.io.ByteArrayOutputStream;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.util.zip.GZIPOutputStream;
 
 import org.springframework.http.HttpRequest;
 import org.springframework.http.client.ClientHttpRequestExecution;
 import org.springframework.http.client.ClientHttpRequestInterceptor;
 import org.springframework.http.client.ClientHttpResponse;
+import org.springframework.util.Base64Utils;
+import org.springframework.util.StringUtils;
 import org.springframework.web.client.RestTemplate;
 
 /**
@@ -47,6 +50,11 @@ public class DefaultZipkinRestTemplateCustomizer implements ZipkinRestTemplateCu
 		if (this.zipkinProperties.getCompression().isEnabled()) {
 			restTemplate.getInterceptors().add(0, new GZipInterceptor());
 		}
+		if(this.zipkinProperties.getUser() != null){
+			String username = this.zipkinProperties.getUser().getName();
+			string password = this.zipkinProperties.getUser().getPassword();
+			restTemplate.getInterceptors().add(1, new BasicAuthorizationInterceptor(username, password));
+		}
 	}
 
 	private class GZipInterceptor implements ClientHttpRequestInterceptor {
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@yoker Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=815)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=815) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@yoker Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=815)!

marcingrzejszczak:
Hi @yoker . Thanks for the PR. Can you write some tests for this functionality? Also @adriancole , WDYT?

marcingrzejszczak:
Ping? @yoker

adriancole:
usually a bit odd to add support for something that zipkin doesn't pack-in
(for example basic auth). I'm not sure if we should expect a URL to have
user info embedded or not. Usually I solve non-standard things by allowing
folks to configure an interceptor. I suppose this is an effort to allow a
non-standard server setup with configuration as opposed to a filter that
does the same..

yoker:
@marcingrzejszczak 
I pushed a repo at https://github.com/yoker/zipkin-server-sample.
In `zipkin/src/main/resources/application.yml`, if `security.basic.enabled=false`, there is ok. i can get trace info.
if `security.basic.enabled=true`, the SpanReporter got 401 error even though i added user:password in `spring.zipkin.base-url` that made it like this `http://user:password@zipkin/` in gateway, service etc.

@adriancole 
I haven't completely understood your means.  I guess there is a filter or an interceptor can do it.  I am sorry that I don't know which class is.

marcingrzejszczak:
You can use https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/ZipkinRestTemplateCustomizer.java to apply any customizations to the rest template that sends spans to zipkin. Including adding authorization headers.

yoker:
thanks @marcingrzejszczak 
i reverted the pr code, and i added BasicAuthorizationInterceptor to do it. maybe this is a good way for that.

marcingrzejszczak:
My point is that the PR is unnecessary cause you can create just 1 bean and inject the security headers.

yoker:
@marcingrzejszczak 
i don't agree with you. 
yes, i can do it just create 1 bean and inject, but i will do it again and again for all of my sub-services.
if the framework can do it, why not added?

spencergibb:
@yoker custom auto-configuration? Only include one dependency?

adriancole:
one hesitation about the approach is this implies a hard-coded
user/password combo. This means your business services need to restart in
order to update the credentials etc. Might be fine for your scenario, but
not ideal to recommend generically. If we get a lot of requests we could
still consider it.. usually there's a "rule of 3" type of deal, especially
as meanwhile you can do what spencer mentioned which is to make an
autoconfig jar added to your apps like any other dep.

yoker:
@adriancole 
most of us to collect trace using MQ like kafka, rabbitmq etc. so i guess there are a few people like the request.
Before sending PR, i found i can use credentials on `eureka.client.service-url.default-zone`, but i used it on `spring.zipkin.base-url`, i was failed.
i just think that was confusing to our developer who can't see full docs. 
we can do it to give the same experience.

adriancole:
The point remains that such a change and expectation set should be
requested by multiple people especially considering this is not a standard
service setup.

Right now it is only you asking for this whereas maintenance burden alone
is more than one person over time and causes questions as people will ask
how to setup basic auth (which isnt present on the server!)

So please use the workaround mentioned as it lets us focus efforts
somewhere besides resummarizing the same answer. If things change, we can
revisit.. hope this is fair.

yoker:
I am sorry that i haven't understood your means. 
The PR can't affect anyone if they have used by kafka or rabbitmq stream, even include web (the default mode).
it only was a complement.

adriancole:
Yeah I totally misunderstood you! Sorry about that.

Meanwhile, here is an open issue on http auth on the server. Whoever wants
basic auth as out of box feature on client should upvote or otherwise on
the server. Eventhough I am not a fan of basic auth, if there is enough
interest and no active disinterest we can find a way to progress.

The other neat thing in general is anyone can make an autoconfiguration
client side for their favorite mechanism. It would need to configure a bean
of type zipkin2.reporter.Sender. This is how other transports which are not
default,  such as SQS get added.. but it could be done equally to wrap your
favorite http sender with an auth interceptor of choice.

adriancole:
By here  I meant here https://github.com/openzipkin/zipkin/issues/782

PR #829, opened 2018-01-19
Sleuth now uses Brave

### LINKED ISSUE #92: Rename or rejig Sampler
Sampled returning false in sleuth is different than most instrumentation I've seen. For example, a span is managed regardless of whether it is sampled (i.e. still generates callbacks etc). The Sampled state controls the exportable field, which in turn controls propagation and storage (and a special-case for users of `TraceManager.addAnnotation(key, value)` as opposed to `getCurrentSpan`)

I'd suggest 2 paths, which could both be taken.
- Rename Sampler to ShouldExport. Document the propagation and storage aspects there.
- Move the ShouldExport logic to the end of a span. Since we are gathering a Span regardless anyway, we might as well use it in the export decision. In other words, change the invocation from before-the-fact (before a span id is created) until after-the-fact (when a span ends or times out).

If controlling overhead thing is an intended goal of the Sampler, then maybe the before-the-fact is still the right choice. I'd have a few suggestions on that.
- Trim the state needed.. seems we shouldn't be creating callbacks and the like for unsampled spans. That or it should be clear why we do (maybe it is to others).
- Expand guarding to things beyond storage, propagation, TraceManager.setAttribute, or invert the logic. Ex. don't use the Span object at all if not sampled.. use something different (assuming the goal of keeping it around is to coordinate ids or something).

### LINKED ISSUE #143: Propagate context not spans
We could make sleuth (or commons?) propagate not `Span` but a `DistributionContext` whose part is a `Span`. In other words tracing would be a part of the distribution context that would get propagated.

### LINKED ISSUE #255: Document how to use other span stores in sleuth-zipkin-stream
The stream server is a zipkin server, so supports other span stores such as elasticsearch or cassandra. We should document that, probably after modularization work completes (to avoid thrash).

Ex here are the properties, but they will probably move to another library https://github.com/openzipkin/zipkin/blob/master/zipkin-autoconfigure/storage-elasticsearch/src/main/java/zipkin/autoconfigure/storage/elasticsearch/ZipkinElasticsearchStorageProperties.java

https://github.com/openzipkin/zipkin-java/issues/108

### LINKED ISSUE #305: Integration with gRPC
Is instrumentation for gRPC in your roadmap for anytime soon?

### LINKED ISSUE #459: Support AWS X-Ray

### LINKED ISSUE #577: Cleanup message building in TraceChannelInterceptor

### LINKED ISSUE #711: Add support for Brave autoconfiguration

marcingrzejszczak:
with this pull request we have rewritten the whole Sleuth internals to use [Brave](https://github.com/openzipkin/brave). That way we can leverage all the functionalities & instrumentations that Brave already has (https://github.com/openzipkin/brave/tree/master/instrumentation).

Migration guide is available here: https://github.com/spring-cloud/spring-cloud-sleuth/wiki/Spring-Cloud-Sleuth-2.0-Migration-Guide

fixes #711 - Brave instrumentation
fixes #92 - we move to Brave's Sampler
fixes #143 - Brave is capable of passing context
fixes #255 - we've moved away from Zipkin Stream server
fixes #305 - Brave has GRPC instrumentation (https://github.com/openzipkin/brave/tree/master/instrumentation/grpc)
fixes #459 - Brave (https://github.com/openzipkin/brave/pull/510) & Zipkin (https://github.com/openzipkin/zipkin/issues/1754) will deal with the AWS XRay instrumentation 
fixes #577 - Messaging instrumentation has been rewritten


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/ExceptionMessageErrorParser.java
```diff
@@ -13,16 +12,20 @@
...
-		if (span != null && span.isExportable()) {
-			String errorMsg = ExceptionUtils.getExceptionMessage(error);
+	public void parseErrorTags(SpanCustomizer span, Throwable error) {
+		if (span != null && error != null) {
+			String errorMsg = getExceptionMessage(error);
 			if (log.isDebugEnabled()) {
 				log.debug("Adding an error tag [" + errorMsg + "] to span " + span);
 			}
-			span.tag(Span.SPAN_ERROR_TAG_NAME, errorMsg);
+			span.tag("error", errorMsg);
```
spencergibb: constant?

PR #847, opened 2018-02-06
Fix resource leak. Close InetUtils and nested executor service

igor-suhorukov:


spencergibb:
Can you explain your reasoning? The reason it is assigned is that it gets reused. What resource leak is there?

igor-suhorukov:
@spencergibb it used only as local variable and this reference will be lost after return. This case detected by static code analysis. Under the hood InetUtils open executor service

spencergibb:
`InetUtils` should be a spring bean and be subject to a bean lifecycle (ie: `Closable` beans will get their `close()` method called. In what situation do you encounter the case where `inetUtils` is null?

spencergibb:
It's rather require `InetUtils` to not be null, then remove the `if` statement.

igor-suhorukov:
>> In what situation do you encounter the case where inetUtils is null?
So we just need 
		if (inetUtils == null) {
		    throw new IllegalArgumentException("inetUtils");
		}

spencergibb:
I'd reccomend using Springs `Assert` class in the constructor.
```
		Assert.notNull(inetUtils, "inetUtils may not be null");
		this.firstNonLoopbackAddress = findFirstNonLoopbackAddress(inetUtils);
```

igor-suhorukov:
thanks  @spencergibb  This advice sounds reasonable

spencergibb:
@igor-suhorukov looks like some tests don't pass in an `InetUtils` and failed, please update.

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/ZipkinAutoConfiguration.java
```diff
@@ -93,6 +94,12 @@ public ZipkinRestTemplateCustomizer zipkinRestTemplateCustomizer(ZipkinPropertie
 		return new DefaultZipkinRestTemplateCustomizer(zipkinProperties);
 	}
 
+	@Bean
+	@ConditionalOnMissingBean
+	InetUtils zipkinInetUtils(){
```
spencergibb: I don't think this is needed. Why did you add this?

igor-suhorukov: Because you say:
`InetUtils should be a spring bean and be subject to a bean lifecycle (ie: Closable beans will get their close() method called. `

spencergibb:
But it's already created by a different project

Issue #821, opened 2018-01-05
Tracing is not working with spring-cloud/spring-cloud-stream defined @input channels


lchayoun:
Spring Cloud Sleuth is using TraceChannelInterceptor in order to continue the spans through the messaging system.
TraceChannelInterceptor is using the ExecutorChannelInterceptor#beforeHandle method.
If the channel is not an ExecutorSubscribableChannel tracing is not possible along with other capabilities regarding the parsing of incoming messages headers.

Currently the only way to handle messages before the method annotated with "@streamlistener" is called is on PollableChannels with the "postReceive" method.
When using SubscribableChannel this is possible only using the ExecutorChannelInterceptor#beforeHandle method.
But ExecutorSubscribableChannel is not supported by the "@input" annotation factory,

spring-cloud/spring-cloud-stream#1173

artembilan:
I think the situation is like `TraceChannelInterceptor` doesn't support `AbstractMessageChannel` from Spring Integration. I guess the simplest test would be don't worry about SCSt but just try to apply `TraceChannelInterceptor` to the `DirectChannel` and verify afterward.

There is a `ThreadStatePropagationChannelInterceptor` abstraction in Spring Integration to consider for this situation.

The other problem that this interceptor doesn't support `PollableChannel` as well...

adriancole:
https://github.com/openzipkin/brave/pull/572 will fix all of this or die trying

PR #860, opened 2018-02-14
Fixes DirectChannel support

marcingrzejszczak:
without this change for DirectChannel `beforeHandle` and `afterHandle` were not called at all
with this change we're removing any hacks and we're executing those methods

fixes gh-821

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TraceChannelInterceptor.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TraceChannelInterceptor.java
@@ -25,6 +25,7 @@ import org.springframework.beans.factory.BeanFactory;
 import org.springframework.cloud.sleuth.Log;
 import org.springframework.cloud.sleuth.Span;
 import org.springframework.cloud.sleuth.sampler.NeverSampler;
+import org.springframework.integration.channel.DirectChannel;
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageChannel;
 import org.springframework.messaging.MessageHandler;
@@ -53,6 +54,9 @@ public class TraceChannelInterceptor extends AbstractTraceChannelInterceptor {
 	@Override
 	public void afterSendCompletion(Message<?> message, MessageChannel channel,
 			boolean sent, Exception ex) {
+		if (channel instanceof DirectChannel) {
+			afterMessageHandled(message, channel, null, ex);
+		}
 		Message<?> retrievedMessage = getMessage(message);
 		MessageBuilder<?> messageBuilder = MessageBuilder.fromMessage(retrievedMessage);
 		Span currentSpan = getTracer().isTracing() ? getTracer().getCurrentSpan()
@@ -61,13 +65,7 @@ public class TraceChannelInterceptor extends AbstractTraceChannelInterceptor {
 			log.debug("Completed sending and current span is " + currentSpan);
 		}
 		getTracer().continueSpan(currentSpan);
-		if (containsServerReceived(currentSpan)) {
-			if (log.isDebugEnabled()) {
-				log.debug("Marking span with server send");
-			}
-			currentSpan.logEvent(Span.SERVER_SEND);
-		}
-		else if (currentSpan != null) {
+		if (currentSpan != null) {
 			if (log.isDebugEnabled()) {
 				log.debug("Marking span with client received");
 			}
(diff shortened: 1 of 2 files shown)


spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/messaging/TraceChannelInterceptorTests.java
```diff
@@ -194,6 +196,12 @@ public void parentSpanIncluded() {
 		then(traceId).isEqualTo(10L);
 		then(spanId).isNotEqualTo(20L);
 		then(this.accumulator.getSpans()).hasSize(1);
+		List<Log> logs = this.accumulator.getSpans().get(0).logs();
+		then(logs).hasSize(4);
+		then(logs.get(0).getEvent()).isEqualTo(Span.CLIENT_SEND);
+		then(logs.get(1).getEvent()).isEqualTo(Span.SERVER_RECV);
+		then(logs.get(2).getEvent()).isEqualTo(Span.SERVER_SEND);
+		then(logs.get(3).getEvent()).isEqualTo(Span.CLIENT_RECV);
```
artembilan: Yeah... That's cool! That's exactly what I would like to see in the test-case. 😄 
Although these names are a bit misleading. Anyway I think this now maps directly to the:
```
preSend -> beforeHandle -> afterHandled -> postSend(afterSendComletion)
```

And this should be exactly the same what we have with the `ExecutorChannel`.
Although there thanks to the thread shifting we have a slightly different model:
```
preSend -> postSend(afterSendComletion)
      beforeHandle -> afterHandled
```

Anyway I think that is exactly what we are expecting from such a behavior!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TraceChannelInterceptor.java
```diff
@@ -53,6 +54,9 @@ public TraceChannelInterceptor(BeanFactory beanFactory) {
 	@Override
 	public void afterSendCompletion(Message<?> message, MessageChannel channel,
 			boolean sent, Exception ex) {
+		if (channel instanceof DirectChannel) {
```
artembilan: I think that's not good. The channel might be an AOP proxy, so we don't have an access to the target class.
And also there are many other `AbstractMessageChannel` which maybe tracing candidates as well.

In other words we really should check here for those channel types who are *interceptable*.

In my best feeling that should be like:
```
if (!(channel instanceof InterceptableChannel) && 
          !(channel instanceof ExecutorChannelInterceptorAware) && 
           channel instanceof ChannelInterceptorAware)
```
No, classes to check, only interfaces!
And looks like all the cases are covered.

The `InterceptableChannel` and `ExecutorChannelInterceptorAware` works as is without this fix though.

WDYT?

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/messaging/TraceChannelInterceptorTests.java
marcingrzejszczak: Yeah... the test was wrong. It wasn't testing enough. Now it's testing that: 

- the message was sent and waiting for response (Client Sent)
- the message was received and started processing (Server Received)
- the message was received and finished processing (Server Sent)
- the message was sent and response was received (Client Received)

Before the change no "Server side" was checked

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TraceChannelInterceptor.java
marcingrzejszczak: As discussed offline, I've added the check on the proxy of the `DirectChannel` and we came to the conclusion that the only class that makes sense to verify is indeed `DirectChannel` cause the processing will take care in the same JVM process.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TraceChannelInterceptor.java
```diff
@@ -206,14 +215,32 @@ public void afterMessageHandled(Message<?> message, MessageChannel channel,
 			addErrorTag(ex);
 		}
 		// related to #447
-		if (getTracer().isTracing()) {
+		if (getTracer().isTracing() && !(channel instanceof DirectChannel)) {
```
artembilan: You have missed this to fix as well 😄

marcingrzejszczak: 🤦‍♂️

PR #861, opened 2018-02-14
Fixed NPE when the passed message is null.

TYsewyn:
This is the case when working with a QueueChannel and a Poller, which will emit a null message when the queue is empty.


marcingrzejszczak:
Some changes were applied to the same classes so you have to rebase. Also, I prefer when we assert against a negative condition.

```java
if (!shouldTrace(..))) {
return;
}
...
```

that way we don't have to wrap in an if but fail fast and we don't have indented code.

marcingrzejszczak:
Also please apply the change against `1.3.x` branch and I'll merge it to master afterwards

Issue #873, opened 2018-03-01
Align HTTP instrumentation with Brave


marcingrzejszczak:
- [ ] TraceFilter
- [ ] TraceHandlerInterceptor

PR #878, opened 2018-03-03
Aligned HTTP instrumentation with Brave

marcingrzejszczak:
with this change we remove TraceFilter and make TraceHandlerInterceptor a legacy component. All HTTP instrumentation is done via Brave.

fixes gh-873

NOTE:

Need to update https://github.com/spring-cloud/spring-cloud-sleuth/wiki/Spring-Cloud-Sleuth-2.0-Migration-Guide with info that:
-  `TraceFilter` got removed 
- `TraceHandlerInterceptor` got renamed to `SleuthTraceHandlerInterceptor` and made package scoped
- `TraceFilter.ORDER` got moved to `TraceWebServletAutoConfiguration.TRACING_FILTER_ORDER`


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebServletAutoConfiguration.java
```diff
@@ -54,6 +56,8 @@
 @AutoConfigureAfter(TraceHttpAutoConfiguration.class)
 public class TraceWebServletAutoConfiguration {
 
+	public static final int TRACING_FILTER_ORDER = Ordered.HIGHEST_PRECEDENCE + 5;
```
marcingrzejszczak: I don't know if this is the best place to put it...

Issue #802, opened 2017-12-04
Trace tags are not accounting for parameterized URLs


marcingrzejszczak:
https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-spring-legacy/src/main/java/io/micrometer/spring/web/servlet/WebMvcTags.java#L68

jkschneider:
@marcingrzejszczak [Here](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/servlet/WebMvcTags.java).

marcingrzejszczak:
related to https://github.com/openzipkin/brave/pull/602

marcingrzejszczak:
Done through https://github.com/spring-cloud/spring-cloud-sleuth/pull/878

PR #880, opened 2018-03-04
Hacks standard span names in, but redundantly parsing

adriancole:
This has no tests, and needs to be done similarly for flux. feel free to close this with a different PR or commit. The approach is redundantly parsing at the mvc layer even when servlet closes the span.

This is kindof ok for now I think, as we have other work in brave needed to make the practice more standard (because we might try to avoid redundant parsing for the small overhead it makes.. not sure it will be possible)

Once this is in, things look pretty by default. ex building then using snapshot on https://github.com/openzipkin/sleuth-webmvc-example Once something like this is together, I'd recommend cutting a milestone as pretty spans are better than "get", "get", "get"

screen shot 2018-03-04 at 10 47 52 am

BTW, this is https://github.com/spring-cloud/spring-cloud-sleuth/issues/802, however it doesn't add the http.route tag by default, rather just uses it for the span name. I think micrometer's tag policy is mildly different at least in case format anyway, but anyway span names are a great start

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHandlerInterceptor.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHandlerInterceptor.java
@@ -25,6 +25,7 @@ import brave.Tracer;
 import brave.http.HttpServerHandler;
 import brave.http.HttpTracing;
 import brave.servlet.HttpServletAdapter;
+import javax.servlet.http.HttpServletResponseWrapper;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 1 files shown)


adriancole:
actually I think I have a more sustainable alternative:

All things layered on servlet add the request attribute "http.route" some time before a response completes. If the servlet closes the span, it can find and parse this. If something else closes the span, they use however they like.

adriancole:
we can do something like what's in this change temporarily until https://github.com/openzipkin/brave/pull/635 or we can merge that and rebase this.

marcingrzejszczak:
I guess that's made obsolete via https://github.com/spring-cloud/spring-cloud-sleuth/pull/878 ?

marcingrzejszczak:
I'd go with merging https://github.com/openzipkin/brave/pull/635 and updating Brave in Sleuth

adriancole:
yeah the other approach is better

PR #885, opened 2018-03-06
Prefers "plain bean names" for community instrumentation

adriancole:
Instrumentation around brave is more likely to use a simple bean name
like "tracing" vs "sleuthTracing". This changes the commodity beans to
simple names so that we can avoid having to teach naming prefixes unless
necessary.

This affects the following beans:
* httpTracing
* tracing
* tracer
* spanCustomizer

This came up when integrating dubbo, as their [extension loader](https://github.com/alibaba/dubbo/blob/master/dubbo-config/dubbo-config-spring/src/main/java/com/alibaba/dubbo/config/spring/extension/SpringExtensionFactory.java#L43) prefers
stable bean names. For example, loading `brave.Tracing` with their
spring extension silently failed because our bean was named
"sleuthTracing". Even if we can provide instructions to override this,
seems best to dodge.


adriancole:
Now, dubbo works out of the box with sleuth like this..

screen shot 2018-03-06 at 3 44 39 pm

Thanks to Jun Liu for the hint.

Issue #886, opened 2018-03-06
Allow custom HttpClientParser and HttpServerParser


saturnism:
Working on https://github.com/spring-cloud/spring-cloud-gcp/pull/438 to provide auto-configuration to propagate trace data to Stackdriver Trace.

Sleuth 2.0.0.M6+ is now using Brave. The default configuration doesn't provide the same information as Sleuth 1.x.  A couple of tags are missing and would be beneficial to Stackdriver Trace.  Namely, `http/host` and `http/url`.

Upon inspecting the Sleuth legacy implementation, we may also want to not use it but use custom parsers to produce spans with name and tags that's better suited for Stackdriver Trace.

However, there is currently no way to produce `HttpTracing` bean to override the Sleuth auto-configuration, because several `SkipPatternProvider`, `SleuthHttpClientParser`, `SleuthHttpServerParser`, and `SleuthHttpSampler` are all package protected.  Even if we produce the `HttpClientParser` and `HttpServerParser`, there is still no way to create `SleuthHttpSampler` nor injecting existing `SkipPatternProvider`.

After discussing w/ @marcingrzejszczak, it would be beneficial to be able to take custom `HttpClientParser` and `HttpServerParser` beans and adding/appending them to the existing auto-configured `HttpTracing` configuration.

adriancole:
sounds fine. just if someone overrides all of HttpTracing, they'll also override these parsers, which is probably fine (more a documentation thing). FWIW exactly the same issue for x-ray

PR #888, opened 2018-03-07
Added support for customization of HttpTracing

marcingrzejszczak:
with this change you can define beans that will get added / set to the `HttpTracing` bean

fixes gh-886

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-sleuth.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-sleuth.adoc
@@ -762,11 +762,13 @@ Running the preceding method with a value of `15` leads to setting a tag with a 
 
 //=== Spring Integration
 
-// TODO: Update this
+=== HTTP
 
-// === HTTP
-
-// TODO: Update this
+If a customization of client / server parsing of the HTTP related spans is required,
+just register a bean of type `brave.http.HttpClientParser` or
+`brave.http.HttpServerParser`. If client /server sampling is required, just
+register a bean of type `brave.http.HttpSampler` and name the bean
+ `sleuthClientSampler` for client sampler and `sleuthServerSampler` for server sampler.
 
 === `TraceFilter`
 
--- a/docs/src/main/asciidoc/spring-cloud-sleuth.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-sleuth.adoc
@@ -770,6 +770,18 @@ just register a bean of type `brave.http.HttpClientParser` or
 register a bean of type `brave.http.HttpSampler` and name the bean
  `sleuthClientSampler` for client sampler and `sleuthServerSampler` for server sampler.
 
+Check out Brave's code to see an example of how to make a path-based sampler
+https://github.com/openzipkin/brave/tree/master/instrumentation/http#sampling-policy
+
+If you want to completely rewrite the `HttpTracing` bean you can use the `SkipPatternProvider`
+interface to retrieve the URL `Pattern` for spans that should be not sampled. Below you can see
+an example of usage of `SkipPatternProvider` inside a `HttpSampler`.
+
+[source,java]
+----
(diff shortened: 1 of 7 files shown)


docs/src/main/asciidoc/spring-cloud-sleuth.adoc
```diff
@@ -762,11 +762,13 @@ Running the preceding method with a value of `15` leads to setting a tag with a
 
 //=== Spring Integration
 
-// TODO: Update this
-
-// === HTTP
+=== HTTP
 
-// TODO: Update this
+If a customization of client / server parsing of the HTTP related spans is required,
```
adriancole: I might say what you have and link to brave's docs for more

ex https://github.com/openzipkin/brave/tree/master/instrumentation/http#sampling-policy shows how to make a path-based sampler

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
```diff
@@ -762,11 +762,13 @@ Running the preceding method with a value of `15` leads to setting a tag with a
...
+just register a bean of type `brave.http.HttpClientParser` or
+`brave.http.HttpServerParser`. If client /server sampling is required, just
+register a bean of type `brave.http.HttpSampler` and name the bean
+ `sleuthClientSampler` for client sampler and `sleuthServerSampler` for server sampler.
```
adriancole: do you mean name the bean or use the annotation?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHttpAutoConfiguration.java
```diff
@@ -39,28 +48,122 @@
...
+	}
+
+	@Bean
+	@ConditionalOnMissingBean(name = "sleuthClientSampler")
+	HttpSampler sleuthClientSampler() {
+		return HttpSampler.TRACE_ID;
+	}
+}
+
+class CompositeClientParser extends HttpClientParser {
```
adriancole: composite is trickier than just allowing someone to override the type. For example, they can't remove tags this way. I might start with just letting people override it. If composition becomes common we can add

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHttpAutoConfiguration.java
```diff
@@ -39,28 +48,122 @@
...
+		this.parsers.forEach(parser -> parser.request(adapter, req, customizer));
+	}
+
+	@Override public <Resp> void response(HttpAdapter<?, Resp> adapter, Resp res,
+			Throwable error, SpanCustomizer customizer) {
+		this.parsers.forEach(parser -> parser.response(adapter, res, error, customizer));
+	}
 }
+
+class CompositeServerSampler extends HttpSampler {
```
adriancole: Personally I think the skipPattern sampler is fantastically easy to override. Just have folks use the property as opposed to composing over it?

Ex SkipPattern is public so someone can easily include it in their own sampler.:
```java
class MyHttpSampler extends HttpSampler {

	private final Pattern pattern;

	SleuthHttpSampler(SkipPatternProvider provider) {
		this.pattern = provider.skipPattern();
	}

	@Override public <Req> Boolean trySample(HttpAdapter<Req, ?> adapter, Req request) {
		String url = adapter.path(request);
		boolean shouldSkip = this.pattern.matcher(url).matches();
		if (shouldSkip) {
			return false;
		}
		return null;
	}
}
```

adriancole:
Added a commit which is mostly just polish. I do think we'll eventually get a request to not use the sleuth sampling pattern, but then again, this can also be done by overriding the pattern. So basically sounds good to me.

marcingrzejszczak:
Fixed via https://github.com/spring-cloud/spring-cloud-sleuth/commit/375f244279265281c8fe91894522e5641ebcb7cb

PR #901, opened 2018-03-14
Adds test to make sure RestTemplate sender doesn't mess with json

adriancole:
There was some doubt on gitter about sleuth generating spans with floating-point
rendered timestamps. While this shouldn't break people we certainly shouldn't
create json like this.

This was presumed to be caused by sleuth's rest template sender by @mojsha who
was trying Jaeger. Jaeger's parser doesn't give any errors with a 400.
```json
[
{
"traceId": "a65dac9d8bbb57f8",
"parentId": "a65dac9d8bbb57f8",
"id": "f223bb97e1eba9b2",
"kind": "SERVER",
"name": "https:\/hello3",
"timestamp": 1.521027116604e+15,
"duration": 1525,
"localEndpoint": {
"serviceName": "${spring.application.name}",
"ipv4": "10.128.6.98",
"port": 8443
},
```

I added a test to make sure it is not the case.

### DIFF
--- a/spring-cloud-sleuth-zipkin/pom.xml
+++ b/spring-cloud-sleuth-zipkin/pom.xml
@@ -149,6 +149,13 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
+			<groupId>io.zipkin.zipkin2</groupId>
+			<artifactId>zipkin</artifactId>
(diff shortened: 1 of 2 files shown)


adriancole:
also tested on 1.3.x

PR #906, opened 2018-03-20
Adds the ability to configure things needed by stackdriver

adriancole:
We will eventually need to change the encoder to use their format. Also,
stackdriver does not support shared spans.


adriancole:
@saturnism fyi things are aligned now with https://github.com/openzipkin/zipkin-gcp/pull/62

saturnism:
thank you!

PR #923, opened 2018-04-03
Improve reactive server span name generation

lightoze:
The patch implements `HttpAdapter.methodFromResponse(response)` and `HttpAdapter.route(response)` which are required for Brave HTTP span name generation (`HttpParser.spanNameFromRoute`).

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebFilter.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebFilter.java
@@ -31,6 +31,7 @@ import org.springframework.core.Ordered;
 import org.springframework.http.HttpHeaders;
 import org.springframework.http.server.reactive.ServerHttpRequest;
 import org.springframework.http.server.reactive.ServerHttpResponse;
+import org.springframework.http.server.reactive.ServerHttpResponseDecorator;
 import org.springframework.web.method.HandlerMethod;
 import org.springframework.web.reactive.HandlerMapping;
 import org.springframework.web.server.ServerWebExchange;
@@ -125,9 +126,7 @@ public final class TraceWebFilter implements WebFilter, Ordered {
 			// clear any previous trace
 			tracer().withSpanInScope(null);
 		}
-		ServerHttpRequest request = exchange.getRequest();
-		ServerHttpResponse response = exchange.getResponse();
-		String uri = request.getPath().pathWithinApplication().value();
+		String uri = exchange.getRequest().getPath().pathWithinApplication().value();
 		if (log.isDebugEnabled()) {
 			log.debug("Received a request to uri [" + uri + "]");
 		}
@@ -149,15 +148,22 @@ public final class TraceWebFilter implements WebFilter, Ordered {
 							} else {
 								continuation = Mono.empty();
 							}
+							String httpRoute = null;
 							Object attribute = exchange
 									.getAttribute(HandlerMapping.BEST_MATCHING_HANDLER_ATTRIBUTE);
 							if (attribute instanceof HandlerMethod) {
 								HandlerMethod handlerMethod = (HandlerMethod) attribute;
 								addClassMethodTag(handlerMethod, span);
 								addClassNameTag(handlerMethod, span);
+								Object pattern = exchange
+										.getAttribute(HandlerMapping.BEST_MATCHING_PATTERN_ATTRIBUTE);
+								httpRoute = pattern != null ? pattern.toString() : "";
 							}
-							addResponseTagsForSpanWithoutParent(exchange, response, span);
-							handler().handleSend(response, t, span);
+							addResponseTagsForSpanWithoutParent(exchange, exchange.getResponse(), span);
+							DecoratedServerHttpResponse delegate = new DecoratedServerHttpResponse(
+									exchange.getResponse(), exchange.getRequest().getMethodValue(),
+									httpRoute);
+							handler().handleSend(delegate, t, span);
 							if (log.isDebugEnabled()) {
 								log.debug("Handled send of " + span);
 							}
@@ -181,7 +187,7 @@ public final class TraceWebFilter implements WebFilter, Ordered {
 									}
 								} else {
 									span = handler().handleReceive(extractor(),
-											request.getHeaders(), request);
+											exchange.getRequest().getHeaders(), exchange.getRequest());
 									if (log.isDebugEnabled()) {
 										log.debug("Handled receive of span " + span);
 									}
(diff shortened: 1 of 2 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebFilter.java
```diff
@@ -156,6 +158,11 @@ TraceKeys traceKeys() {
 								addClassMethodTag(handlerMethod, span);
 								addClassNameTag(handlerMethod, span);
 							}
+							Object httpRoute = exchange
+									.getAttribute(HandlerMapping.BEST_MATCHING_PATTERN_ATTRIBUTE);
```
adriancole: the contract is empty string when route is available (which I think is always the case here), just no match.

Ex. in webmvc we do..
```
    Object httpRoute = request.getAttribute(BEST_MATCHING_PATTERN_ATTRIBUTE);
    request.setAttribute("http.route", httpRoute != null ? httpRoute.toString() : "");
```

So here probably

```
    Object httpRoute = request.getAttribute(BEST_MATCHING_PATTERN_ATTRIBUTE);
    response.httpRoute = httpRoute != null ? httpRoute.toString() : "";
```

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebFilter.java
```diff
@@ -156,6 +158,11 @@ TraceKeys traceKeys() {
...
+							if (httpRoute != null) {
+								response.httpRoute = httpRoute.toString();
+							}
 							addResponseTagsForSpanWithoutParent(exchange, response, span);
 							handler().handleSend(response, t, span);
```
adriancole: I'd move the new DecoratedServerHttpResponse(exchange.getResponse()) here. it is only used by handleSend

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebFilter.java
```diff
@@ -255,6 +262,15 @@ private void addClassNameTag(Object handler, Span span) {
 		return ORDER;
 	}
 
+	static final class DecoratedServerHttpResponse extends ServerHttpResponseDecorator {
+
+		String method, httpRoute;
```
adriancole: if you move instantiating this to where handleSend is called, you can make these final and always assign a value

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebFilter.java
lightoze: Valid point. Speaking of that - is there any reason for `request` and `response` being referenced from lambdas instead of retrieving them from the exchange? Not a big optimization but still...

adriancole: you can fix this so long as other tests still pass. We haven't optimized this code yet for performance.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebFilter.java
lightoze: There you already know for sure that request is handled by MVC. In this case it is not always true (think reactive functions for example). Also I was not able to reproduce null `BEST_MATCHING_PATTERN_ATTRIBUTE` while staying in MVC context. However I noticed following corner cases:
- default error page provides `/**` as a pattern
- for catch-all `GetMapping` provided pattern equals to URI - whatever the client supplies

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebFilter.java
```diff
@@ -192,6 +196,15 @@ TraceKeys traceKeys() {
 						}));
 	}
 
+	private static String getHttpRoute(ServerWebExchange exchange) {
```
adriancole: maybe make the notes about the corner cases as comments here (better as unit tests, but if no tests, then at least comments keeps the knowledge around)

I suspect that if you are in the `attribute instanceof HandlerMethod` block, then you'd know you are in a controller, so if the pattern string is null in this case, you'd coerse to empty to fit the contract.

lightoze: I am still convinced that coercion is not needed because pattern attribute is never null for MVC calls. For other calls returning empty string will overwrite span name even if user have already manually set custom one - can be the case with router functions.

I'll add some comments, extend existing test to check generated span name and let's call it a day, OK? :)

adriancole: to see what I mean, you can clone brave locally. add a test that makes the thing you are concerned about, for example, in webmvc. the current approach to span customization (for example if you have a better name) is to have your own parser.

https://github.com/openzipkin/brave/blob/master/instrumentation/http-tests/src/main/java/brave/test/http/ITHttpServer.java

Even if we dodge it here, if there is a problem, the trap is set elsewhere. Not trying to be arbitrary, just this sort of inconsistency limits the ability of a couple people to maintain projects that a lot of people use.

lightoze: About `HttpAdapter.route` result - I've read the docs again and it says "Null indicates this instrumentation doesn't understand http routes". `TraceWebFilter` does support routes so its implementation of HttpAdapter should never return null. Is that correct?

I feel like I've lost the direction of your though. What do you suggest?

adriancole: At least when "attribute instanceof HandlerMethod" I suspect we can coerce null to empty. I think making sure that this happens (if it is correct that attribute instanceof HandlerMethod when defined on an annotated class), then we make progress and keep things consistent. So, when you are inside the  "attribute instanceof HandlerMethod" if statement make sure getHttpRoute null coerces to empty somehow

There may be other cases, too, but we can defer that to later tests.

Make sense?

Independently, we need to put a subclass of ITHttpServer here, because it is better to show not tell. This is something I will try to do offline.

lightoze: OK, I redid the whole patch according to this. Also added a test for router function to demonstrate span naming behaviour.

adriancole:
thanks for the help!

marcingrzejszczak:
Great work!

PR #928, opened 2018-04-05
Moves to Brave's ErrorParser and ScopedSpan

adriancole:
See https://github.com/openzipkin/brave/pull/666
See https://github.com/openzipkin/brave/pull/678


devinsba:
When I started working on this in Brave I had no idea how much it would simplify sleuth. This is great not having to propagate `errorParser` all over the place.

pom.xml
```diff
@@ -272,9 +272,9 @@
 		<spring-cloud-stream.version>Elmhurst.BUILD-SNAPSHOT</spring-cloud-stream.version>
 		<spring-cloud-netflix.version>2.0.0.BUILD-SNAPSHOT</spring-cloud-netflix.version>
 		<spring-cloud-openfeign.version>2.0.0.BUILD-SNAPSHOT</spring-cloud-openfeign.version>
-		<brave.version>4.18.2</brave.version>
+		<brave.version>4.19.0</brave.version>
 		<!-- Version set until zipkin-junit gets defined in Brave BOM -->
```
adriancole: this comment lies a little even if once true.. it is actually about zipkin's test jar I think.

PR #962, opened 2018-04-28
Adds proto3 encoding

adriancole:
Note: Zipkin server 2.8+ is required to be running for this to work.


adriancole:
tested by setting "spring.zipkin.encoder=PROTO3" in https://github.com/openzipkin/sleuth-webmvc-example using this build. The spans sent were half the size as usual.

PR #995, opened 2018-05-30
fix #994 Make SpanSubscriber fake fuseability

### LINKED ISSUE #994: Reactor (via Boot 2.0.2) still broken with Sleuth 2.0.0.RC2
Project that replicates the issue - https://github.com/joke/sleuth

simonbasle:
This is done by implementing NO-OP QueueSubscription on the subscriber
and removing the guard against `Fuseable` publisher sources.

As `Fuseable` is a marker interface there should be no reason for code
to perform casts so for now the fact that the `lift` call in
ReactorSleuth doesn't actually produces a `Fuseable` publisher should
not be a problem.


simonbasle:
Looks like I only changed one of the 3 subscribers. Will do a follow-up.

PR #1008, opened 2018-06-13
follow-up to #995 All SpanSubscription now fake QueueSubscription

simonbasle:


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/SpanSubscription.java
```diff
@@ -18,9 +18,40 @@
...
 
 /**
+ * A {@link SpanSubscription} is a {@link Subscription} that fakes being {@link Fuseable}
+ * (implementing {@link reactor.core.Fuseable.QueueSubscription} with default no-op methods
+ * and always negotiating fusion to be {@link Fuseable#NONE}).
+ *
  * @author Marcin Grzejszczak
  */
-interface SpanSubscription<T> extends Subscription, CoreSubscriber<T> {
+interface SpanSubscription<T> extends Subscription, CoreSubscriber<T>, Fuseable.QueueSubscription<T> {
```
adriancole: the current non-deprecated extender of this is `ScopePassingSpanSubscriber` which also implements Scannable.

It forwards to its delegate `Subscriber<? super T> subscriber`. I'm trying to understand why this implements so many interfaces besides its delegate? The above javadoc does a good idea explaining what is going on with the fuseable aspect, but I don't understand why, and I also don't understand the other interfaces. Just in case some of this is historical and no longer needed, can you or @robotmrv unpack this for me? I can re-add comments or javadoc based on this.

adriancole: If I take out the QueueSubscription thing, it fails because of a cast like this, somewhere late down the code.

```
java.lang.ClassCastException: org.springframework.cloud.sleuth.instrument.reactor.ScopePassingSpanSubscriber cannot be cast to reactor.core.Fuseable$QueueSubscription

	at reactor.core.publisher.FluxFilterFuseable$FilterFuseableSubscriber.onSubscribe(FluxFilterFuseable.java:80)
	at org.springframework.cloud.sleuth.instrument.reactor.ScopePassingSpanSubscriber.onSubscribe(ScopePassingSpanSubscriber.java:67)
```

This is one of the parts I'd like to understand..

simonbasle: @adriancole historically operator fusion is represented by a marker interface at the assembly level (`Publisher`s such as `Flux` or `Mono`) and a `QueueSubscription` at runtime (intermediate `Subscribers` use their upstream `Subscription` as a `Queue` and also negotiate the type of fusion via this interface).

Most Reactor operators detect if their source can _potentially_ be fused by checking said source `Publisher` implements `Fuseable` interface. From there on, it instantiates its own `Fuseable` variant of `Publisher` (so that subsequent fuseable operators can all be fused together).

But that's assembly time. What really needs to be fused is the chain of `Subscriber` once the reactive chain is activated/subscribed to, the one through which the data flows.

Historically, operators that have two variants one of which is `Fuseable` will just assume that since their source is `Fuseable`, the `Subscription` it will pass to them is a `QueueSubscription`. For instance `map` has [`FluxMap`](https://github.com/reactor/reactor-core/blob/master/reactor-core/src/main/java/reactor/core/publisher/Flux.java#L5767) and [`FluxMapFuseable`](https://github.com/reactor/reactor-core/blob/master/reactor-core/src/main/java/reactor/core/publisher/FluxMapFuseable.java#L89).

There are some exceptions where we did only one implementation of a `Publisher`, which checks the `Subscription` to be a `QueueSubscription` at the point where fusion mode is negotiated by the downstream, but that is at the expense of making said operator always appear "fuseable" (when sometimes it will just answer `Fuseable.NONE` to _all_ `requestFusion` calls).
For example, [`FluxElapsed`](https://github.com/reactor/reactor-core/blob/master/reactor-core/src/main/java/reactor/core/publisher/FluxElapsed.java#L125)

We are thinking about removing that assumption (by either switching all operators' subscriptions to a `QueueSubscription`-like interface and only rely on `requestFusion` OR having all fuseable subscribers have a second field to store a confirmed `QueueSubscription`), but that would need more careful thought and planning, and could end up a major architectural change.

simonbasle: TL;DR of the above: the `ClassCastException` happens because most operators will assume that if the source it has is a `Fuseable`, the `Subscription`s it will receive are `QueueSubscription`s.

Issue #1027, opened 2018-07-12
Improve support for reactor : Record span start at subscription phase, end after termination rather then invocation phase


kptfh:
Currently, if I annotate method that returns Mono or Flux with NewSpan annotation,
span started and finished before actual subscription happened.
But I'd like to track span start on subscription and span finish on termination.

kptfh:
Actually I can introduce tests that are broken and fix that eliminates this issue :)

marcingrzejszczak:
Awesome, be our guest!

kptfh:
[https://github.com/spring-cloud/spring-cloud-sleuth/pull/1028](https://github.com/spring-cloud/spring-cloud-sleuth/pull/1028)

kptfh:
BTW, noticed that there is [TraceReactorConfiguration ](https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/TraceReactorAutoConfiguration.java) related to reactor support, that register reactor hooks.
Can you explain briefly or give a link to doc how it expected to work and why there are to strategies: ConditionalOnNotWebApplication and ConditionalOnWebApplication

robotmrv:
@kptfh, what about the case:
I want to log method invocation and it's input arguments and do it by aspect or manually inside a method body or just log smth inside it. 
If you start span only on subscribe then my Logger would get MDC from parent Span and I lose context.
And it is not desired result

kptfh:
Agree. Seems we need 2 spans to be created: sync and reactive. One will work on invocation phase and one on subscribe/completion. What about: 
1) append "_reactor" to reactive span or 
2) just add "reactive" tag to reactive span

marcingrzejszczak:
Hi! We haven't yet taken a look into this PR with @adriancole . Reactor instrumentation is very non trivial and before we add any new tags / features to it we need to very carefully analyze all changes.

kptfh:
@marcingrzejszczak  To finish with reactor stuff we need to consider @robotmrv comment. 
Currently I use one span for to not intersecting periods: 
- method invocation (construction phase) (span not started yet but set in scope)
- subscription-completion
It will be correct to have 2 different spans for this.

marcingrzejszczak:
Wdyt @adriancole?

PR #1028, opened 2018-07-12
Improve support for reactor #1027

kptfh:
[https://github.com/spring-cloud/spring-cloud-sleuth/issues/1027](https://github.com/spring-cloud/spring-cloud-sleuth/issues/1027)

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
@@ -206,7 +206,9 @@ class SleuthInterceptor implements IntroductionInterceptor, BeanFactoryAware  {
 	private Object proceedUnderSynchronousSpan(
 			MethodInvocation invocation, NewSpan newSpan, ContinueSpan continueSpan) throws Throwable {
 		Span span = tracer().currentSpan();
-		if (newSpan != null || span == null) {
+		//in case of @ContinueSpan and no span in tracer we start new span and should close it on completion
+		boolean startNewSpan = newSpan != null || span == null;
+		if (startNewSpan) {
 			span = tracer().nextSpan();
 			newSpanParser().parse(invocation, newSpan, span);
 			span.start();
@@ -220,16 +222,17 @@ class SleuthInterceptor implements IntroductionInterceptor, BeanFactoryAware  {
 			onFailure(span, log, hasLog, e);
 			throw e;
 		} finally {
-			after(span, newSpan != null, log, hasLog);
+			after(span, startNewSpan, log, hasLog);
 		}
 	}
 
 	private Object proceedUnderReactorSpan(
 			MethodInvocation invocation, NewSpan newSpan, ContinueSpan continueSpan) throws Throwable{
-		boolean isNewSpan = newSpan != null;
 		Span spanPrevious = tracer().currentSpan();
+		//in case of @ContinueSpan and no span in tracer we start new span and should close it on completion
+		boolean startNewSpan = newSpan != null || spanPrevious == null;
 		Span span;
-		if (isNewSpan || spanPrevious == null) {
+		if (startNewSpan) {
 			span = tracer().nextSpan();
 			newSpanParser().parse(invocation, newSpan, span);
 		} else {
@@ -244,7 +247,7 @@ class SleuthInterceptor implements IntroductionInterceptor, BeanFactoryAware  {
 			Publisher<?> publisher = (Publisher) invocation.proceed();
 
 			Mono<Span> startSpan = Mono.defer(() -> withSpanInScope(span, () -> {
-				if (isNewSpan || spanPrevious == null) {
+				if (startNewSpan) {
 					span.start();
 				}
 
@@ -255,12 +258,12 @@ class SleuthInterceptor implements IntroductionInterceptor, BeanFactoryAware  {
 			if(publisher instanceof Mono){
 				return startSpan.flatMap(spanStarted -> ((Mono<?>)publisher)
 						.doOnError(onFailureReactor(log, hasLog, spanStarted))
-						.doOnTerminate(afterReactor(isNewSpan, log, hasLog, spanStarted)));
+						.doOnTerminate(afterReactor(startNewSpan, log, hasLog, spanStarted)));
 			}
 			else if(publisher instanceof Flux){
 				return startSpan.flatMapMany(spanStarted -> ((Flux<?>)publisher)
 						.doOnError(onFailureReactor(log, hasLog, spanStarted))
-						.doOnTerminate(afterReactor(isNewSpan, log, hasLog, spanStarted)));
+						.doOnTerminate(afterReactor(startNewSpan, log, hasLog, spanStarted)));
 			}
 			else {
 				throw new IllegalArgumentException("Unexpected type of publisher: "+publisher.getClass());
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
@@ -276,7 +276,7 @@ class SleuthInterceptor implements IntroductionInterceptor, BeanFactoryAware  {
 
 	private Runnable afterReactive(boolean isNewSpan, String log, boolean hasLog, Span span) {
 		return () -> {
-			try(Tracer.SpanInScope ws1 = tracer().withSpanInScope(span)) {
+			try(Tracer.SpanInScope ws = tracer().withSpanInScope(span)) {
 				after(span, isNewSpan, log, hasLog);
 			}
 		};
@@ -284,7 +284,7 @@ class SleuthInterceptor implements IntroductionInterceptor, BeanFactoryAware  {
 
 	private Consumer<Throwable> onFailureReactive(String log, boolean hasLog, Span span) {
 		return throwable -> {
-			try(Tracer.SpanInScope ws1 = tracer().withSpanInScope(span)) {
+			try(Tracer.SpanInScope ws = tracer().withSpanInScope(span)) {
 				onFailure(span, log, hasLog, throwable);
 			}
 		};
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
@@ -209,7 +209,7 @@ class SleuthInterceptor implements IntroductionInterceptor, BeanFactoryAware  {
 		if (newSpan != null || span == null) {
 			span = tracer().nextSpan();
 			newSpanParser().parse(invocation, newSpan, span);
-			span = span.start();
+			span.start();
 		}
 		String log = log(continueSpan);
 		boolean hasLog = StringUtils.hasText(log);
@@ -244,15 +244,12 @@ class SleuthInterceptor implements IntroductionInterceptor, BeanFactoryAware  {
 			Publisher<?> publisher = (Publisher) invocation.proceed();
 
 			Mono<Span> startSpan = Mono.defer(() -> withSpanInScope(span, () -> {
-				Span spanStarted;
 				if (isNewSpan || spanPrevious == null) {
-					spanStarted = span.start();
-				} else {
-					spanStarted = span;
+					span.start();
 				}
 
-				before(invocation, spanStarted, log, hasLog);
-				return Mono.just(spanStarted);
+				before(invocation, span, log, hasLog);
+				return Mono.just(span);
 			}));
 
 			if(publisher instanceof Mono){
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
@@ -16,17 +16,13 @@
 
 package org.springframework.cloud.sleuth.annotation;
 
-import java.lang.annotation.Annotation;
-import java.lang.reflect.Method;
-import java.util.concurrent.atomic.AtomicBoolean;
-import javax.annotation.PostConstruct;
-
 import brave.Span;
 import brave.Tracer;
 import org.aopalliance.aop.Advice;
 import org.aopalliance.intercept.MethodInvocation;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
+import org.reactivestreams.Publisher;
 import org.springframework.aop.ClassFilter;
 import org.springframework.aop.IntroductionInterceptor;
 import org.springframework.aop.Pointcut;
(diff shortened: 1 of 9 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
```diff
@@ -143,18 +146,13 @@ public AnnotationMethodsResolver(Class<? extends Annotation> annotationType) {
...
-						@Override
-						public void doWith(Method method) throws IllegalArgumentException,
-								IllegalAccessException {
-							if (found.get()) {
-								return;
-							}
-							Annotation annotation = AnnotationUtils.findAnnotation(method,
-									SleuthAdvisorConfig.AnnotationMethodsResolver.this.annotationType);
-							if (annotation != null) { found.set(true); }
+			ReflectionUtils.doWithMethods(clazz, method -> {
```
adriancole: I assume there is no real change here, just beauty. correct me if wrong

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
```diff
@@ -183,44 +181,141 @@ public Object invoke(MethodInvocation invocation) throws Throwable {
 		if (method == null) {
 			return invocation.proceed();
 		}
+
 		Method mostSpecificMethod = AopUtils
-				.getMostSpecificMethod(method, invocation.getThis().getClass());
+				.getMostSpecificMethod(invocation.getMethod(), invocation.getThis().getClass());
```
adriancole: revert as I think this effectively is same as earlier right?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
adriancole: I assume there is no real change here, just beauty. correct me if wrong

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
```diff
@@ -183,44 +181,141 @@ public Object invoke(MethodInvocation invocation) throws Throwable {
...
 		NewSpan newSpan = SleuthAnnotationUtils.findAnnotation(mostSpecificMethod, NewSpan.class);
 		ContinueSpan continueSpan = SleuthAnnotationUtils.findAnnotation(mostSpecificMethod, ContinueSpan.class);
 		if (newSpan == null && continueSpan == null) {
 			return invocation.proceed();
 		}
-		Span span = tracer().currentSpan();
-		if (newSpan != null || span == null) {
-			span = tracer().nextSpan().start();
+
+		Class<?> returnType = method.getReturnType();
```
adriancole: nit: returnType isn't used later, you can inline as method.getReturnType isn't clarified with the variable name

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
```diff
@@ -183,44 +181,141 @@ public Object invoke(MethodInvocation invocation) throws Throwable {
...
+		if (newSpan != null || spanPrevious == null) {
+			span = tracer().nextSpan();
 			newSpanParser().parse(invocation, newSpan, span);
+			span = span.start();
+		} else {
+			span = spanPrevious;
 		}
 		String log = log(continueSpan);
 		boolean hasLog = StringUtils.hasText(log);
 		try (Tracer.SpanInScope ws = tracer().withSpanInScope(span)) {
```
adriancole: if spanPrevious is not null, this scoping is redundant

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
```diff
@@ -183,44 +181,141 @@ public Object invoke(MethodInvocation invocation) throws Throwable {
...
+						}));
 			}
+			else {
+				throw new IllegalArgumentException("Unexpected type of publisher: "+publisher.getClass());
+			}
+		}
+	}
+
+	private void before(MethodInvocation invocation, Span span, String log, boolean hasLog) {
+		if (hasLog) {
```
adriancole: these annotations will be costly to index, are we sure we want them? Also these should be named "event" or similar, not log

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
```diff
@@ -183,44 +181,141 @@ public Object invoke(MethodInvocation invocation) throws Throwable {
...
+
+		String log = log(continueSpan);
+		boolean hasLog = StringUtils.hasText(log);
+
+		try(Tracer.SpanInScope ws = tracer().withSpanInScope(span)) {
+
+			Publisher<?> publisher = (Publisher) invocation.proceed();
+
+			Mono<Span> startSpan = Mono.defer(() -> {
+				try(Tracer.SpanInScope ws1 = tracer().withSpanInScope(span)) {
```
adriancole: there's a fair amount of code complexity (you can tell by indent levels etc). Should be cleaned up.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
```diff
@@ -183,44 +181,141 @@ public Object invoke(MethodInvocation invocation) throws Throwable {
...
+					Span spanStarted;
```
adriancole: this is a shell game, the reference doesn't change when start is called.. just call start

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
kptfh: correct. Don't you mind about some beauty?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
kptfh: inlined

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
kptfh: fixed

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
kptfh: then better refactor Span class to `void start()`. Current interface tells me to use returned value. There is probability that some implementation of `start()` may return another instance of Span.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
kptfh: Not clear for me as it's just safe refactoring(extracted code to method to not duplicate it). All code and variable names are kept unchanged.

adriancole:
I understand your point about the Span methods returning Span, but if I
recall the docs say this is a convenience for chaining. I can clarify the
docs that the same instance is always returned.

breaking api just to make this clear not only hurts usability but also will
break every code out there. doc is better choice I hope you agree.

adriancole:
I don't mind beauty just asking. because if you say just beauty I can
ignore the code and focus elsewhere. I like beauty actually ;)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
kptfh: fixed

kptfh:
Quite strange why build have failed. Have run successfully all the tests in the project locally. Debugged this failed tests, it even doesn't step into SleuthInterceptor as there is no `@NewSpan` or `@ContinueSpan` annotations in this test.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
```diff
@@ -143,19 +148,14 @@ public AnnotationMethodsResolver(Class<? extends Annotation> annotationType) {
...
-								IllegalAccessException {
-							if (found.get()) {
-								return;
-							}
-							Annotation annotation = AnnotationUtils.findAnnotation(method,
-									SleuthAdvisorConfig.AnnotationMethodsResolver.this.annotationType);
-							if (annotation != null) { found.set(true); }
-						}
-					});
+			ReflectionUtils.doWithMethods(clazz, method -> {
```
adriancole: for next time.. make a comment when you have blocks of code that are just reformatting. Ex in the PR, make a line comment like "This is just reformatting.. same behaviour" Doing so allows reviewers to skip down and ignore it.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
```diff
@@ -183,44 +183,138 @@ public Object invoke(MethodInvocation invocation) throws Throwable {
...
+
+		if(isReactiveReturnType(method.getReturnType())){
+			return proceedUnderSpanReactively(invocation, newSpan, continueSpan);
+		} else {
+			return proceedUnderSpanSynchronously(invocation, newSpan, continueSpan);
+		}
+	}
+
+	private boolean isReactiveReturnType(Class<?> returnType) {
+		return Flux.class.equals(returnType) || Mono.class.equals(returnType);
```
adriancole: Write a javadoc or comment about how we know this is the universe of return types. For example, This implies there are no subtypes of Flux or Mono, and also that there are no other things such as Flowable that can be returned.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
```diff
@@ -183,44 +183,138 @@ public Object invoke(MethodInvocation invocation) throws Throwable {
...
+			after(span, newSpan != null, log, hasLog);
+		}
+	}
+
+	private Object proceedUnderSpanReactively(
+			MethodInvocation invocation, NewSpan newSpan, ContinueSpan continueSpan) throws Throwable{
+		boolean isNewSpan = newSpan != null;
+		Span spanPrevious = tracer().currentSpan();
+		Span span;
+		if (isNewSpan || spanPrevious == null) {
```
adriancole: make a comment mentioning how when spanPrevious was null, we somehow know to close it later

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
kptfh: renamed so now reactive scope is clear from the name

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
kptfh: ok

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
kptfh: Can you help me to clarify some points in original version of code?
In case of `@ContinueSpan` and span is not set before in tracer then new span will be created and not closed on method completion. 
In tests this case is not covered.
Is it by design?

adriancole: I think the case of continuespan we don't close. that's the main thing I think we should have a unit test about now. Correct me if I am wrong @marcingrzejszczak

kptfh: @adriancole @marcingrzejszczak please, check my last commit. I think it should fix issue with not closed spans on `@ContinueSpan`

marcingrzejszczak:
We don't want to close a span when continue span annotation is used.

marcingrzejszczak:
Congrats @kptfh on your first contribution :)

spring-cloud-sleuth-zipkin/pom.xml
```diff
@@ -156,6 +156,16 @@
...
 			<scope>test</scope>
 		</dependency>
+		<dependency>
+			<groupId>io.projectreactor</groupId>
+			<artifactId>reactor-core</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.reactivestreams</groupId>
+			<artifactId>reactive-streams</artifactId>
```
artembilan: We don't need this dependency. It is transitive from the `reactor-core`: https://github.com/reactor/reactor-core/blob/master/build.gradle#L247

The `SleuthAdvisorConfig` is not in `test`. Not sure how does it work since `reactor-core` is in the `test` scope 🤷‍♂️ .

Also you need to consider to have `reactor-core` as a `optional` dependency and move all the Reactor stuff to its own auto-config based on the `Flux` presence condition.

Does it make sense for you?

kptfh: > Also you need to consider to have reactor-core as a optional dependency and move all the Reactor stuff to its own auto-config based on the Flux presence condition.

@artembilan  Can you create new issue? It may take some time to refactor.

artembilan: There is one already: https://github.com/spring-cloud/spring-cloud-sleuth/issues/1074

Exactly that one drove me to take a look into your code.
So, you can consider to refactor and fix at the same time relying on that issue.

kptfh: Those issue should be fixed just by removing test scope. Don't see easy way how to make it optional again. Need to reach agreement in key concepts with contributors to refactor it. So new issue will be right place for this.

marcingrzejszczak: I fixed the issue https://github.com/spring-cloud/spring-cloud-sleuth/commit/b13f15f9d7db49a0cda1e6777e01d0ad3deb61d8

PR #1034, opened 2018-07-24
Adjusts expectations of data reporting of loopback spans

adriancole:
The latest version of Brave corrects a data merge problem where shared
spans (created via Tracer.joinSpan) would lose some data. Along the way,
this fixes some misuse of `Tracer.joinSpan` which should only be used
for RPC server spans.

See https://github.com/openzipkin/brave/pull/734 applied in Brave 5.1.3


adriancole:
cc @zeagord @shakuzen

PR #1035, opened 2018-07-24
ability to disable instrumentation in ThreadPoolTaskExecutor

jalogar:
We have a spring-boot application (spring-boot-starter-parent-2.0.0.RELEASE) using spring-cloud-starter-zipkin for writing "spans" to zipkin.

We use spring-integration too (through spring-boot-starter-integration) and we have added an integration flow with a PollableChannel to be used within a poller:

```
@Bean
public PollableChannel pollableChannel() {
    return new QueueChannel(100);
}

@Bean
@ServiceActivator(poller = @Poller(taskExecutor="batchTaskExecutor"), 
                  inputChannel= "pollableChannel")
public MyHandler myHandler() {
    return new MyHandler();
}
```

Since adding this configuration, we are having an "asynch" span every second. I raised a question in stackoverflow -> https://stackoverflow.com/questions/51438273/zipkin-asynch-span-every-second. So after investigating the issue, it turned out the span comes from the ThreadPoolTaskExecutor that the @Poller uses to check if there are new items.

I've made an improvement to disable instrumentation of ThreadPoolTaskExecutor based on the name of the executor. You can use spring.sleuth.async.ignoredBeans to configure a list of executors without instrumentation.

For instance, in the previous example:

```
spring:
  sleuth:
    async:
      ignoredBeans: batchTaskExecutor 
```

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/AsyncAutoConfiguration.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/AsyncAutoConfiguration.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright 2013-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.sleuth.instrument.async;
+
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.scheduling.annotation.AsyncConfigurer;
+
+/**
+ * {@link org.springframework.boot.autoconfigure.EnableAutoConfiguration Auto-configuration}
+ * that wraps an existing custom {@link AsyncConfigurer} in a {@link LazyTraceAsyncCustomizer}
+ *
+ * @author Jesus Alonso
+ * @since 2.0.1
+ */
+
+@Configuration
+@EnableConfigurationProperties(AsyncProperties.class)
+public class AsyncAutoConfiguration {
+}
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/AsyncAutoConfiguration.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/AsyncAutoConfiguration.java
@@ -25,7 +25,7 @@ import org.springframework.scheduling.annotation.AsyncConfigurer;
  * that wraps an existing custom {@link AsyncConfigurer} in a {@link LazyTraceAsyncCustomizer}
  *
  * @author Jesus Alonso
- * @since 2.0.1
+ * @since 2.1.0
  */
 
 @Configuration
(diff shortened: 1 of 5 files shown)


pivotal-issuemaster:
@jalogar Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=1035)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=1035) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@jalogar Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=1035)!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/AsyncAutoConfiguration.java
```diff
@@ -0,0 +1,34 @@
...
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.scheduling.annotation.AsyncConfigurer;
+
+/**
+ * {@link org.springframework.boot.autoconfigure.EnableAutoConfiguration Auto-configuration}
+ * that wraps an existing custom {@link AsyncConfigurer} in a {@link LazyTraceAsyncCustomizer}
+ *
+ * @author Jesus Alonso
+ * @since 2.0.1
```
marcingrzejszczak: It's a new feature, set it to `2.1.0`

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/ExecutorBeanPostProcessor.java
```diff
@@ -81,7 +83,12 @@ public Object postProcessAfterInitialization(Object bean, String beanName)
 			boolean classFinal = Modifier.isFinal(bean.getClass().getModifiers());
 			boolean cglibProxy = !classFinal;
 			ThreadPoolTaskExecutor executor = (ThreadPoolTaskExecutor) bean;
-			return createThreadPoolTaskExecutorProxy(bean, cglibProxy, executor);
+			AsyncProperties asyncProperties = asyncConfigurationProperties();
```
marcingrzejszczak: We need a test for that

jalogar: I can add the test. What part do you want to exactly test?

marcingrzejszczak:
```
given: a `spring.sleuth.async.ignoredBeans: fooExecutor `property was set
and: a `fooExecutor` bean exists
when: `fooExecutor` is executed
then: there is no span reported to zipkin
```

marcingrzejszczak:
The build is broken due to breaking checkstyle rules. Have you built the project locally?

jalogar:
I've fix them locally. I'll push the changes together with the unit test

marcingrzejszczak:
Congrats on your first contribution @jalogar !

jalogar:
Hello @marcingrzejszczak , would it be possible to add this commit to the 2.0.x branch? 

Regards,
Jesús

marcingrzejszczak:
Yeah, why not. Feel free to send a pr. Also please update the documentation.

adriancole:
In the docs please put a short example. like...

```
By default, any tasks started with a scheduler starts a new trace. This is to help understand the root cause of distributed work. Some polling loops do not create distributed work, and can cause "junk traces" in search results. To prevent an executor from starting new traces, set blah.blah. For example, you can disable creating spans for spring integration pollers by excluding the bean name: `batchTaskExecutor`
```

adriancole:
I'm ok with this, though not sure if this will lead to another change later.. ex batchTaskExecutor, but only exclude poller tasks :D

Issue #1033, opened 2018-07-23
Add a filter that allows to drop spans


marcingrzejszczak:
Right now we allow to mutate a Zipkin Span before we send it to Zipkin. Sometimes, however, it could be useful to filter out certain spans.

Adding a filter (`SpanFilter`?) before the `SpanAdjuster`  would allow to filter out undesired spans.

**IMPORTANT** ONE MUST NOT DROP SPANS THAT HAVE CHILDREN!

adriancole:
used with care this can be helpful until there's another way to do the same.

adriancole:
technically someone can do this now by wrapping the reporter.

adriancole:
See also https://github.com/openzipkin/zipkin/pull/2072 which is server-side

adriancole:
See https://github.com/spring-cloud/spring-cloud-sleuth/pull/1096

marcingrzejszczak:
Closing in favour of https://github.com/spring-cloud/spring-cloud-sleuth/pull/1096

PR #1037, opened 2018-07-25
Added SpanFilter

marcingrzejszczak:
without this change we cannot filter out spans to be sent to Zipkin
with this change we give users a hook to achieve that

fixes gh-1033


adriancole:
I really like this, but I think firehose mode in brave will land before sleuth 2.1. Let's try to do what we need with this. https://github.com/openzipkin/brave/issues/557

Probably will be something like..

```
boolean beforeReport(context, spanRecord);
```

or something where you can skip out like you are now, but you can also mutate the record. This would be invoked regardless of whether the data goes to something like micrometer or zipkin

marcingrzejszczak:
Dropping in favour of Brave changes

Issue #1038, opened 2018-07-25
Bug report. Span not close when ClientAbortException: java.io.IOException: Broken pipe


mm23504570:
For some reason  throw the org.apache.catalina.connector.ClientAbortException: java.io.IOException: Broken pipe(ex. client timeout).  The TraceFilter not close the span.
I have a sample application to test it. The /normal-request return a normal content and httpstatus is 200. 
The ClientAbortException throw from AbstractJackson2HttpMessageConverter.writeInternal. Then exception propagate to TraceFilter. After that, In detachOrCloseSpans 
```
TraceFilter:
if (httpStatusSuccessful(response) && tracer().isTracing()) {
				if (log.isDebugEnabled()) {
					log.debug("Closing the span " + span + " since the response was successful");
				}
				if (exception == null || !hasErrorController()) {
					tracer().close(span);
					clearTraceAttribute(request);
				}
			}
```
The status is 200 and exception is not null, cause span not close.  So the exception propagate to StandardHostValve.invoke

```
StandardHostValve: 

protected void throwable(Request request, Response response,
                             Throwable throwable) {
        ......
        // If this is an aborted request from a client just log it and return
        if (realError instanceof ClientAbortException ) {
            if (log.isDebugEnabled()) {
                log.debug
                    (sm.getString("standardHost.clientAbort",
                        realError.getCause().getMessage()));
            }
            return;
        }
        ....
```
Tomcat ignore the ClientAbortException. So that request not forward to the /error.
After all, Span have no chance to close.

### Version
spring-boot-starter-parent 1.5.14.RELEASE
spring-cloud.version Edgware.SR4

### My Workaround 
use HandlerExceptionResolver to handle this. Then return normal ModelAndView. But this bug mabye need fix.

mm23504570:
[test-duplicate-trace-id.zip](https://github.com/spring-cloud/spring-cloud-sleuth/files/2227577/test-duplicate-trace-id.zip)

marcingrzejszczak:
Care for a PR ?

mm23504570:
Ok. I'm trying to fix it.

marcingrzejszczak:
Fixed via https://github.com/spring-cloud/spring-cloud-sleuth/commit/707f3714d60bf280133b43bbfbc2ef84e1a2455f

addozhang:
[...]

PR #1040, opened 2018-07-26
Close span when Exception is ClientAbortException

mm23504570:
Fixes gh-1038

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
@@ -21,10 +21,12 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.regex.Pattern;
-import javax.servlet.*;
+import javax.servlet.FilterChain;
+import javax.servlet.ServletException;
+import javax.servlet.ServletRequest;
+import javax.servlet.ServletResponse;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
-
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.springframework.beans.factory.BeanFactory;
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
@@ -21,10 +21,7 @@ import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.regex.Pattern;
-import javax.servlet.FilterChain;
-import javax.servlet.ServletException;
-import javax.servlet.ServletRequest;
-import javax.servlet.ServletResponse;
+import javax.servlet.*;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
 
@@ -241,6 +238,11 @@ public class TraceFilter extends GenericFilterBean {
 				if (exception == null || !hasErrorController()) {
 					tracer().close(span);
 					clearTraceAttribute(request);
+				} else if (exception != null) {
+					if(isClientAbortExpcetion(exception)){
+						tracer().close(span);
+						clearTraceAttribute(request);
+					}
 				}
 			} else if (errorAlreadyHandled(request) && tracer().isTracing() && !shouldCloseSpan(request)) {
 				if (log.isDebugEnabled()) {
@@ -269,6 +271,13 @@ public class TraceFilter extends GenericFilterBean {
 				}
 			}
 		}
+	}
+
+	/**
+	 * check exception is org.apache.catalina.connector.ClientAbortException.
+	 */
+	private boolean isClientAbortExpcetion(Throwable exception) {
+		return exception.getClass().getName().equals("org.apache.catalina.connector.ClientAbortException");
 	}
 
 	// null check is only for tests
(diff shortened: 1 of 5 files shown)


pivotal-issuemaster:
@mm23504570 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=1040)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=1040) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@mm23504570 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=1040)!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
```diff
@@ -21,10 +21,7 @@
 import java.util.Collections;
 import java.util.Enumeration;
 import java.util.regex.Pattern;
-import javax.servlet.FilterChain;
-import javax.servlet.ServletException;
-import javax.servlet.ServletRequest;
-import javax.servlet.ServletResponse;
+import javax.servlet.*;
```
marcingrzejszczak: We prefer explicit import statements

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
```diff
@@ -271,6 +273,13 @@ private void detachOrCloseSpans(HttpServletRequest request,
 		}
 	}
 
+	/**
+	 * check exception is org.apache.catalina.connector.ClientAbortException.
+	 */
+	private boolean isClientAbortExpcetion(Throwable exception) {
```
marcingrzejszczak: I don't like this. What if some other exceptions will be thrown? Will we have a list of exceptions for which we close the spans? I think we have to think of a different way of solving this problem.

adriancole: also typo. The javadoc should say why we are special-casing.. or a comment where we are special casing

mm23504570: @marcingrzejszczak I will create a class that holds a list of exceptions that let the user configure which exceptions should be closed in this case. 
What do you think?

adriancole:
seems if something like this were the case, it could be auto-configured on
presence of the tomcat class in question? (ex undertow should not need to
be bothered...) This is ideals, not saying the solution needs to be ideal

mm23504570:
Good. I like your solution, it is indeed more generic and graceful. I will try to implement it. Thanks.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceFilter.java
```diff
@@ -273,11 +272,14 @@ private void detachOrCloseSpans(HttpServletRequest request,
 		}
 	}
 
-	/**
-	 * check exception is org.apache.catalina.connector.ClientAbortException.
-	 */
-	private boolean isClientAbortExpcetion(Throwable exception) {
-		return exception.getClass().getName().equals("org.apache.catalina.connector.ClientAbortException");
+
+	private boolean needIgnoreExpcetion(Throwable exception) {
```
marcingrzejszczak: typo :(

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/TraceFilterTests.java
```diff
@@ -386,6 +392,9 @@ public void closesSpanWhenResponseStatusIs2xxAndExceptionIsClientAbortException(
 				.header(Span.TRACE_ID_NAME, 20L).buildRequest(new MockServletContext());
 		TraceFilter filter = new TraceFilter(beanFactory());
 		BDDMockito.given(beanFactory.getBean(ErrorController.class)).willReturn(() -> "/error");
+		List<ExceptionToIgnoreInTraceFilter> filters = Lists.newArrayList(getClientAbortExpcetionToIgnoreInTraceFilter());
```
marcingrzejszczak: I think we should add one test with a filter that throws the `ClientAbortException` and we would like to ensure that the span got closed and reported. WDYT?

mm23504570: like this?
``` java
@Test
public void closesSpanWhenResponseStatusIs2xxAndClientAbortExceptionThrowAfterTraceFilter() throws Exception {
	this.request = builder().header(Span.SPAN_ID_NAME, PARENT_ID)
			.header(Span.TRACE_ID_NAME, 20L).buildRequest(new MockServletContext());
	TraceFilter filter = new TraceFilter(beanFactory());
	BDDMockito.given(beanFactory.getBean(ErrorController.class)).willReturn(() -> "/error");
	List<ExceptionToIgnoreInTraceFilter> filters = Lists.newArrayList(getClientAbortExpcetionToIgnoreInTraceFilter());
	BDDMockito.given(beanFactory.getBean(ExceptionToIgnoreInTraceFilterProvider.class))
			.willReturn(getExceptionToIgnoreInTraceFilterProvider(filters));
	this.response = new MockHttpServletResponse();
	response.setStatus(200);
	this.filterChain = new MockFilterChain(){
		@Override
		public void doFilter(ServletRequest request, ServletResponse response) throws IOException, ServletException {
			throw new ClientAbortException();
		}
	};
	try {
		filter.doFilter(this.request, this.response, this.filterChain);
	}catch (ClientAbortException e){
		// ig
	}
	then(TestSpanContextHolder.getCurrentSpan()).isNull();
}
```

marcingrzejszczak: yeah and also

```java
	then(TestSpanContextHolder.getCurrentSpan()).isNull();
         // verify that the span got reported
         then(spanReporter.getSpans()).hasSize(1);
```

marcingrzejszczak:
Congrats @mm23504570 on your contribution! Awesome work.

Issue #1044, opened 2018-07-29
Enhancement: Support specification of span tag for method return value via annotation


mmanciop:
Spring Sleuth allows to specify span tags for parameters and error occurring when executing methods annotated with `@NewSpan`, but not for the return values resulting from the invocation of the respective methods.

The idea is to introduce a new `@SpanTagFromReturnValue`, applicable only at method level, with the same fields and very similar semantics to `@SpanTag`.

I think it is best to avoid "repurposing" `@SpanTag` for this use-case as well, adding to it the applicability to method level, in order to avoid confusion with trying to tag parameters and return values on the same method. (Also, `@SpanTagFromReturnValue` may be verbose, but hopefully it is pretty clear what one intends to do with it.)

In case the annotation `@SpanTagFromReturnValue` is applied to a `void` method, Spring Sleuth should ignore it: AFAIK there is no way to make this case fail at compile time, and logging errors or warnings would result in amazing, unwarranted log spam.

adriancole:
can you give an example of the buisiness code and the type of tag you are creating?

mmanciop:
For example, to persist data about size of datasets returned via JDBC queries (number of rows, total amount of bytes, etc.): it is all too common that code, old and new, bombs processes running it by downloading half the database because a developer has not thought of “going the extra mile” and use server-side pagination. Having that data in the spans is a much better option than in logs as it correlates directly (and often more than linearly!) with the amount of time necessary to process it.

adriancole:
in this case, wouldn't it be better to instrument the JDBC layer vs using a
proxy?

mmanciop:
If one has control of the JDBC layer, instrumenting that may be a good idea.

But it happens often that one just has to use a client library from a different team that does not lend itself to being instrumented in code by the users or the auhors (“sure, we could add the thing for you in 6 months, just hang one; and no, we do not have time to review pull requests and make a release of the code, we are too busy with backlog”).

And I am speaking here about JDBC because I have a few chips on my shoulder thanks to it. The same argument applies to pretty much any wrapper code around 3rd-party libs one does not control.

adriancole:
I get your point.

Does this need to be elevated right now to a new annotation? We would
expect multiple requests, right?

What would happen if you decorated the type again, between the tracing
interceptor, to add the tag yourself, using an annotation you define..
until there's more interest?

mmanciop:
[...]

PR #1045, opened 2018-07-29
Introduce @SpanTagFromReturnValue annotation #1044

mmanciop:
#1044


marcingrzejszczak:
Let's close this until more people get interested in this. I'm personally not convinced about this feature just yet.

Issue #1047, opened 2018-07-31
Skip patterns for outbound web requests 


kmoroe22:
Hi guys 

We have libraries in place which periodically send out health checks to a monitoring service. Because of this we tend to see exported traces for those calls. We can't however disable entire web client capability because we have other web clients with which we interact with which need to be monitored. Is there anything currently in place to address an issue like this?

PR #1048, opened 2018-08-01
Added an option to filter out outbound requests

marcingrzejszczak:
without this change there's no way to filter outbound requests basing on the path via a property.
with this change we're introducing that

fixes gh-1047

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHttpAutoConfiguration.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHttpAutoConfiguration.java
@@ -112,8 +112,7 @@ public class TraceHttpAutoConfiguration {
 	@Bean
 	@ConditionalOnMissingBean(name = ClientSampler.NAME)
 	HttpSampler sleuthClientSampler(SleuthWebProperties sleuthWebProperties) {
-		return new CompositeHttpSampler(new PathMatchingHttpSampler(
-				sleuthWebProperties), HttpSampler.TRACE_ID);
+		return new PathMatchingHttpSampler(sleuthWebProperties);
 	}
 }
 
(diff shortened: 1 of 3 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHttpAutoConfiguration.java
```diff
@@ -138,3 +139,19 @@ HttpSampler sleuthClientSampler() {
...
+	PathMatchingHttpSampler(SleuthWebProperties properties) {
+		this.properties = properties;
+	}
+
+	@Override public <Req> Boolean trySample(HttpAdapter<Req, ?> adapter, Req request) {
+		String path = adapter.path(request);
+		if (path == null) {
+			return null;
+		}
+		return path.matches(this.properties.getClient().getSkipPattern()) ? false : null;
```
shakuzen: If the skip pattern isn't set, does this NPE?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHttpAutoConfiguration.java
```diff
@@ -111,8 +111,9 @@ HttpServerParser defaultHttpServerParser() {
 
 	@Bean
 	@ConditionalOnMissingBean(name = ClientSampler.NAME)
-	HttpSampler sleuthClientSampler() {
-		return HttpSampler.TRACE_ID;
+	HttpSampler sleuthClientSampler(SleuthWebProperties sleuthWebProperties) {
+		return new CompositeHttpSampler(new PathMatchingHttpSampler(
```
shakuzen: What is the point in using the composite here since the right always defers? It seems equivalent to just using the `PathMatchingHttpSampler` directly.

marcingrzejszczak: Oh yeah, I haven't noticed that

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHttpAutoConfiguration.java
marcingrzejszczak: Good point - I'll fix it to empty string

PR #1053, opened 2018-08-09
improve performance : excessive check removed

kptfh:


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/SleuthAdvisorConfig.java
```diff
@@ -105,7 +105,7 @@ private Pointcut buildPointcut() {
 
 		@Override
 		public boolean matches(Method method, Class<?> targetClass, Object... args) {
-			return getClassFilter().matches(targetClass);
```
marcingrzejszczak: Doesn't it break any logic?

kptfh: No. Actual check takes place in SleuthInterceptor.invoke(MethodInvocation)

gavlyukovskiy: Doesn't it mean that all classes will be proxied regardless of existence of `NewSpan` or `ContinueSpan` annotations?

marcingrzejszczak: Yeah, that's what i was afraid of. But after looking at the code very briefly and following @kptfh suggestions, it seems that the advisor will apply the interceptor if there's an annotation on a class or a method. But maybe I'm wrong, like I said i haven't read the code thoroughly.

kptfh: No. Check for class remains unchanged `getClassFilter()`. Just remove duplicated check for method being annotated.

Issue #1030, opened 2018-07-14
@NewSpan does not generate a new span Id for methods using Mono.subscriberContext


nnanda2016:
I am using:
- `Finchley.RELEASE` release train
- Spring boot `2.0.3`

If I use `Mono.subscriberContext()` in a method and annotate the method with `@NewSpan(name="myMethod")`, no new span ID is generated for that method.

Reference Repo: https://github.com/nnanda2016/springboot-sleuth-issue 

The method `com.demo.springboot.sleuth.UserApiHandler#getById2` is the buggy one I am talking about. No new span Id is generated if I invoke this method (URL to invoke this: `curl -v http://localhost:5002/users/ids/U1`).

If I remove `Mono.subscriberContext()`, everything is working fine as expected. Example method in this sample app is `com.demo.springboot.sleuth.UserApiHandler#getById` (URL to invoke this: `curl -v http://localhost:5002/users/U1`).

kptfh:
@marcingrzejszczak  Investigating this issue found that there is a bug in my recent pull request (didn't set new span in trace context). Will add tests and fix soon.

kptfh:
@nnanda2016 Don't use Tracing.current().currentTraceContext().get().spanId() in your app.  You should tracer.currentSpan().context().spanId() instead.
[https://github.com/spring-cloud/spring-cloud-sleuth/pull/1056#issuecomment-412571580](https://github.com/spring-cloud/spring-cloud-sleuth/pull/1056#issuecomment-412571580)

nnanda2016:
@kptfh I checked the review comments on #1056 and I will change the code as per your suggestion; however, does this usage cause this issue?

marcingrzejszczak:
For sure it doesn't help ;) @kptfh has done some improvements in the way reactor stuff is handled so that should help

nnanda2016:
oh okay...where can i see those details? I want to make code change n verify.

marcingrzejszczak:
The whole discussion is happening in the PR (https://github.com/spring-cloud/spring-cloud-sleuth/pull/1056). The previous changes were done here https://github.com/spring-cloud/spring-cloud-sleuth/pull/1028

nnanda2016:
So, looks like I have to upgrade the version to `2.0.1`

marcingrzejszczak:
Try with the latest `2.1.0` snapshots

PR #1056, opened 2018-08-10
Set span in context

kptfh:
also should fix [@NewSpan does not generate a new span Id for methods using Mono.subscriberContext #1030](https://github.com/spring-cloud/spring-cloud-sleuth/issues/1030)


kptfh:
Have no idea why tests fails on CI environment. Locally they run without any errors. 
@marcingrzejszczak can you have a quick look or give a hint?

marcingrzejszczak:
```
Your job ran 346 tests with 4 failures
shouldReturnNewSpanFromSubscriberContext - org.springframework.cloud.sleuth.annotation.SleuthSpanCreatorAspectFluxTests
java.lang.NullPointerException
	at org.springframework.cloud.sleuth.annotation.SleuthSpanCreatorAspectFluxTests.shouldReturnNewSpanFromSubscriberContext(SleuthSpanCreatorAspectFluxTests.java:327)
shouldReturnNewSpanFromSubscriberContext - org.springframework.cloud.sleuth.annotation.SleuthSpanCreatorAspectMonoTests
java.lang.NullPointerException
	at org.springframework.cloud.sleuth.annotation.SleuthSpanCreatorAspectMonoTests.shouldReturnNewSpanFromSubscriberContext(SleuthSpanCreatorAspectMonoTests.java:351)
shouldReturnNewSpanFromTraceContext - org.springframework.cloud.sleuth.annotation.SleuthSpanCreatorAspectFluxTests
java.lang.NullPointerException
	at org.springframework.cloud.sleuth.annotation.SleuthSpanCreatorAspectFluxTests.shouldReturnNewSpanFromTraceContext(SleuthSpanCreatorAspectFluxTests.java:316)
```

kptfh:
Yeah, I saw this errors but was not able to reproduce them locally. Think that thay may be somehow related to environment.

kptfh:
@marcingrzejszczak  removed LastOperatorWrapper and broken tests become fixed and stable and no new tests failed. (I suspect ConditionalOnNotWebApplication to be unstable)
Can I broke something not covered with tests that LastOperatorWrapper was meant for?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/TraceReactorAutoConfiguration.java
```diff
@@ -94,10 +81,6 @@ static HookRegisteringBeanDefinitionRegistryPostProcessor traceHookRegisteringBe
 	}
 }
 
-interface LastOperatorWrapper {
```
marcingrzejszczak: Why have you removed that?

kptfh: Just trying to figure out the reason why tests fail on CI environment as I can't reproduce this on local env.
JFY: LastOperatorWrapper  code seems to be not covered by tests. When I removed this code no new tests failed. Don't you know what this code (LastOperatorWrapper) for?

marcingrzejszczak: The concept is such that if you have a non web application we want to apply a different last operator wrapper. It seems that we're missing a test for that. We should add it instead of removing this functionality

kptfh: Not going to merge this change just experimenting with build on CI. Can you point me to an issue or case that will describe why we need this additional hook on last operator for not web application?

marcingrzejszczak: For non web application we assume that if you're calling reactor's flux, you're doing it manually. In that case we want to close the span when the flux is finished.

In a web application we want a filter to close the flux or any other higher level component. We don't want multiple components try to star, close spans and report them to Zipkin.

kptfh: Don't know all the corner cases but it seems that with new Mono/Flux  `@NewSpan` and `@ConinueSpan` annotations processing span get started after subscription and closed on flux termination

marcingrzejszczak: So maybe you fixed it! In fact there's a test that should fail if something goes wrong https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/reactor/SpanSubscriberTests.java . Do you have a WebFlux based app which you can click to see that your changes don't break anything?  If not you can use this https://github.com/spring-cloud-samples/sleuth-documentation-apps

kptfh: Thanks! Replaced `Tracing.current().currentTraceContext().get().spanId()` 
with `tracer.currentSpan().context().spanId()` as in `SpanSubscriberTest` and get tests fixed !!!

kptfh:
And finally where Tracing.current().currentTraceContext().get().spanId() can be used? In my case it has some race condition issue and leads to inconsistency.

marcingrzejszczak:
I think that `Tracing.current().currentTraceContext().get().spanId()` should not be used at all. We should always use the bean.

marcingrzejszczak:
> @marcingrzejszczak Can you check my last commit. Trying to make test more strict replaced traceId with spanId. Does it make sense?

👍

kptfh:
@marcingrzejszczak  Should I add test with WebFlux based app in this PR?

marcingrzejszczak:
Just build it locally and double check on the sample. If that's fine just ping me back and we'll merge this

marcingrzejszczak:
Ok, can we merge this @kptfh ?

kptfh:
Have not checked on WebFlux app. Plan to write test.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
```diff
@@ -41,59 +40,6 @@
 
 	private static final Log log = LogFactory.getLog(ReactorSleuth.class);
 
-	/**
```
marcingrzejszczak: This looks like a breaking change... First, why have you removed it? Second if we can remove it, maybe just deprecate it first?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/SpanSubscriber.java
```diff
@@ -1,156 +0,0 @@
...
-import java.util.concurrent.atomic.AtomicBoolean;
-
-/**
- * A trace representation of the {@link Subscriber}
- *
- * @author Stephane Maldini
- * @author Marcin Grzejszczak
- * @since 2.0.0
- */
-final class SpanSubscriber<T> extends AtomicBoolean implements SpanSubscription<T> {
```
marcingrzejszczak: Why have you removed this?

kptfh:
@marcingrzejszczak added test

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/reactor/SpanSubscriberTests.java
```diff
@@ -23,10 +23,7 @@
 import brave.Tracing;
 import brave.sampler.Sampler;
 import org.junit.BeforeClass;
-import reactor.core.publisher.BaseSubscriber;
-import reactor.core.publisher.Flux;
-import reactor.core.publisher.Hooks;
-import reactor.core.publisher.Mono;
+import reactor.core.publisher.*;
```
marcingrzejszczak: Can you have explicit imports, please?

marcingrzejszczak:
We're almost there! I've added one question (this PR is big complicated so I want to be sure we're on the same page), and a small remark about import statement.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/SpanSubscriber.java
kptfh: SpanSubscriber was used only for lastOperation hook, which is not used anymore. Do you think it may be used outside sleuth? Should I revert and deprecate it?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
kptfh: SpanSubscriber was used only for lastOperation hook, which is not used anymore. It has package private access level. So it shouldn't be used out of sleuth project. But if you wish I can revert and deprecate it.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/SpanSubscriber.java
marcingrzejszczak: Ah ok. The class is indeed package private, but the static method calling it is public, isn't it? I would deprecate both the class and the static method and add a warning log statement that this method will be deleted in the next major release.

marcingrzejszczak:
Awesome work @kptfh!

Issue #1062, opened 2018-08-17
Allow customization of span names for Spring Messaging


saturnism:
https://github.com/spring-cloud/spring-cloud-gcp/issues/953

Spring Cloud Sleuth automatically instruments Spring Cloud Stream and Spring Integration when sending/processing messages.  When sending a message, the generated span name is always "send". This is not customizable like the HTTP span names.

Introduce a mechanism to allow easy span customization.

meltsufin:
@marcingrzejszczak When will this land in snapshots?

marcingrzejszczak:
Once we do it in Brave

marcingrzejszczak:
Also, @meltsufin, if you don't want to wait, you can copy what we've done in Sleuth, register your own version of `TraceChannelInterceptor` and inject a bean to allow customization of names.

meltsufin:
I can wait. :smile: Thanks!

adriancole:
we have limited horsepower. If someone wants to try to help with this, they can. It would need to be portable over jms, kafka, rabbit at least. Otherwise, tell people to use interceptors

marcingrzejszczak:
So with the new Sleuth & Stream approach this should be easier since you'll be able to access the span in your function or leverage the `MessagingSleuthOperator` utility class. I can help you apply either of those if you're interested @meltsufin @saturnism

spring-issuemaster:
If you would like us to look at this issue, please provide the requested information. If the information is not provided within the next 7 days this issue will be closed.

spring-issuemaster:
Closing due to lack of requested feedback. If you would like us to look at this issue, please provide the requested information and we will re-open the issue.

PR #1066, opened 2018-08-22
Message span customization added

marcingrzejszczak:
fixes gh-1062


adriancole:
I like the direction here, but the timing isn't perfect. We're literally working on messaging in brave and will need an abstract type (or at least attempt to have one) to avoid having N parsers ex for each library. Let's defer this and revisit. https://github.com/openzipkin/brave/issues/768

PR #1069, opened 2018-08-23
Fix #1065 : Wrong span name when request failing with a 401

### LINKED ISSUE #1065: Wrong span name when request failing with a 401
We're using spring cloud sleuth 2.01 with webflux and spring security. 
It all works fine except in case of authorization error, in this case the span name is `get` instead of `get /api/{cloudid}/etc` as expected.

**Possible cause:**
- `TraceWebFilter` uses `HandlerMapping.BEST_MATCHING_PATTERN_ATTRIBUTE` to generate the span nam
- `HandlerMapping.BEST_MATCHING_PATTERN_ATTRIBUTE` is set by `RequestMappingInfoHandlerMapping#handleMatch`
- unfortunately this is not called in case of authentication error because the request will be rejected by Spring Security filter.

spring-cloud-sleuth version: 2.0.1

madgnome:
This is more a proof of concept than an actual PR as I'm not convinced it's the proper way to fix this issue.

At the minimum the test case could be useful when providing a proper solution.


madgnome:
Too much change on `TraceWebFilter` to resolving conflict safely. I will close the PR for now.

PR #1078, opened 2018-08-29
Replace InMemoryReporterMetrics with No Op Reporter Metrics

worldtiki:
This was causing a memory leak in a production service where the Zipkin service was unavailable as InMemoryReporterMetrics kept incrementing the messagesDropped.

Since InMemoryReporterMetrics should be used for troubleshooting I think a no-op one is a more sensible default.

This is a trivial chance and I first want some feedback, so I'm not sure if signing a contributor license agreement is required.

### DIFF
--- a/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/ZipkinAutoConfiguration.java
+++ b/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/ZipkinAutoConfiguration.java
@@ -22,7 +22,6 @@ import brave.sampler.Sampler;
 import zipkin2.Span;
 import zipkin2.codec.BytesEncoder;
 import zipkin2.reporter.AsyncReporter;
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@worldtiki Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=1078)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=1078) to manually synchronize the status of this Pull Request.

devinsba:
I'm not one of the committers but I can say, I think this change will require it to be conditional on a property of some kind, where the current default stays the default. This seems like it's possibly a breaking change for 2.x but might be good to consider whenever 3.x comes along

worldtiki:
Issue fixed upstream in https://github.com/openzipkin/zipkin-reporter-java/commit/ba476a00d5985d7f74e42c21ab510fff95fc92d6

PR #1111, opened 2018-10-17
Reduce scoping overhead of WebFlux and adds Benchmarks

mm23504570:
Hi.
I'm starting to reduce the overhead of sleuth when tracing WebFlux server components.
I add some benchmarks use jmh then i noticed that changing the scope commands(in  ScopePassingTraceContextSpringWebFluxBenchmarks) from newScope to maybeScope cuts overhead in half.

### The report: 
image


benchmarks/src/main/java/org/springframework/cloud/sleuth/benchmarks/jmh/benchmarks/SpringWebFluxBenchmarks.java
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	@Benchmark
+	public void tracedClient_get_resumeTrace() throws Exception {
+		try (CurrentTraceContext.Scope scope = Tracing.current().currentTraceContext().newScope(defaultTraceContext)) {
+			get(tracedClient);
+		}
+	}
+
+//	@Bean
```
adriancole: remove the commented code as we don't check this in

benchmarks/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingTraceContextSubscriber.java
```diff
@@ -0,0 +1,110 @@
...
+import brave.propagation.TraceContext;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.reactivestreams.Subscriber;
+import org.reactivestreams.Subscription;
+import reactor.util.context.Context;
+
+import java.util.concurrent.atomic.AtomicBoolean;
+
+final class ScopePassingTraceContextSubscriber<T> extends AtomicBoolean implements SpanSubscription<T> {
```
adriancole: make a javadoc what this is replacing? I think before merge, we change the actual code

benchmarks/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingTraceContextHookConfig.java
```diff
@@ -0,0 +1,128 @@
...
+import reactor.core.publisher.Hooks;
+import reactor.core.publisher.Operators;
+import reactor.core.scheduler.Schedulers;
+import reactor.util.context.Context;
+
+import java.util.concurrent.ScheduledExecutorService;
+import java.util.function.Function;
+import java.util.function.Supplier;
+
+@Configuration
```
adriancole: make comment about what this doing differently? I think before merge, we change the actual code

adriancole:
@mm23504570 as far as I am concerned, you've proven the point. Assuming tests pass, I would remove the old implementation and replace it with yours

benchmarks/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingTraceContextHookConfig.java
```diff
@@ -0,0 +1,130 @@
...
+ */
+@Configuration
+public class ScopePassingTraceContextHookConfig {
+	@Bean
+	public ScopePassingTraceContextHookRegisteringBeanDefinitionRegistryPostProcessor hookRegisteringBeanDefinitionRegistryPostProcessor(ConfigurableApplicationContext context) {
+		return new ScopePassingTraceContextHookRegisteringBeanDefinitionRegistryPostProcessor(context);
+	}
+}
+
+@Slf4j
```
marcingrzejszczak: Let's not use Lombok please

benchmarks/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingTraceContextSubscriber.java
```diff
@@ -0,0 +1,116 @@
...
+				ctx != null ? ctx : Context.empty();
+		if (log.isTraceEnabled()) {
+			log.trace("Root span [" + root + "], context [" + this.context + "]");
+		}
+	}
+
+	@Override
+	public void onSubscribe(Subscription subscription) {
+		this.s = subscription;
+		CurrentTraceContext.Scope scope = this.currentTraceContext.maybeScope(root);
```
adriancole: you should be fine with try-with-resources

```suggestion
		try (CurrentTraceContext.Scope scope = this.currentTraceContext.maybeScope(root)) {
```

benchmarks/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingTraceContextSubscriber.java
```diff
@@ -0,0 +1,116 @@
...
+		try {
+			this.subscriber.onNext(o);
+		} finally {
+			scope.close();
+		}
+	}
+
+	@Override
+	public void onError(Throwable throwable) {
+		this.subscriber.onError(throwable);
```
adriancole: probably all of these should be scoped right @marcingrzejszczak ?

marcingrzejszczak: It should be the same as it was previously. All `onX` should be scoped

adriancole:
for those following along, the reason this cuts overhead is because scoping has a cost to it (especially MDC). the "maybeScope" feature peeks to see if a trace context is already current. So if you are scheduled to go onto the same thread, this eliminates the otherwise redundant thread local activities. If a lot of hops happen you can see it can reduce overhead quite a bit.

The benchmarks should be re-run after all onXXX callbacks are scoped just in case they are faster due to an oversight!

benchmarks/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingTraceContextHookConfig.java
```diff
@@ -0,0 +1,133 @@
...
+				return new ScopePassingTraceContextSubscriber(
+						sub,
+						sub != null ? sub.currentContext() : Context.empty(),
+						tracing);
+			}
+		};
+	}
+
+	private Schedulers.Factory factoryInstance(final BeanFactory beanFactory) {
+		return new Schedulers.Factory() {
```
adriancole: this looks like it could be a lambda? if so, make it one?

benchmarks/src/main/java/org/springframework/cloud/sleuth/benchmarks/jmh/benchmarks/WithOutSleuthSpringWebFluxBenchmarks.java
```diff
@@ -0,0 +1,47 @@
...
+	}
+
+	public static void main(String[] args) throws RunnerException {
+		Options opt = new OptionsBuilder()
+				.include(".*" + WithOutSleuthSpringWebFluxBenchmarks.class.getSimpleName() + ".*")
+				.build();
+
+		new Runner(opt).run();
+	}
+}
```
adriancole: nit try to make these have EOL before EOF

https://unix.stackexchange.com/questions/31947/how-to-add-a-newline-to-the-end-of-a-file

benchmarks/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingTraceContextHookConfig.java
mm23504570: It's count not be a lambda.

mm23504570:
In I was not from the latest master checkout the branch. This is latest test result. Terrible - -!

image

benchmarks/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingTraceContextHookConfig.java
```diff
@@ -0,0 +1,133 @@
...
+				return sourcePub;
+			}
+			//no more POINTCUT_FILTER since mechanism is broken
+			Function<? super Publisher<T>, ? extends Publisher<T>> lift = Operators.lift((scannable, sub) -> {
+				//rest of the logic unchanged...
+				if (beanFactory.isActive()) {
+					if (log.isTraceEnabled()) {
+						log.trace("Spring Context already refreshed. Creating a scope " + "passing span subscriber with Reactor Context " + "[" + sub.currentContext() + "] and name [" + scannable.name() + "]");
+					}
+					return scopePassingTraceContextSubscription(beanFactory, scannable, sub).get();
```
adriancole: seems more efficient to `return new ScopePassingTraceContextSubscriber(...` vs making a provider and immediately calling get.

benchmarks/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingTraceContextSubscriber.java
```diff
@@ -0,0 +1,107 @@
...
+	private final TraceContext root;
+	private Subscription s;
+
+	ScopePassingTraceContextSubscriber(Subscriber<? super T> subscriber, Context ctx, Tracing tracing) {
+		this.subscriber = subscriber;
+		this.tracing = tracing;
+		this.currentTraceContext = this.tracing.currentTraceContext();
+		TraceContext root = ctx != null ? ctx.getOrDefault(TraceContext.class, this.currentTraceContext.get()) : null;
+		this.root = root;
+		this.context = ctx != null && root != null ? ctx.put(Span.class, root) :
```
adriancole: change the context key from `Span.class` to `TraceContext.class` as that's what is stored

adriancole: ps where is this read? I don't see anything in this code reading ctx.. this might imply something is broke.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingSpanSubscriber.java
```diff
@@ -38,69 +40,71 @@
...
 			Tracing tracing) {
 		this.subscriber = subscriber;
-		this.tracer = tracing.tracer();
-		Span root = ctx != null ? ctx.getOrDefault(Span.class, this.tracer.currentSpan())
-				: null;
-		this.span = root;
-		this.context = ctx != null && root != null ? ctx.put(Span.class, root)
+		this.tracing = tracing;
+		this.currentTraceContext = this.tracing.currentTraceContext();
+		TraceContext root = ctx != null ? ctx.getOrDefault(TraceContext.class, this.currentTraceContext.get()) : null;
```
adriancole: I wonder if this is a little inefficient. `currentTraceContext.get()` will be called even if there is a TraceContext there.. is there a way to pass a method reference instead? otherwise, maybe get if it is null

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingSpanSubscriber.java
```diff
@@ -38,69 +40,71 @@
...
+		this.root = root;
+		this.context = ctx != null && root != null ? ctx.put(TraceContext.class, root)
 				: ctx != null ? ctx : Context.empty();
+
 		if (log.isTraceEnabled()) {
-			log.trace("Root span [" + root + "], context [" + this.context + "]");
+			log.trace("Root traceContext [" + root + "], context [" + this.context + "]");
```
adriancole:
```suggestion
			log.trace("Root context [" + root + "], this context [" + this.context + "]");
```

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingSpanSubscriber.java
```diff
@@ -46,65 +44,70 @@
...
-			Tracing tracing) {
+	                           Tracing tracing) {
 		this.subscriber = subscriber;
 		this.tracing = tracing;
+		Tracer tracer = tracing.tracer();
 		this.currentTraceContext = this.tracing.currentTraceContext();
-		TraceContext root = ctx != null ? ctx.getOrDefault(TraceContext.class, this.currentTraceContext.get()) : null;
-		this.root = root;
-		this.context = ctx != null && root != null ? ctx.put(TraceContext.class, root)
+		Span rootSpan = ctx != null ? ctx.getOrDefault(Span.class, tracer.currentSpan())
```
adriancole: hmm why are we putting both in context? both Span.class and TraceContext.class? Is there other code using one or the other? If so, please make comment to what that code is.. probably we should only do one as it is double overhead to do both.

adriancole: ex `tracer.currentSpan()` inside calls `currentTraceContext.get()`

adriancole:
Thanks very much @mm23504570! let's carry work further on the next pull request.

tony-clarke-amdocs:
@adriancole  Will this change be merged into spring cloud sleuth 2.0.3?

Issue #344, opened 2016-07-20
Introduce Reactor instrumentation


marcingrzejszczak:
with proper Reactor instrumentation we will improve the RxJava instrumentation too.

marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-sleuth/commit/3e7e54e09bd3dd41d9a0b3c717d0984832bfbcef

marcingrzejszczak:
I remove the reactor support for `1.3.0.M1` release (https://github.com/spring-cloud/spring-cloud-sleuth/commit/57b0ac0e5191e0cef2a4f588c6053acfef7e83b4). That's because of reactor version mismatch. We'll discuss what to do about this for `1.3.0.M2` release

sean-brandt:
Is this still going to make 1.3.0.M2? Very much in need of this support.

spencergibb:
Sweet!

marcingrzejszczak:
Thanks :D That's only for Edgware though. No Webflux support yet (it's on a branch waiting for better 2.0.x times ;) )

spencergibb:
Spring 5 and reactor 3.1 are GA, there won't be breaking changes the anymore

PR #1117, opened 2018-11-01
Add a SpanCustomizer that works on TraceWebFilter instrumented ServerWebExchanges

fitzoh:
I came across this attempting to add route id tags to a spring-cloud-gateway proxy, and couldn't figure out an easy way to get the current Span.
It's entirely possible that:
*  I don't understand how the current span is tracked
* `CurrentSpanCustomizer` works fine for that use case
* this is completely unnecessary
but... my understanding was that the existing span trackers use threadlocals, and aren't necessarily safe in a webflux environment.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/ServerWebExchangeSpanCustomizer.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/ServerWebExchangeSpanCustomizer.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright 2013-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


fitzoh:
So my thought was that this was just be an implementation of `SpanCustomizer` that works with `ServerWebExchange`s that something like `SpanCustomizingHandlerInterceptor` could be built off of (like something mentioned in #1120).

If it makes sense to just bundle that into this PR I could toss that in.

adriancole:
Span customizer is a term for how to gain access to the current span. what
is in the request is indeed usually that, but this feature is unlike
anything else so far, as it is encouraging a framework specific accessor.

In general, my thinking is to be like everything else until you cant. less
to explain and we already get tons of questions. let's try the conventional
for now as it gets the feature done and away from code I think?

later if we do framework specific accessors we can revisit this approach
here just yeah..

lemme know your thoughts.

fitzoh:
So I guess it goes back to the way that you get a span to customize it.

I would assume that `CurrentSpanCustomizer` is the most commonly used `SpanCustomizer`, which depends on `CurrentTraceContext`, which is based on `ThreadLocal`'s, which aren't usable for WebFlux applications, which might not have all operations happen on the same thread.

If those assumptions are correct, there's currently no defined way to *get* the current `Span` in a WebFlux application to customize it, which is what this PR provides.

adriancole:
if flux is broken in scoping other things will break though right? like
child span
 are we walking around a different problem?

fitzoh:
I'm not sure, and it's entirely possible I'm making incorrect assumptions ¯\\\_(ツ)\_/¯

Any thoughts @marcingrzejszczak? (saw you handled #344, which is related-ish)

PR #1126, opened 2018-11-06
Stephane help

marcingrzejszczak:


smaldini:
Moar Reactor optimization + fixing thread local populated on WebClient body emissions (see Spring Cloud Sample ITIssue)

anatoliy-balakirev:
After this PR it looks like tracing headers are added to the context only when `subscribe` is called, while previously they were added in the `subscriberContext` (I'm not so familiar with all this reactive stuff, but in debug I see that it was happening earlier with the old code). So my `ExchangeFilterFunction`, where I relied on presence of tracing headers stopped working now (again, I'm not sure if that's a bug or I just shouldn't rely on that there). So could someone clarify if that's expected or rather some regression? If it's expected - is there any option to somehow force this code to be executed before my filter is applied?
screen shot 2019-01-28 at 11 35 27

PR #1139, opened 2018-11-20
First attempt to add GRPC instrumentation

tkvangorder:


spring-cloud-sleuth-core/pom.xml
```diff
@@ -32,6 +32,11 @@
 	</parent>
 
 	<dependencies>
+		<dependency>
```
adriancole: I don't think this can be a mandatory dependency.

spring-cloud-sleuth-samples/spring-cloud-sleuth-sample-grpc/README.md
```diff
@@ -0,0 +1,33 @@
...
+
+1. Edit src/main/resources/application.yml and uncomment the "sample.zipkin.enable=false"
+2. Import this maven project into your favorite IDE (note, I have not setup m2e to automatically build the stubs from within Eclipse)
+3. Run the spring boot application.
+4. In a browser "http://localhost:8080/hi/fred"
+
+
+## Running with Zipkin:
+
+1. For conveinence, there is a docker-compose file in the project that can be used to launch zipkin locally. Otherwise, you will need to run zipkin separately.
```
adriancole: Typo

```suggestion
1. For convenience, there is a docker-compose file in the project that can be used to launch zipkin locally. Otherwise, you will need to run zipkin separately.
```

spring-cloud-sleuth-samples/spring-cloud-sleuth-sample-grpc/docker-compose.yml
```diff
@@ -0,0 +1,62 @@
...
+#
+# This runs the zipkin and zipkin-mysql containers, using docker-compose's
+# default networking to wire the containers together.
+#
+# Note that this file is meant for learning Zipkin, not production deployments.
+
+version: '2'
+
+services:
+  storage:
```
adriancole: mysql is unnecessary for this.. just use default in-mem.. probably docker-compose itself is overkill

See also https://github.com/openzipkin/sleuth-webmvc-example

tkvangorder:
@adriancole Thanks for the quick turn-around, I will move the sample out of the core project.

spring-cloud-sleuth-core/pom.xml
tkvangorder: yeah, after thinking about this, I can make this a test scope dependency, as it is during the tests that the failures manifest. Still not a great solution, but I am not sure what else would make more sense?

spring-cloud-sleuth-samples/spring-cloud-sleuth-sample-grpc/docker-compose.yml
tkvangorder: I will clean that up as well, thanks.

adriancole:
>
> yeah, after thinking about this, I can make this a test scope dependency,
> as it is during the tests that the failures manifest. Still not a great
> solution, but I am not sure what else would make more sense?
>
I agree it is a problem with grpc as it strictly depends on guava. You can
try dependencyManagement section to fix the version, but in practice sleuth
or boot bom will set the dependency versions. optional or provided scope I
think can also help make the build work but not interfere with versions
used by apps.

tkvangorder:
- Sample Project has been pulled out of this repo and is now here: [https://github.com/tkvangorder/sleuth-grpc-example](https://github.com/tkvangorder/sleuth-grpc-example). @adriancole Once/if this PR is merged, I can work with you on migrating it to openzipkin.
- I moved the "forcing of the guava version" to test scope.
- I cleaned up the two failing tests.
- I still need to add some testing.

tkvangorder:
Pretty close, just fighting with my integration test. It runs fine if I run just those three tests, but fails when running the whole suite. I am thinking that something is dirtying the trace context or I am not understanding how the reporter works. The spans of the interceptors are not being stored in the reporter when I run the suite..I will dig in some more tomorrow.

tkvangorder:
@adriancole @marcingrzejszczak Looks like all tests are now passing, but something has changed in the pipeline : 

```
#!/bin/bash -eo pipefail
mkdir -p $CIRCLE_TEST_REPORTS/junit/
find . -type f -regex ".*/target/.*-reports/.*" -exec cp {} $CIRCLE_TEST_REPORTS/junit/ \;
bash <(curl -s https://codecov.io/bash)
/dev/fd/63: line 1: html: No such file or directory
/dev/fd/63: line 2: syntax error near unexpected token `<'
'
Exited with code 2
```

tkvangorder:
@adriancole @marcingrzejszczak I did a little refactoring and added some documentation around the implementation details. How do you both feel about the dependency on grpc-spring-boot-starter? This can work for now but it feels like it would be better if there were a spring-sanctioned gRPC project: spring-grpc or spring-cloud-grpc?

devinsba:
A gRPC sanctioned Spring-Boot starter would also be awesome: https://github.com/grpc-ecosystem could be a good home. I know there are other starters hanging around so it would be good to have a community maintained one that gRPC is willing to put their name on

spencergibb:
I might think https://github.com/spring-cloud/spring-cloud-gcp

tkvangorder:
I think after Google open sourced gRPC they handed things over to the community, so I am not sure if it makes sense to put it under the GCP umbrella? I am a relative newbie to gRPC, so take my opinion with a grain of salt.

Regardless of home, having an agreed upon home for Spring + gRPC would be great. In addition to [https://github.com/LogNet/grpc-spring-boot-starter](https://github.com/LogNet/grpc-spring-boot-starter), there is also an interesting reactive effort [https://github.com/salesforce/reactive-grpc](https://github.com/salesforce/reactive-grpc)

spencergibb:
From what I recall, the boot team prefers the starter to be external until there is some Framework integration (which there isn't, but there is a jira issue).

marcingrzejszczak:
I'm merging what we have here ATM and we can think of changing the starter to sth else if necessary.

yishibai02:
Another repo has done this. 
https://github.com/yidongnan/grpc-spring-boot-starter
This repo seems to provide a lot of features.

tkvangorder:
@yishibai02 This PR relies on grpc-spring-boot-starter being on the classpath and will add tracing interceptors to the servers (by introducing a brave's server tracing interceptor) and to clients (by introducing a spring-aware ManagedChannelBuilder and brave's client tracing interceptor)

yishibai02:
I see this repo also supports Spring Sleuth to trace application.
[GrpcServerAutoConfiguration.java#L147-L159](https://github.com/yidongnan/grpc-spring-boot-starter/blob/13b41762c668c2851335944e9ffef4322398cb77/grpc-server-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/server/autoconfigure/GrpcServerAutoConfiguration.java#L147-L159)
[GrpcClientAutoConfiguration.java#L158-L170](https://github.com/yidongnan/grpc-spring-boot-starter/blob/13b41762c668c2851335944e9ffef4322398cb77/grpc-client-spring-boot-autoconfigure/src/main/java/net/devh/boot/grpc/client/autoconfigure/GrpcClientAutoConfiguration.java#L158-L170)

I think these commit be better placed in the third-party starter repo.

tkvangorder:
@yishibai02. Thanks for the clarification!

The repo's have the same name but are very different in the implementation details.

For now, this PR introduces auto-configuration only when the [LogNet/grpc-spring-boot-starter](https://github.com/LogNet/grpc-spring-boot-starter) is on the classpath.

There is still work to be done to introduce an "official" project for using grpc with spring/spring boot. Perhaps [yidongnan/grpc-spring-boot-starter](https://github.com/yidongnan/grpc-spring-boot-starter) might be a good place to start as a contribution towards that goal?

PR #1140, opened 2018-11-21
Fixes #1138 Report trace in case of cancellation when calling webclient.

### LINKED ISSUE #1138: TraceWebClient should report trace in case of cancellation
When using Sleuth with WebFlux we get traces for webclient calls on success and error but we don't get any trace on cancellation. The most common case of cancellation on a stream originating from a webClient would be timeout.

```
webClient.post()
                 .uri("http://example.com")
                 .retrieve()
                .bodyToMono(String.class)
                .timeout(Duration.ofSeconds(1))
```

It would be perfect if `TraceWebClientBeanPostProcessor` could be changed to handle cancellation and report them as a special type of error maybe?

I'm happy to raise a PR if you think it's a good idea.

madgnome:


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebFilter.java
```diff
@@ -199,7 +200,18 @@ public void subscribe(CoreSubscriber<? super Void> subscriber) {
...
-				this.actual.onSubscribe(subscription);
+				this.actual.onSubscribe(new Subscription() {
+					@Override
+					public void request(long n) {
+						subscription.request(n);
+					}
+
+					@Override
+					public void cancel() {
+						terminateSpan(CancelledSubscriptionException.INSTANCE);
```
adriancole: I generally don't like publicly exposing new exception types especially as it encourages people to lock to instrumentation code types. Secondarily, the only effect desired is to add an error tag and allocating exceptions isn't cheap.

If we think cancel is an error, which is debatable, please delete the exception and just add a tag("error", "canceled");

if there is an existing practice in webflux calling canceled things as exceptions, we could revisit this as well.

adriancole: I notice this is a singleton exception, which is even more odd even if allocations are fixed. :) folks will expect the stack trace to be correct and it definitely won't be. let's just use tag and skip the ceremony.

madgnome: Cool, will switch to using tag directly. If I use tag directly I can use another tag name than `error` any preference? `cancel=true`?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java
```diff
@@ -289,6 +299,16 @@ void handleReceive(Span clientSpan, Tracer.SpanInScope ws,
...
 			}
 
+			void terminateSpanOnCancel() {
+				if (log.isDebugEnabled()) {
+					log.debug("Subscription was cancelled. Will close the span ["
+							+ span + "]");
+				}
+
+				span.customizer().tag("error", CANCELLED_SUBSCRIPTION_ERROR);
+				handleReceive(span, ws, null, null);
```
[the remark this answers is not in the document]
adriancole: if you already have a span ref, probably just span.tag is best.

madgnome:
The build seems flaky, how do I rerun it?

PR #1150, opened 2018-12-04
Skip only Actuator endpoints when context path set

### LINKED ISSUE #1146: Regression: server.servlet.context-path skips all endoints
**Sprint Boot:** 2.1.0
**Spring Cloud:** Greenwich.M3

Regression from this issue: https://github.com/spring-cloud/spring-cloud-sleuth/issues/971

If you include a servlet context path like this:
```yaml
server:
  port: 8080
  servlet:
      context-path: /melloware/
```

All endpoints are skipped and no Zipkin messages are sent.  If you remove the context-path then all Zipkin messages are sent and received.
```yaml
server:
  port: 8080
```

shakuzen:
Previously, all endpoints under the set context path would be skipped. Instead, this skips only Actuator endpoints under the set context-path.

Resolves #1146

---

We might want to reconsider the current approach, though. Previously, the management context path was appended to the default pattern instead of added as an OR to the pattern. The default pattern includes more things than just the Actuator endpoints, which won't be skipped if a context path is set currently. Also, if the Actuator base path is set to `/` then the approach in this change will have issues because it will again skip all endpoints for the application. This should be an improvement in most cases, though.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebAutoConfiguration.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebAutoConfiguration.java
@@ -23,6 +23,7 @@ import java.util.stream.Collectors;
 
 import brave.Tracing;
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.actuate.autoconfigure.endpoint.web.WebEndpointProperties;
 import org.springframework.boot.actuate.autoconfigure.web.server.ManagementServerProperties;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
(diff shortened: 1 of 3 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebAutoConfiguration.java
```diff
@@ -96,23 +97,27 @@ public SingleSkipPattern skipPatternForManagementServerProperties(
...
-				return Optional.of(Pattern.compile(contextPath + ".*"));
+				return Optional.of(Pattern.compile(
+						contextPath + webEndpointProperties.getBasePath() + ".*"));
 			}
 			return Optional.empty();
 		}
 
 		@Bean
-		@ConditionalOnBean(ServerProperties.class)
+		@ConditionalOnBean({ ServerProperties.class, WebEndpointProperties.class })
```
marcingrzejszczak: You know that it's an OR not an AND?

PR #1156, opened 2018-12-07
Allow customizing the base Propagation.Factory

devinsba:
In order to use the Sleuth propagation customization with a propagation mechanism other than B3 it requires copying the entire bean definition. To make user customization not able to fall behind the official implementation of this bean it would be good to be able to customize the base implementation.

This PR adds an interface to facilitate that, ideally we could allow the user to just create an instance of `Propagation.Factory` and maybe set a property to enable a wrapping behavior. I'll play with that idea, but for now this gets the idea out for feedback.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
@@ -24,12 +24,13 @@ import brave.ErrorParser;
 import brave.Tracer;
 import brave.Tracing;
 import brave.handler.FinishedSpanHandler;
-import brave.propagation.B3Propagation;
 import brave.propagation.CurrentTraceContext;
 import brave.propagation.ExtraFieldPropagation;
 import brave.propagation.Propagation;
 import brave.propagation.ThreadLocalCurrentTraceContext;
 import brave.sampler.Sampler;
+import org.springframework.cloud.sleuth.propagation.DefaultBasePropagationFactoryProvider;
+import org.springframework.cloud.sleuth.propagation.BasePropagationFactoryProvider;
 import zipkin2.Span;
 import zipkin2.reporter.Reporter;
 
@@ -121,13 +122,21 @@ public class TraceAutoConfiguration {
 
 	@Bean
 	@ConditionalOnMissingBean
-	Propagation.Factory sleuthPropagation(SleuthProperties sleuthProperties) {
+	BasePropagationFactoryProvider basePropagationFactoryProvider() {
+		return new DefaultBasePropagationFactoryProvider();
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	Propagation.Factory sleuthPropagation(
+			SleuthProperties sleuthProperties,
+			BasePropagationFactoryProvider basePropagationFactoryProvider) {
 		if (sleuthProperties.getBaggageKeys().isEmpty()
 				&& sleuthProperties.getPropagationKeys().isEmpty()) {
-			return B3Propagation.FACTORY;
+			return basePropagationFactoryProvider.get();
 		}
 		ExtraFieldPropagation.FactoryBuilder factoryBuilder = ExtraFieldPropagation
-				.newFactoryBuilder(B3Propagation.FACTORY);
+				.newFactoryBuilder(basePropagationFactoryProvider.get());
 		if (!sleuthProperties.getBaggageKeys().isEmpty()) {
 			factoryBuilder = factoryBuilder
 					// for HTTP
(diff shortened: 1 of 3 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/propagation/DefaultBasePropagationFactoryProvider.java
```diff
@@ -0,0 +1,33 @@
...
+import brave.propagation.B3Propagation;
+import brave.propagation.Propagation;
+
+/**
+ * Provides the default {@link brave.propagation.Propagation.Factory}
+ *
+ * @author Brian Devins-Suresh
+ * @since 2.1.0
+ */
+public class DefaultBasePropagationFactoryProvider implements BasePropagationFactoryProvider {
```
marcingrzejszczak: Maybe `B3PropagationFactoryProvider` ? Also you can make it package scope

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/propagation/BasePropagationFactoryProvider.java
```diff
@@ -0,0 +1,40 @@
...
+import javax.validation.constraints.NotNull;
+
+/**
+ * Provides a {@link brave.propagation.Propagation.Factory} that Sleuth customizes with baggage and propagated key
+ * configuration.
+ *
+ * @author Brian Devins-Suresh
+ * @since 2.1.0
+ */
+public interface BasePropagationFactoryProvider {
```
marcingrzejszczak: Why not `PropagationFavtoryProvider` ?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
```diff
@@ -121,13 +122,21 @@ SpanNamer sleuthSpanNamer() {
...
 	@Bean
 	@ConditionalOnMissingBean
-	Propagation.Factory sleuthPropagation(SleuthProperties sleuthProperties) {
+	BasePropagationFactoryProvider basePropagationFactoryProvider() {
+		return new DefaultBasePropagationFactoryProvider();
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	Propagation.Factory sleuthPropagation(
```
marcingrzejszczak: Another thing that we could actually do is to 

- create a bean of `ExtraFieldPropagation.FactoryBuilder` 
- we would pass it to `sleuthPropagation`
- that way we don't need a separate interface

WDYT?

devinsba: That sounds much nicer, I'll do a second pass with that. Want me to squash into one commit in the PR?

marcingrzejszczak: Nah, that's ok.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/propagation/DefaultBasePropagationFactoryProvider.java
devinsba: Removed

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
```diff
@@ -43,6 +37,11 @@
 import org.springframework.cloud.sleuth.SpanNamer;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import zipkin2.Span;
+import zipkin2.reporter.Reporter;
+
+import java.util.ArrayList;
+import java.util.List;
```
devinsba: I'll revert this, IntelliJ organize imports apparently didn't like the old order

devinsba:
Test failure looks like a possible test flake in reactor

1991wangliang:
about dubbo support ExtraFieldPropagation question.

i was created ExtensionFactory by dubbo or create spring bean of Tracing . but i will get a null param. that's my coding.https://github.com/1991wangliang/sleuth-webmvc-example,
i try when i remove  build Tracing bean coding,that will be  ok, 

```
  //when remove this code,that will get groupId value.
  @Bean
  public Tracing tracing(){
    return Tracing.newBuilder()
            .propagationFactory(ExtraFieldPropagation.newFactory(B3Propagation.FACTORY,"groupId"))
            .build();
  }

```

that was a bug?

1991wangliang:
when create Tracing, i debug find when i set extra that `TraceContext.currentTraceContext()` was null .
when execute get method.
```
  @Nullable static TraceContext currentTraceContext() {
    Tracing tracing = Tracing.current();
    return tracing != null ? tracing.currentTraceContext().get() : null;
  }
```

Issue #1071, opened 2018-08-23
Add configuration to put baggage automatically to MDC


marcingrzejszczak:
with this change you can set a property to pass the values from baggage automatically to MDC

tdanylchuk:
along with this, baggage could be propagated to span tags, so they can be collected by zipkin to enable trace filtering by baggage.

adriancole:
>
> along with this, baggage could be propagated to span tags, so they can be
> collected by zipkin to enable trace filtering by baggage.
>
please be careful with this as it will dramatically reduce cost
effectiveness of storage. Usually you do not need to copy a field to every
span (except the trace ID). IOTW, if such a feature happens, please make it
very very very very optional or parsimoniously handled.

Right now, anyone can do this themselves by setting up a custom parser.

>

PR #1158, opened 2018-12-12
sleuth extra propagation fields to tags feature.

tdanylchuk:
As part of this conversation - https://github.com/spring-cloud/spring-cloud-sleuth/issues/1071#issuecomment-415447228

by default this feature is disabled, but if someone what to propagate baggage to spans in order to enable filtering by it - they could use this feature, by just adding needed property.

In this way we are omitting changing code as described here - https://cloud.spring.io/spring-cloud-sleuth/single/spring-cloud-sleuth.html#_baggage_versus_span_tags

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/propagation/SleuthTagPropagationAutoConfiguration.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/propagation/SleuthTagPropagationAutoConfiguration.java
@@ -14,12 +14,10 @@
  * limitations under the License.
  */
 
-package org.springframework.cloud.sleuth.log;
+package org.springframework.cloud.sleuth.propagation;
 
-import brave.propagation.CurrentTraceContext;
-import org.slf4j.MDC;
+import brave.handler.FinishedSpanHandler;
 import org.springframework.boot.autoconfigure.AutoConfigureBefore;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.sleuth.autoconfig.SleuthProperties;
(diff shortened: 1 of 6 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/propagation/TagPropagationFinishedSpanHandler.java
```diff
@@ -0,0 +1,69 @@
...
+			SleuthTagPropagationProperties tagPropagationProperties) {
+		this.sleuthProperties = sleuthProperties;
+		this.tagPropagationProperties = tagPropagationProperties;
+	}
+
+	@Override
+	public boolean handle(TraceContext context, MutableSpan span) {
+		if (context == null) {
+			return true;
+		}
```
devinsba: If you update to Brave `5.6.0` then you can add another exit condition of: `!context.isLocalRoot()` that way this doesn't tag every span and blow up your storage requirements.

PS: I have a feeling someone else will block this being merged until that is done

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/propagation/SleuthTagPropagationAutoConfiguration.java
```diff
@@ -0,0 +1,52 @@
...
+ * @author Taras Danylchuk
+ * @since 2.1.0
+ */
+@Configuration
+@ConditionalOnProperty(value = "spring.sleuth.enabled", matchIfMissing = true)
+@AutoConfigureBefore(TraceAutoConfiguration.class)
+public class SleuthTagPropagationAutoConfiguration {
+
+	@Configuration
+	@ConditionalOnProperty(value = "spring.sleuth.propagation.tag.enabled", matchIfMissing = true)
```
devinsba: I don't think you need this conditional since you are already making the bean within this class conditional

tdanylchuk: At this point ```FinishedSpanHandler``` is not being initialized, or I misunderstood your statement? :D

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/propagation/TagPropagationFinishedSpanHandler.java
tdanylchuk: I don't really aware about this condition, so if members will require it, I will add - no problem

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/propagation/TagPropagationFinishedSpanHandler.java
```diff
@@ -0,0 +1,69 @@
...
+
+	public TagPropagationFinishedSpanHandler(SleuthProperties sleuthProperties,
+			SleuthTagPropagationProperties tagPropagationProperties) {
+		this.sleuthProperties = sleuthProperties;
+		this.tagPropagationProperties = tagPropagationProperties;
+	}
+
+	@Override
+	public boolean handle(TraceContext context, MutableSpan span) {
+		if (context == null) {
```
adriancole: this is incorrect as the context will never be null

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/propagation/TagPropagationFinishedSpanHandler.java
```diff
@@ -0,0 +1,69 @@
...
+			return true;
+		}
+		Stream.concat(whitelistedKeys(sleuthProperties.getBaggageKeys()),
```
adriancole: I will suggest a more efficient way to accomplish this.. gimme a little bit :P

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/propagation/SleuthTagPropagationAutoConfiguration.java
marcingrzejszczak: This feature was supposed to be disabled by default, correct? It seems that it's turned on by default.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/propagation/SleuthTagPropagationAutoConfigurationTest.java
```diff
@@ -0,0 +1,60 @@
...
+
+import org.junit.Test;
+import org.springframework.boot.autoconfigure.AutoConfigurations;
+import org.springframework.boot.test.context.runner.ApplicationContextRunner;
+import org.springframework.cloud.sleuth.autoconfig.TraceAutoConfiguration;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+
+public class SleuthTagPropagationAutoConfigurationTest {
```
marcingrzejszczak: Can you make the suffix `Tests` instead of `Test`, please?

tdanylchuk:
Sorry, was on vacation ;(

marcingrzejszczak:
No problem! Congrats on your contribution!

PR #1166, opened 2018-12-18
Fix conditional on exception-throwing-filter-enabled

### LINKED ISSUE #1165: exception-throwing-filter-enabled property is ignored
The `spring.sleuth.web.exception-throwing-filter-enabled` property is ignored. This property is described in the following part of the documentation:

https://cloud.spring.io/spring-cloud-sleuth/single/spring-cloud-sleuth.html#_http_filter

mziemba:
Fixes #1165 

Related: https://github.com/spring-projects/spring-boot/issues/13449#issuecomment-396520674


marcingrzejszczak:
Maybe you could add a test that ensures that the bean is not registered with that property?

marcingrzejszczak:
I applied your suggestion related to the name (great catch). Congrats on your contribution :) Gratulacje!

Issue #1180, opened 2019-01-10
Support multiple tracing systems


TYsewyn:
Currently one of the requirements I have is to support multiple tracing systems (GCP Stackdriver Trace & Zipkin). Unfortunately it is not possible to use the auto-configuration which comes with the 2 dependencies (`spring-cloud-gcp-starter-trace ` and `spring-cloud-starter-sleuth-zipkin`).
Supporting both systems also requires to duplicate almost all of the auto-configuration.

It'd be interesting if `spring-cloud-sleuth-core` comes shipped with an `@Primary` bean which is a composite of all the configured `SpanReporter`s, or something that is a brave based `FinishedSpanHandler ` (which will be more efficient) like @adriancole said on gitter.

adriancole:
yep and to clarify.. right now, no reporter is based on `FinishedSpanHandler` but anyway the overall design you mention seems to apply regardless

marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-sleuth/pull/1181

PR #1181, opened 2019-01-13
Added support for multiple reporters

TYsewyn:
Targets issue #1180.


spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/ZipkinAutoConfiguration.java
```diff
@@ -72,10 +73,11 @@
 	 * Accepts a sender so you can plug-in any standard one. Returns a Reporter so you can
 	 * also replace with a standard one.
 	 */
-	@Bean
+	@Bean("zipkinReporter")
```
marcingrzejszczak: maybe extract to some constant?

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/sender/ZipkinKafkaSenderConfiguration.java
```diff
@@ -42,7 +42,7 @@
 	@Value("${spring.zipkin.kafka.topic:zipkin}")
 	private String topic;
 
-	@Bean
+	@Bean("zipkinSender")
```
marcingrzejszczak: maybe extract to some constant?

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/sender/ZipkinRabbitSenderConfiguration.java
```diff
@@ -36,7 +36,7 @@
 	@Value("${spring.zipkin.rabbitmq.queue:zipkin}")
 	private String queue;
 
-	@Bean
+	@Bean("zipkinSender")
```
marcingrzejszczak: maybe extract to some constant?

spring-cloud-sleuth-zipkin/src/test/java/org/springframework/cloud/sleuth/zipkin2/ZipkinAutoConfigurationTests.java
```diff
@@ -184,6 +189,25 @@ public void rabbitWinsWhenKafkaPresent() throws Exception {
 		this.context.close();
 	}
 
+	@Test
+	public void supportsMultipleReporters() {
```
marcingrzejszczak: Can you report a span and verify that both senders and reporters were actually called?

devinsba:
I'm worried that this will result in a hidden mis-configuration. I think this feature should require a flag

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/ZipkinAutoConfiguration.java
```diff
@@ -68,14 +70,27 @@
...
 	/**
 	 * Accepts a sender so you can plug-in any standard one. Returns a Reporter so you can
 	 * also replace with a standard one.
 	 */
-	@Bean
+	@Bean(REPORTER_BEAN_NAME)
 	@ConditionalOnMissingBean
 	public Reporter<Span> reporter(ReporterMetrics reporterMetrics,
-			ZipkinProperties zipkin, Sender sender, BytesEncoder<Span> spanBytesEncoder) {
+			ZipkinProperties zipkin, @Qualifier(SENDER_BEAN_NAME) Sender sender,
```
devinsba: I think this needs a doc update associated with it since now overriding requires the correct bean name for a `Sender`

marcingrzejszczak:
@devinsba 

> IE: Maybe I bring in 2 different starters that have Reporters in them and neither of them with @ConditionalOnMissingBean

Maybe I don't get it yet, but if neither is `@ConditionalOnMissingBean` we will get two reporters that are being used. I guess that's the point?

>  I'm worried that this will result in a hidden mis-configuration. I think this feature should require a flag

I'm for making this opt in - that makes sense.

TYsewyn:
> I'm for making this opt in - that makes sense.

Honestly I do not agree, this shouldn't be configured with a property.
If I put 2 separate tracing systems intentionally on my classpath, both should work out of the box without adding more custom configuration.

TYsewyn:
> Fair enough, makes sense. So let's make it opt out :)

opt-out can be done using eg. `spring.zipkin.enabled=false` or `spring.cloud.gcp.trace.enabled=false` :)

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/ZipkinAutoConfiguration.java
TYsewyn: This whole change needs a doc update because the overall configuration has been updated to support multiple tracers.
I already made the change for spring-cloud-gcp-trace (https://github.com/spring-cloud/spring-cloud-gcp/pull/1383).

TYsewyn:
Pinging @adriancole to have a second pair of eyes.
@marcingrzejszczak and I went through it and it looks good to us. :)

TYsewyn:
That's weird... It's the second time that `SleuthSpanCreatorAspectMonoTests` is behaving flakey on CircleCI while everything is passing locally.
@marcingrzejszczak @adriancole can any of you verify this?

marcingrzejszczak:
Fixed via https://github.com/spring-cloud/spring-cloud-sleuth/commit/490a1d8626f286935d8eb8cfd4df8bfbca7bfb5e and other commits by @TYsewyn

PR #1183, opened 2019-01-15
The serviceName is not being resolved properly

jwcarman:


pivotal-issuemaster:
@jwcarman Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=1183)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=1183) to manually synchronize the status of this Pull Request.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
```diff
@@ -83,7 +88,9 @@ Tracing tracing(
 			Reporter<zipkin2.Span> reporter, Sampler sampler, ErrorParser errorParser,
 			SleuthProperties sleuthProperties) {
 		Tracing.Builder builder = Tracing.newBuilder().sampler(sampler)
-				.errorParser(errorParser).localServiceName(serviceName)
+				.errorParser(errorParser)
```
marcingrzejszczak: the default `default` service name is coming from Spring. I don't think that we should hardcode it as you did that

pivotal-issuemaster:
@jwcarman Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=1183)!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
jwcarman: It's hardcoded in the @Value expression in this very file already

```
@Value("${spring.zipkin.service.name:${spring.application.name:default}}") String serviceName
```

jwcarman: The default value that comes from Spring Boot is "".  So, this will likely never get the value of "default" with that @Value expression.

marcingrzejszczak: Oh, indeed it is. I wonder why it's resolved to an empty String. I think we should investigate that first

jwcarman: Spring boot sets it to empty string.  It's never null.

marcingrzejszczak: Yeah, I see that. So how about instead of:

`"".equals(serviceName) ? DEFAULT_SERVICE_NAME : serviceName`

we would have

`StringUtils.isEmpty(serviceName) ? DEFAULT_SERVICE_NAME : serviceName`

WDYT?

jwcarman: That'll work.  I didn't realize that we had that at our disposal, but that's certainly more readable.  You want me to fix this PR?

marcingrzejszczak: Yeah, if you could. `StringUtils` from `org.springframework.utils` AFAIR

jwcarman: How about this (I'm an Apache Commons committer, so I couldn't resist ;)?
```
StringUtils.defaultIfEmpty(serviceName, DEFAULT_SERVICE_NAME)
```

marcingrzejszczak: Let's use the Spring one please.

marcingrzejszczak:
Congrats @jwcarman on your first contribution :)

PR #1196, opened 2019-01-27
Remove cycles from auto configuration

bartekbp:
I work on the app that had in the past problems with deployment reproducibility. The app couldn't boot due to unresolved cycles between beans. After inspecting the problem, it turned out to be related to jar packaging. The order of classes within jar had an impact on Spring beans loading, so if you took jar known to have cycles and run it multiple times, you got always unresolved cycles. If you took it and repackaged classes and run, it was quite common for an app to boot properly. 

My solution to that problem was to disable resolving circular references between beans altogether and fix each cycle by hand by remodelling relationships between beans or delaying bean creation. This pr removes cycles in configuration classes so that Sleuth can be used by apps with `setAllowCircularReferences` set to false.


marcingrzejszczak:
@bartekbp siemano! Congrats on your first contribution :)

Issue #1199, opened 2019-01-28
Regression: sleuth.propagation-keys aren't available in filter with WebFlux


anatoliy-balakirev:
Bug report:

Propagation keys are not available in the filter with WebFlux.

Spring Boot version: `2.1.2.RELEASE`
Spring Cloud version: `Greenwich.RELEASE`. No issue with `Finchley.RELEASE`
Sample project: https://github.com/anatoliy-balakirev/sleuth-test

So basically if I set propagation key:
```
spring.sleuth.propagation-keys=x-b3-user-id
```
Within my first controller:
```
@Slf4j
@RestController
@RequiredArgsConstructor
public class DemoController {

    private final WebClient webClient;

    @GetMapping("/demo")
    public Mono<String> getMessageFromContext() {
        ExtraFieldPropagation.set(USER_ID_KEY, "123");
        log.info("Sending request to another controller. Set userId as an extra field: {}", USER_ID_KEY);

        return webClient.get().uri("/userId").retrieve().bodyToMono(String.class).doOnNext(userId -> {
            log.info("User id from another controller is: {}", userId);
        });
    }

}
```
And then try to access it within filter, which I register in the WebClient:
```
    @Bean
    public WebClient myWebClient(final WebClient.Builder webClientBuilder) {
        return webClientBuilder.baseUrl("http://localhost:8080")
            .filter((request, exchangeFunction) -> {
                final String userId = request.headers().getFirst(USER_ID_KEY);
                log.info("User id in filter is: {}", userId);
                return exchangeFunction.exchange(request);
            })
            .build();
    }
```
I'm getting `null` there. Still that key is propagated to my next controller, I can access it there and send back:
```
@Slf4j
@RestController
@RequiredArgsConstructor
public class UserIdController {


    @GetMapping("/userId")
    public Mono<String> getMessage() {
        return Mono.fromCallable(() -> {
            log.info("Got into userId controller");
            return ExtraFieldPropagation.get(USER_ID_KEY);
        });
    }

}
```
Here is the log with `Greenwich.RELEASE`:
```
2019-01-28 13:37:02.098  INFO [-,a2058a4c791cf166,a2058a4c791cf166,false] 87989 --- [ctor-http-nio-3] com.sleuth.test.demo.DemoController      : Sending request to another controller. Set userId as an extra field: x-b3-user-id
[...]

PR #1206, opened 2019-02-05
Fix regression introduced in #1126 (brave headers propagation)

bsideup:
Before #1126, the headers were eagerly set in
`TraceExchangeFilterFunction#filter`. After it, the side effect was
moved to lazy `MonoWebClientTrace#subscribe`.

However, we have everything to instrument the request in `filter`,
and it can be done eagerly.


bsideup:
@spencergibb 

> Test?

That's exactly what we're discussing with @marcingrzejszczak right now in Slack :D
I do agree that this change should not be accepted without a test, but I need to sync whether I have additional time to do it (since the issue was purely in Spring Cloud Sleuth).

Will check shortly, I like writing tests 👍

bsideup:
FYI the manual testing was done with the repro project from https://github.com/spring-cloud/spring-cloud-sleuth/issues/1199

bsideup:
@spencergibb ok, writing a test took less than 15 minutes 🎉

Issue #1217, opened 2019-02-21
Add support for SQS


marcingrzejszczak:

devinsba:
@marcingrzejszczak can you elaborate a bit on this? Have you looked at: https://github.com/SmartThingsOSS/smartthings-brave/tree/master/brave-sqs if you are considering the propagation. Or are you thinking more along the lines of native integration with `spring-cloud-aws-messaging`. I've been working through this exact thing the last couple weeks so I'm interested in the approach and how I can make our apps less custom

marcingrzejszczak:
> Have you looked at: https://github.com/SmartThingsOSS/smartthings-brave/tree/master/brave-sqs if you are considering the propagation

I haven't. Thanks for pointing it out!

> Or are you thinking more along the lines of native integration with `spring-cloud-aws-messaging`.

I was also thinking about it. I wanted to make sure that it's actually working. `spring-cloud-aws-messaging` is using Spring messaging under the hood, so I wanted to double check if it maybe works out of the box right now (by any chance).

>  I've been working through this exact thing the last couple weeks so I'm interested in the approach and how I can make our apps less custom.

I'd love to join forces then. Whatever works, we can autoconfigure it in Sleuth.

devinsba:
@marcingrzejszczak I spiked an attempt in #1218

marcingrzejszczak:
Fixed via https://github.com/spring-cloud/spring-cloud-sleuth/pull/1218

PR #1218, opened 2019-02-22
SQS Messaging Autoconfig

devinsba:
partially addressed #1217 

Attempts to make spring-messaging generic component for use elsewhere `TracingMethodMessageHandlerAdapter`. I'm still thinking about how to abstract/wrap the existing classes so code doesn't have to me copied from the Impl on the other side

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-sleuth.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-sleuth.adoc
@@ -1381,6 +1381,13 @@ To block this feature, set `spring.sleuth.messaging.jms.enabled` to `false`.
 
 IMPORTANT: We don't support baggage propagation for JMS
 
+==== Spring Cloud AWS Messaging SQS
+
+We instrument `@SqsListener` which is provided by `org.springframework.cloud:spring-cloud-aws-messaging`
+so that tracing headers get extracted from the message and a trace gets put into the context.
+
+To block this feature, set `spring.sleuth.messaging.sqs.enabled` to `false`.
+
 === Zuul
 
 We instrument the Zuul Ribbon integration by enriching the Ribbon requests with tracing information.
--- a/spring-cloud-sleuth-core/pom.xml
+++ b/spring-cloud-sleuth-core/pom.xml
@@ -84,6 +84,11 @@
 		</dependency>
 		<dependency>
 			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-aws-messaging</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-starter-gateway</artifactId>
 			<optional>true</optional>
 		</dependency>
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TraceMessagingAutoConfiguration.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TraceMessagingAutoConfiguration.java
@@ -17,12 +17,15 @@
 package org.springframework.cloud.sleuth.instrument.messaging;
 
 import java.lang.reflect.Field;
+import java.util.Collections;
+import java.util.List;
 
 import brave.Span;
 import brave.Tracer;
 import brave.Tracing;
 import brave.jms.JmsTracing;
 import brave.kafka.clients.KafkaTracing;
+import brave.propagation.Propagation;
 import brave.spring.rabbit.SpringRabbitTracing;
 import org.aopalliance.intercept.MethodInterceptor;
 import org.aopalliance.intercept.MethodInvocation;
@@ -49,6 +52,8 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.cloud.aws.messaging.config.QueueMessageHandlerFactory;
+import org.springframework.cloud.aws.messaging.listener.QueueMessageHandler;
 import org.springframework.cloud.sleuth.autoconfig.TraceAutoConfiguration;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
@@ -62,6 +67,11 @@ import org.springframework.kafka.listener.MessageListener;
 import org.springframework.kafka.listener.MessageListenerContainer;
 import org.springframework.kafka.listener.adapter.MessagingMessageListenerAdapter;
 import org.springframework.kafka.support.DefaultKafkaHeaderMapper;
+import org.springframework.messaging.Message;
+import org.springframework.messaging.MessagingException;
+import org.springframework.messaging.converter.MessageConverter;
+import org.springframework.messaging.support.MessageHeaderAccessor;
+import org.springframework.util.CollectionUtils;
 import org.springframework.util.ReflectionUtils;
 
 /**
@@ -73,7 +83,7 @@ import org.springframework.util.ReflectionUtils;
  */
 @Configuration
 @ConditionalOnBean(Tracing.class)
-@AutoConfigureAfter({ TraceAutoConfiguration.class })
+@AutoConfigureAfter({ TraceAutoConfiguration.class, TraceSpringMessagingAutoConfiguration.class })
 @OnMessagingEnabled
 @EnableConfigurationProperties(SleuthMessagingProperties.class)
 public class TraceMessagingAutoConfiguration {
@@ -184,6 +194,27 @@ public class TraceMessagingAutoConfiguration {
 		@Bean
 		TracingJmsBeanPostProcessor tracingJmsBeanPostProcessor(BeanFactory beanFactory) {
 			return new TracingJmsBeanPostProcessor(beanFactory);
+		}
+
+	}
+
+	@Configuration
+	@ConditionalOnProperty(value = "spring.sleuth.messaging.sqs.enabled", matchIfMissing = true)
+	@ConditionalOnClass(QueueMessageHandler.class)
+	protected static class SleuthSqsConfiguration {
+
+		@Bean
+		TracingMethodMessageHandlerAdapter tracingMethodMessageHandlerAdapter(
+				Tracing tracing,
+				Propagation.Getter<MessageHeaderAccessor, String> traceMessagePropagationGetter) {
+			return new TracingMethodMessageHandlerAdapter(tracing,
+					traceMessagePropagationGetter);
+		}
+
+		@Bean
+		QueueMessageHandlerFactory sqsQueueMessageHandlerFactory(
+				TracingMethodMessageHandlerAdapter tracingMethodMessageHandlerAdapter) {
+			return new SqsQueueMessageHandlerFactory(tracingMethodMessageHandlerAdapter);
 		}
 
 	}
@@ -410,8 +441,7 @@ class SleuthKafkaHeaderMapperBeanPostProcessor implements BeanPostProcessor {
 			throws BeansException {
 		if (bean instanceof SleuthDefaultKafkaHeaderMapper) {
 			return sleuthDefaultKafkaHeaderMapper(bean);
-		}
-		else if (bean instanceof DefaultKafkaHeaderMapper) {
+		} else if (bean instanceof DefaultKafkaHeaderMapper) {
 			((DefaultKafkaHeaderMapper) bean).setMapAllStringsOut(true);
 		}
 		return bean;
@@ -420,5 +450,47 @@ class SleuthKafkaHeaderMapperBeanPostProcessor implements BeanPostProcessor {
 	Object sleuthDefaultKafkaHeaderMapper(Object bean) {
 		return bean;
 	}
+}
 
+class SqsQueueMessageHandlerFactory extends QueueMessageHandlerFactory {
+
+	private TracingMethodMessageHandlerAdapter handlerAdapter;
+
+	SqsQueueMessageHandlerFactory(TracingMethodMessageHandlerAdapter handlerAdapter) {
+		this.handlerAdapter = handlerAdapter;
+	}
+
+	@Override
+	public QueueMessageHandler createQueueMessageHandler() {
+		if (CollectionUtils.isEmpty(getMessageConverters())) {
+			return new SqsQueueMessageHandler(handlerAdapter, Collections.emptyList());
+		}
+		return new SqsQueueMessageHandler(handlerAdapter, getMessageConverters());
+	}
+
+}
+
+class SqsQueueMessageHandler extends QueueMessageHandler {
+
+	private static final String LOGICAL_RESOURCE_ID = "LogicalResourceId"; // copied from QueueMessageHandler
+
+	private TracingMethodMessageHandlerAdapter handlerAdapter;
+
+	SqsQueueMessageHandler(TracingMethodMessageHandlerAdapter handlerAdapter,
+			List<MessageConverter> messageConverters) {
+		super(messageConverters);
+		this.handlerAdapter = handlerAdapter;
+	}
+
+	@Override
+	public void handleMessage(Message<?> message) throws MessagingException {
+		handlerAdapter.wrapMethodMessageHandler(message, super::handleMessage, this::messageSpanTagger);
+	}
+
+	private void messageSpanTagger(Span span, Message<?> message) {
+		span.remoteServiceName("sqs");
+		if (message.getHeaders().get(LOGICAL_RESOURCE_ID) != null) {
+			span.tag("sqs.queue_url", message.getHeaders().get(LOGICAL_RESOURCE_ID).toString());
+		}
+	}
 }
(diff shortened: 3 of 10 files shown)


devinsba:
@adriancole when you have a few minutes I'd appreciate feedback on the messaging model here. I'm not that familiar with it and our examples are not the most straightforward (rabbit/kafka)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TraceSpringMessagingAutoConfiguration.java
```diff
@@ -0,0 +1,46 @@
...
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.messaging.support.MessageHeaderAccessor;
+
+@Configuration
+@ConditionalOnClass(MessageHeaderAccessor.class)
+@OnMessagingEnabled
+@EnableConfigurationProperties(SleuthMessagingProperties.class)
+public class TraceSpringMessagingAutoConfiguration {
```
adriancole: nit nothing public unless required to be

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TracingMethodMessageHandlerAdapter.java
```diff
@@ -0,0 +1,62 @@
...
+		this.extractor = tracing.propagation().extractor(traceMessagePropagationGetter);
+	}
+
+	public void wrapMethodMessageHandler(Message<?> message,
+			Consumer<Message<?>> methodMessageHandler) {
+		TraceContextOrSamplingFlags extracted = extractor
+				.extract(new MessageHeaderAccessor(message));
+		Span span;
+		if (extracted.context() != null) {
+			span = tracer.joinSpan(extracted.context());
```
adriancole: we don't use join in messaging.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TracingMethodMessageHandlerAdapter.java
```diff
@@ -0,0 +1,62 @@
...
+	public TracingMethodMessageHandlerAdapter(Tracing tracing,
+			Propagation.Getter<MessageHeaderAccessor, String> traceMessagePropagationGetter) {
+		this.tracer = tracing.tracer();
+		this.extractor = tracing.propagation().extractor(traceMessagePropagationGetter);
+	}
+
+	public void wrapMethodMessageHandler(Message<?> message,
+			Consumer<Message<?>> methodMessageHandler) {
+		TraceContextOrSamplingFlags extracted = extractor
+				.extract(new MessageHeaderAccessor(message));
```
adriancole: this is odd, why we need to new up a message accessor per message.. maybe there's a reason for that

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TracingMethodMessageHandlerAdapter.java
```diff
@@ -0,0 +1,62 @@
...
+
+	public TracingMethodMessageHandlerAdapter(Tracing tracing,
+			Propagation.Getter<MessageHeaderAccessor, String> traceMessagePropagationGetter) {
+		this.tracer = tracing.tracer();
+		this.extractor = tracing.propagation().extractor(traceMessagePropagationGetter);
+	}
+
+	public void wrapMethodMessageHandler(Message<?> message,
+			Consumer<Message<?>> methodMessageHandler) {
+		TraceContextOrSamplingFlags extracted = extractor
```
adriancole: when extracting a message, we should also clear the properties like this https://github.com/openzipkin/brave/blob/master/instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/SpringRabbitTracing.java#L194

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TracingMethodMessageHandlerAdapter.java
```diff
@@ -0,0 +1,62 @@
...
+
+	private TraceContext.Extractor<MessageHeaderAccessor> extractor;
+
+	public TracingMethodMessageHandlerAdapter(Tracing tracing,
+			Propagation.Getter<MessageHeaderAccessor, String> traceMessagePropagationGetter) {
+		this.tracer = tracing.tracer();
+		this.extractor = tracing.propagation().extractor(traceMessagePropagationGetter);
+	}
+
+	public void wrapMethodMessageHandler(Message<?> message,
```
adriancole: in general I would just copy/paste this and re-use exact approach. https://github.com/openzipkin/brave/blob/master/instrumentation/spring-rabbit/src/main/java/brave/spring/rabbit/TracingRabbitListenerAdvice.java#L53 methodMessageHandler is same as methodInvocation.proceed();

below comments will drift into the right spot when above is done

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TraceMessagingAutoConfiguration.java
```diff
@@ -171,6 +181,28 @@ TracingJmsBeanPostProcessor tracingJmsBeanPostProcessor(BeanFactory beanFactory)
 
 	}
 
+	@Configuration
+	@ConditionalOnProperty(value = "spring.sleuth.messaging.sqs.enabled", matchIfMissing = true)
+	@ConditionalOnClass(QueueMessageHandler.class)
+	@AutoConfigureAfter(TraceSpringMessagingAutoConfiguration.class)
```
shakuzen: I don't think `@AutoConfigureAfter` has any effect here since I don't think `SleuthSqsConfiguration` itself is an auto-configuration class, but I could be wrong. I think the `@AutoConfigureAfter` would need to be on the top-level class, which is registered as the auto-configuration class.

devinsba: 👍

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TracingMethodMessageHandlerAdapter.java
devinsba: Yup, I missed the static methods all the way at the bottom of that class

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TraceSpringMessagingAutoConfiguration.java
devinsba: I'm not sure here, I guess spring is probably smart enough proxy this in the same package but I was really just following the example of the other Autoconfig classes

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TracingMethodMessageHandlerAdapter.java
devinsba: Done

devinsba:
@marcingrzejszczak I'm not sure why the test that is failing is failing. It's not in the stuff I modified so I am a bit lost. Maybe my test is not cleaning up after itself completely? Not sure

spring-cloud-sleuth-core/pom.xml
```diff
@@ -82,6 +82,11 @@
 			<groupId>org.springframework.cloud</groupId>
 			<artifactId>spring-cloud-commons</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-aws-messaging</artifactId>
```
devinsba: Not sure if I should bring in the starter of just the code?

marcingrzejszczak: Concrete dep is ok.

marcingrzejszczak:
> @marcingrzejszczak I'm not sure why the test that is failing is failing. It's not in the stuff I modified so I am a bit lost. Maybe my test is not cleaning up after itself completely? Not sure

Flakey test :sob:

marcingrzejszczak:
@devinsba if you rebase against master, you'll see that this test is ignored :| It's been flakey for some time so I've decided not to run it (and the problem is gone ;) )

devinsba:
I was going to rebase to squash all the checkstyle commits anyway so will do

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/messaging/ITTracingMethodMessageHandlerAdapter.java
```diff
@@ -0,0 +1,126 @@
...
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.messaging.support.GenericMessage;
+import org.springframework.test.context.junit4.SpringRunner;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.springframework.boot.test.context.SpringBootTest.WebEnvironment.NONE;
+
+@SpringBootTest(classes = ITTracingMethodMessageHandlerAdapter.TestingConfiguration.class, webEnvironment = NONE)
+@RunWith(SpringRunner.class)
+public class ITTracingMethodMessageHandlerAdapter {
```
marcingrzejszczak: All tests should finish with `Tests`, otherwise they might not be ran

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/messaging/SqsQueueMessageHandlerTest.java
```diff
@@ -0,0 +1,65 @@
...
+import org.mockito.junit.MockitoJUnit;
+import org.mockito.junit.MockitoRule;
+import org.mockito.quality.Strictness;
+
+import org.springframework.messaging.Message;
+import org.springframework.messaging.support.GenericMessage;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+public class SqsQueueMessageHandlerTest {
```
marcingrzejszczak: `Test` -> `Tests`

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TracingMethodMessageHandlerAdapter.java
```diff
@@ -0,0 +1,102 @@
...
+ * {@link org.springframework.messaging.handler.invocation.AbstractMethodMessageHandler} in a
+ * reusable way. When sub-classing a provider specific class of that type you would wrap the
+ * <pre>super.handleMessage(...)</pre> call with a call to this. See
+ * {@link org.springframework.cloud.sleuth.instrument.messaging.SqsQueueMessageHandler} for an
+ * example.
+ *
+ * This implementation also allows for supplying a {@link java.util.function.BiConsumer} instance
+ * that can be used to add queue specific tags and modifications to the span.
+ */
+class TracingMethodMessageHandlerAdapter {
```
marcingrzejszczak: Why not add yourself as an author? ;)

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/messaging/ITTracingMethodMessageHandlerAdapter.java
devinsba: FYI `ITTracingChannelInterceptor` has the wrong name too then

devinsba:
Yup, no docs yet. Only got it to a point I was confident over the weekend

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/messaging/ITTracingMethodMessageHandlerAdapter.java
marcingrzejszczak: :scream: let's fix it then!

devinsba: https://github.com/spring-cloud/spring-cloud-sleuth/pull/1225

devinsba:
Found a large bug, please don't merge

devinsba:
So it turns out it was only a bug in my slightly different internal impl for spring 4 (not boot). @marcingrzejszczak, @adriancole, @shakuzen please review when you have time

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
```diff
@@ -1368,6 +1368,13 @@ To block this feature, set `spring.sleuth.messaging.jms.enabled` to `false`.
 
 IMPORTANT: We don't support baggage propagation for JMS
 
+==== Spring Cloud AWS Messaging SQS
+
+We instrument `@SqsListener` which is provided by `org.springframework.cloud:spring-cloud-aws-messaging`
+so that tracing headers get extracted from the message and a trace gets put into the context.
+
+To block this feature, set `spring.sleuth.messageing.sqs.enabled` to `false`.
```
adriancole:
```suggestion
To block this feature, set `spring.sleuth.messaging.sqs.enabled` to `false`.
```

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TraceMessagingAutoConfiguration.java
```diff
@@ -366,3 +397,48 @@ private boolean typeMatches(Object bean) {
...
+		this.handlerAdapter = handlerAdapter;
+	}
+
+	@Override
+	public void handleMessage(Message<?> message) throws MessagingException {
+		handlerAdapter.wrapMethodMessageHandler(message, super::handleMessage, this::messageSpanTagger);
+	}
+
+	private void messageSpanTagger(Span span, Message<?> message) {
+		span.remoteServiceName("amazon-sqs");
```
adriancole: nit: is it necessary to prefix `amazon-` There are clones of SQS which possibly work regardless

```suggestion
		span.remoteServiceName("sqs");
```

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TraceMessagingAutoConfiguration.java
```diff
@@ -366,3 +397,48 @@ private boolean typeMatches(Object bean) {
...
+		if (message.getHeaders().get(LOGICAL_RESOURCE_ID) != null) {
+			span.tag("queue.url", message.getHeaders().get(LOGICAL_RESOURCE_ID).toString());
```
adriancole: wdyt?

```suggestion
			span.tag("sqs.queue_url", message.getHeaders().get(LOGICAL_RESOURCE_ID).toString());
```

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TracingMethodMessageHandlerAdapter.java
```diff
@@ -0,0 +1,104 @@
...
+ * @author Brian Devins-Suresh
+ */
+class TracingMethodMessageHandlerAdapter {
+
+	private Tracing tracing;
+	private Tracer tracer;
+	private TraceContext.Extractor<MessageHeaderAccessor> extractor;
+
+
+	TracingMethodMessageHandlerAdapter(Tracing tracing,
```
devinsba: Having read through more of the messaging code I'm not sure how reusable this would need to be. It seems most other things will be covered by the existing `spring-integration` style instrumentation. I'd love to get input from maybe someone who works on that project (`spring-messaging`/`spring-integration`) on whether maybe the SQS messaging is a special case

marcingrzejszczak:
Hey @devinsba, how is it going with this PR? Just FYI the build will be failing ATM due to issues with HATEOAS

marcingrzejszczak:
Hey @devinsba , I've fixed the merge conflicts. What do we do about this? :P

neoshadybeat:
Any updates about this PR?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TracingMethodMessageHandlerAdapter.java
devinsba: Looking through the code I can find, this does seem to become useful for `@MessageMapping` and `@SubscribeMapping` which don't look to be covered by the current websockets instrumentation so I think it is useful to leave this here

PR #1222, opened 2019-02-26
Refactored SkipPatternProviderConfigTest to use WebApplicationContextRunner instead of manually instantiated configurations

reta:
Hey @marcingrzejszczak , @shakuzen !

As discussed (in https://github.com/spring-cloud/spring-cloud-sleuth/pull/1208), it would be good to switch the  `SkipPatternProviderConfigTest ` to use application context runner instead of manually instantiated configurations classes.

Thanks!

### DIFF
--- a/spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/SkipPatternProviderConfigTest.java
+++ b/spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/SkipPatternProviderConfigTest.java
@@ -21,7 +21,6 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.Optional;
 import java.util.regex.Pattern;
-import java.util.stream.Collectors;
 
 import org.junit.Test;
 
@@ -98,8 +97,7 @@ public class SkipPatternProviderConfigTest {
 				.withPropertyValues("management.server.servlet.context-path=foo")
 				.run(context -> {
 					final String pattern = extractPattern(context);
-					then(pattern).contains("|foo.*");
-					then(extractAllPatterns(context)).contains("foo.*");
+					then(pattern).isIn(withDefaults("/actuator/(health|health/.*|info|info/.*)", "foo.*"));
 				});
 	}
 
@@ -119,8 +117,7 @@ public class SkipPatternProviderConfigTest {
 		contextRunner.withConfiguration(UserConfigurations.of(ServerPropertiesConfig.class))
 				.run(context -> {
 					final String pattern = extractPattern(context);
-					then(pattern).contains("/actuator/(health|health/.*|info|info/.*)");
-					then(extractAllPatterns(context)).contains("/actuator/(health|health/.*|info|info/.*)");
+					then(pattern).isIn(withDefaults("/actuator/(health|health/.*|info|info/.*)"));
 				});
 	}
 
@@ -130,8 +127,7 @@ public class SkipPatternProviderConfigTest {
 				.withPropertyValues("server.servlet.context-path=foo")
 				.run(context -> {
 					final String pattern = extractPattern(context);
-					then(pattern).contains("foo/actuator/(health|health/.*|info|info/.*)");
-					then(extractAllPatterns(context)).contains("foo/actuator/(health|health/.*|info|info/.*)");
+					then(pattern).isIn(withDefaults("foo/actuator/(health|health/.*|info|info/.*)"));
 				});
 	}
 
@@ -141,8 +137,7 @@ public class SkipPatternProviderConfigTest {
 				.withPropertyValues("management.endpoints.web.base-path=/")
 				.run(context -> {
 					final String pattern = extractPattern(context);
-					then(pattern).contains("/(health|health/.*|info|info/.*)");
-					then(extractAllPatterns(context)).contains("/(health|health/.*|info|info/.*)");
+					then(pattern).isIn(withDefaults("/(health|health/.*|info|info/.*)"));
 				});
 	}
 
@@ -153,8 +148,7 @@ public class SkipPatternProviderConfigTest {
 						"server.servlet.context-path=foo")
 				.run(context -> {
 					final String pattern = extractPattern(context);
-					then(pattern).contains("foo/(health|health/.*|info|info/.*)");
-					then(extractAllPatterns(context)).contains("foo/(health|health/.*|info|info/.*)");
+					then(pattern).isIn(withDefaults("foo/(health|health/.*|info|info/.*)"));
 				});
 	}
 
@@ -165,8 +159,7 @@ public class SkipPatternProviderConfigTest {
 						"management.server.port=0", "server.servlet.context-path=foo")
 				.run(context -> {
 					final String pattern = extractPattern(context);
-					then(pattern).contains("/(health|health/.*|info|info/.*)");
-					then(extractAllPatterns(context)).contains("/(health|health/.*|info|info/.*)");
+					then(pattern).isIn(withDefaults("/(health|health/.*|info|info/.*)"));
 				});
 	}
 
@@ -177,19 +170,7 @@ public class SkipPatternProviderConfigTest {
 						"server.servlet.context-path=foo")
 				.run(context -> {
 					final String pattern = extractPattern(context);
-					then(pattern).contains("foo/mgt/(health|health/.*|info|info/.*)");
-					then(extractAllPatterns(context)).contains("foo/mgt/(health|health/.*|info|info/.*)");
-				});
-	}
-
-	@Test
-	public void should_return_endpoints_with_actuator_default_context_path() {
-		contextRunner.withConfiguration(UserConfigurations.of(ServerPropertiesConfig.class))
-				.withPropertyValues("server.servlet.context-path=foo")
-				.run(context -> {
-					final String pattern = extractPattern(context);
-					then(pattern).contains("foo/actuator/(health|health/.*|info|info/.*)");
-					then(extractAllPatterns(context)).contains("foo/actuator/(health|health/.*|info|info/.*)");
+					then(pattern).isIn(withDefaults("foo/mgt/(health|health/.*|info|info/.*)"));
 				});
 	}
 
@@ -199,8 +180,7 @@ public class SkipPatternProviderConfigTest {
 				.withPropertyValues("management.server.port=0", "server.servlet.context-path=foo")
 				.run(context -> {
 					final String pattern = extractPattern(context);
-					then(pattern).contains("/actuator/(health|health/.*|info|info/.*)");
-					then(extractAllPatterns(context)).contains("/actuator/(health|health/.*|info|info/.*)");
+					then(pattern).isIn(withDefaults("/actuator/(health|health/.*|info|info/.*)"));
 				});
 	}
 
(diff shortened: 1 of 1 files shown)


spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/SkipPatternProviderConfigTest.java
```diff
@@ -97,190 +125,169 @@ public void should_return_empty_when_no_endpoints() {
...
+			.run(context -> {
+				SkipPatternProvider skipPatternProvider = context.getBean(SkipPatternProvider.class);
+				Pattern pattern = skipPatternProvider.skipPattern();
+				then(pattern.pattern()).contains("foo/(health|health/.*|info|info/.*)");
+			});
 	}
-
+	
 	@Test
-	public void should_return_endpoints_with_actuator_context_path_set_to_root() {
```
reta: This test cases is covered by `should_return_endpoints_without_context_path_and_base_path_set_to_root` and `should_return_endpoints_with_context_path_and_base_path_set_to_ot_different_port`

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/SkipPatternProviderConfigTest.java
```diff
@@ -97,190 +116,163 @@ public void should_return_empty_when_no_endpoints() {
...
-				.skipPatternForActuatorEndpointsSamePort(properties,
-						webEndpointProperties, endpointsSupplier)
-				.skipPattern();
-
-		then(pattern).isNotEmpty();
-		then(pattern.get().pattern())
-				.isEqualTo("/actuator/(info|info/.*|health|health/.*)");
+		contextRunner.withConfiguration(UserConfigurations.of(EndpointConfig.class))
+				.withPropertyValues("management.endpoint.health.enabled=true",
+						"management.endpoint.info.enabled=true",
```
shakuzen: Aren't the endpoints enabled by default? Meaning the result is the same without these properties.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/SkipPatternProviderConfigTest.java
```diff
@@ -97,190 +116,163 @@ public void should_return_empty_when_no_endpoints() {
...
+						"management.endpoints.web.exposure.include=info,health",
+						"management.endpoints.web.base-path=/",
+						"management.servet.port=0", "server.servlet.context-path=foo")
+				.run(context -> {
+					SkipPatternProvider skipPatternProvider = context
+							.getBean(SkipPatternProvider.class);
+					Pattern pattern = skipPatternProvider.skipPattern();
+					then(pattern.pattern()).contains("/(health|health/.*|info|info/.*)");
```
shakuzen: We have to be careful with using `contains`. It is safest to use `isEqualTo` and just make a constant in this class for the default pattern to concatenate with the interesting portion. I bring this up here because there is a typo in the properties ("servet" → "server") that changes the resulting pattern (/ vs foo/) but the assert as is doesn't catch it.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/SkipPatternProviderConfigTest.java
```diff
@@ -97,190 +116,163 @@ public void should_return_empty_when_no_endpoints() {
...
-		then(patternDifferentPort).isNotEmpty();
-		then(patternDifferentPort.get().pattern())
-				.isEqualTo("/actuator/(info|info/.*|health|health/.*)");
+	@Test
+	public void should_return_endpoints_with_actuator_default_context_path_different_port() {
+		contextRunner.withConfiguration(UserConfigurations.of(EndpointConfig.class))
+				.withPropertyValues("management.endpoint.health.enabled=true",
+						"management.endpoint.info.enabled=true",
+						"management.endpoints.web.exposure.include=info,health",
+						"management.servet.port=0", "server.servlet.context-path=foo")
```
shakuzen: This is in a few places.

```suggestion
						"management.server.port=0", "server.servlet.context-path=foo")
```

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/SkipPatternProviderConfigTest.java
```diff
@@ -301,29 +293,13 @@ private SingleSkipPattern bar() {
...
 
-	private ExposableWebEndpoint createEndpoint(final String name) {
-		return new ExposableWebEndpoint() {
-
-			@Override
-			public String getRootPath() {
-				return name;
-			}
+	@Configuration
+	@ImportAutoConfiguration({ DispatcherServletAutoConfiguration.class,
```
shakuzen: Why don't we just add all these auto-configuration classes to the WebApplicationContextRunner? Then wouldn't the custom configuration class be unnecessary altogether?

Edit: I tried and it indeed doesn't work, but I don't understand why it is different.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/SkipPatternProviderConfigTest.java
reta: I believe they are, I could simplify it, thanks!

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/SkipPatternProviderConfigTest.java
reta: Yeah, this is a tricky part since now we test all patterns together, let me try to find the better way (if any) to ensure we test the right thing.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/SkipPatternProviderConfigTest.java
reta: Thanks for catching it, I am wondering how come test passes, will check it out

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/SkipPatternProviderConfigTest.java
reta: This is a good question, since I tried this first, and it didn't work out as well. There are few reasons (I think), since some tests do checks when there are no web endpoints. Another thing is `@EnableConfigurationProperties` which triggers the conditional match on `ServerProperties` bean (otherwise it is not present).

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/SkipPatternProviderConfigTest.java
```diff
@@ -39,26 +50,32 @@
...
-		sleuthWebProperties.setSkipPattern("foo.*|bar.*");
-		Pattern pattern = new TraceWebAutoConfiguration.DefaultSkipPatternConfig()
-				.defaultSkipPatternBean(sleuthWebProperties).skipPattern().get();
-
-		then(pattern.pattern()).isEqualTo("foo.*|bar.*");
+		contextRunner.withPropertyValues("spring.sleuth.web.skip-pattern=foo.*|bar.*")
+				.run(context -> {
+					SkipPatternProvider skipPatternProvider = context
+							.getBean(SkipPatternProvider.class);
+					Pattern pattern = skipPatternProvider.skipPattern();
```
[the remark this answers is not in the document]
reta: :+1:

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/SkipPatternProviderConfigTest.java
reta: So I did this @shakuzen, I added the assertion on the presence of the exact pattern in the list of `SingleSkipPattern` beans (which `SkipPatternProvider` is built of), it caught the issue with misspelled management port immediately (should have done it from the beginning). Thanks!

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/SkipPatternProviderConfigTest.java
reta: Addressed now, the checks are more strict, thanks!

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/SkipPatternProviderConfigTest.java
reta: @shakuzen I cut it to a bare minimum, do you think we could keep it? I didn't figure out the other way to ``ServerProperties`, might be part of another configuration.

reta:
@shakuzen may you take a look please? Thanks!

reta:
@shakuzen Being careful was never a bad idea, I changed all matches in favor of exact ones (referencing `SleuthWebProperties.DEFAULT_SKIP_PATTERN`) and removed duplicated test  `should_return_endpoints_with_actuator_default_context_path`, mind please taking a look? Thanks a lot for the review!

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/SpringDataInstrumentationTests.java
```diff
@@ -45,7 +45,7 @@
 import org.springframework.core.env.Environment;
 import org.springframework.data.jpa.repository.JpaRepository;
 import org.springframework.data.rest.core.annotation.RepositoryRestResource;
-import org.springframework.hateoas.PagedResources;
+import org.springframework.hateoas.PagedModel;
```
reta: @shakuzen  This change is unrelated but fixes the compilation issue on master (due to Spring HATEAOS upgrade), please let me know if you want me to revert or keep it (my only intention was to see the CI build status).

shakuzen: @marcingrzejszczak will probably be happy to accept a fix :)

shakuzen:
@reta thanks for bearing with the build issues and my review.

> PS: The order of patterns is non-deterministic, sometimes default goes first, sometimes it comes last, the assertions are exact but with 'one of' kind of logic.

I'm having trouble getting this branch of Sleuth to compile on my machine now (probably some dependency I need to redownload but I'm not on great internet right now). Anyways, though, since the order is non-deterministic, the `tail` pattern(s) could also be in any position. This is getting to be quite the pain to exactly compare. Maybe it would be better to assert on the `List<SingleSkipPattern>` that it `containsExactlyInAnyOrder` the expected patterns, then. What do you think? Otherwise we could make the order deterministic (`.sorted()` before joining) - but that would be changing the main code only for the purpose of the tests since regex won't care about the order.

reta:
@shakuzen Yeah, sure, done :) I think this is the cleanest and most reliable approach so far :) What do you think?

reta:
Thanks a lot for review, @shakuzen !

reta:
@marcingrzejszczak mind please taking a look? The build failure / test cases seem  to be unrelated to this pull request, thanks!

marcingrzejszczak:
It's failing ATM due to issues with HATEOAS. Once they're resolved I'll merge this.

marcingrzejszczak:
Great work @reta

PR #1225, opened 2019-03-04
Rename test class so it runs

devinsba:


devinsba:
Looks like another test flake

adriancole:
Thanks @devinsba

PR #1226, opened 2019-03-05
performance enhancement, newScope and other methods is called too frequently

aftersss:
We are now using `spring-cloud-gateway`, I make two pressure tests on it, one sample with `spring-cloud-sleuth`, another without `spring-cloud-sleuth`, both of these two samples are tested after five minutes of warm up, and I used `jmeter` with 50 threads to test it,  both of them run to 1300 qps, my gateway program runs at a linux server with 4 cores and 2 GB memory, this is my result:

when I integrate `spring-cloud-gateway` with `spring-cloud-sleuth`,cpu usage is about 60%:
image

when I integrate `spring-cloud-gateway` without `spring-cloud-sleuth`,cpu usage is about 25%:
image

As you can see `spring-cloud-sleuth` costs too much cpu.
I tried to find out why this happens, and I found that the key point is  `Slf4jCurrentTraceContext.newScope` is called too much times, every http request will call `Slf4jCurrentTraceContext.newScope` for at least 100 times,  this means that every http request will call `MDC` for at least 100*24 times, I also found that `Slf4jCurrentTraceContext.newScope`s span argument is often the same span, So this pull request tries to fix this problem.

I found that `scannable.name()` costs too much cpu too, because this will be called too frequently, and implement of this method is using `regex`, but actually this name is useless, so I replace this with null.

`beanFactory.getBean` is also called too frequently, so I cache the result of `beanFactory.getBean`.

After the above modification, I run the test again, and the cpu usage drop to about 30%, performance rise up by 50 percent:
image

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *      http://www.apache.org/licenses/LICENSE-2.0
+ *      https://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
@@ -41,6 +41,8 @@ public abstract class ReactorSleuth {
 
 	private static final Log log = LogFactory.getLog(ReactorSleuth.class);
 
+	private static volatile boolean CONTEXT_REFRESHED = false;
+
 	/**
 	 * Return a span operator pointcut given a {@link BeanFactory}. This can be used in reactor
 	 * via {@link reactor.core.publisher.Flux#transform(Function)}, {@link
@@ -141,8 +143,15 @@ public abstract class ReactorSleuth {
 	}
 
 	private static boolean contextRefreshed(BeanFactory beanFactory) {
+		if (CONTEXT_REFRESHED) {
+			return true;
+		}
 		try {
-			return beanFactory.getBean(ApplicationContextRefreshedListener.class).isRefreshed();
+			boolean contextRefreshed = beanFactory.getBean(ApplicationContextRefreshedListener.class).isRefreshed();
+			if (contextRefreshed) {
+				CONTEXT_REFRESHED = true;
+			}
+			return contextRefreshed;
 		} catch (NoSuchBeanDefinitionException e) {
 			return false;
 		}
@@ -154,7 +163,7 @@ public abstract class ReactorSleuth {
 				beanFactory,
 				sub,
 				sub.currentContext(),
-				scannable.name()) {
+				null) {
 			@Override SpanSubscription newCoreSubscriber(Tracing tracing) {
 				return new ScopePassingSpanSubscriber<T>(
 						sub,
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
@@ -93,7 +93,7 @@ public abstract class ReactorSleuth {
 				beanFactory,
 				sub,
 				sub.currentContext(),
-				scannable.name());
+				scannable);
 	}
 
 	/**
@@ -163,7 +163,7 @@ public abstract class ReactorSleuth {
 				beanFactory,
 				sub,
 				sub.currentContext(),
-				null) {
+				scannable) {
 			@Override SpanSubscription newCoreSubscriber(Tracing tracing) {
 				return new ScopePassingSpanSubscriber<T>(
 						sub,
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingSpanSubscriber.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingSpanSubscriber.java
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *      http://www.apache.org/licenses/LICENSE-2.0
+ *      https://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingSpanSubscriber.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingSpanSubscriber.java
@@ -21,6 +21,7 @@ import java.util.concurrent.atomic.AtomicBoolean;
 import brave.Span;
 import brave.Tracing;
 import brave.propagation.CurrentTraceContext;
+import brave.propagation.TraceContext;
 import reactor.util.context.Context;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -38,7 +39,7 @@ final class ScopePassingSpanSubscriber<T> extends AtomicBoolean implements SpanS
 
 	private static final Log log = LogFactory.getLog(ScopePassingSpanSubscriber.class);
 
-	private final Span span;
+	private final TraceContext spanTraceContext;
 	private final Subscriber<? super T> subscriber;
 	private final Context context;
 	private final Tracing tracing;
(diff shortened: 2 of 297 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/log/Slf4jCurrentTraceContext.java
```diff
@@ -66,9 +66,21 @@ public static Slf4jCurrentTraceContext create(CurrentTraceContext delegate) {
 	}
 
 	@Override public Scope newScope(@Nullable TraceContext currentSpan) {
+		TraceContext previousSpan = get();
```
adriancole: the call site into this should already be using maybeScope, which almost identically replicates the logic here..

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/SpanSubscriptionProvider.java
```diff
@@ -38,7 +38,7 @@
 	final Subscriber<? super T> subscriber;
 	final Context context;
 	final String name;
-	private Tracing tracing;
+	private static volatile Tracing TRACING;
```
adriancole: I would suspect this is not generally good, but could be convinced if there is some assumptions in spring boot. If there are some practice like this, it should be in big comments I think.

aftersss: I have revert this code back, but I still think this can be improved, assume if our program run to 5000qps, `beanFactory.getBean` will run to 50W qps, this will more or less affect performance, this can be avoided.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/log/Slf4jCurrentTraceContext.java
aftersss: actually `maybeScope` is not being used, In recent commits, I have modified the code to use `maybeScope`.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/SpanSubscriptionProvider.java
adriancole: I think understanding what's going on with the lazy lookup is important anyway, and if there are other approaches.

maybe @marcingrzejszczak can shed some light about the pattern here and very many other places about lazy bean lookup so that the perf issues can be examined but not addressed in a risk of lockups. others also can feel free to chip in..

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingSpanSubscriber.java
```diff
@@ -59,25 +60,25 @@
 
 	@Override public void onSubscribe(Subscription subscription) {
 		this.s = subscription;
-		try (Tracer.SpanInScope inScope = this.tracer.withSpanInScope(this.span)) {
+		try (CurrentTraceContext.Scope inScope = this.tracing.currentTraceContext().maybeScope(this.spanTraceContext)) {
```
adriancole: if this change is about efficiency, probably you can inject the `CurrentTraceContext` object and use it directly. I'm not sure any of this code needs the Span object.

cc also @dmexe as there's a beautiful, if unrelated, change going on in reactor as well https://github.com/reactor/reactor-netty/pull/622#issuecomment-468353910

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingSpanSubscriber.java
```diff
@@ -38,18 +39,18 @@
...
 	private Subscription s;
 
 	ScopePassingSpanSubscriber(Subscriber<? super T> subscriber, Context ctx, Tracing tracing) {
 		this.subscriber = subscriber;
-		this.tracer = tracing.tracer();
+		this.tracing = tracing;
 		Span root = ctx != null ?
-				ctx.getOrDefault(Span.class, this.tracer.currentSpan()) : null;
-		this.span = root;
+				ctx.getOrDefault(Span.class, tracing.tracer().currentSpan()) : null;
```
adriancole: ahh Span is used here.. maybe you'll need the tracer afterall...

aftersss: tracing.tracer().currentSpan() is only called one time by the constructor, so I think there is no need to use tracer

aftersss:
Sorry, I forgot to figure out how I found these problems. 
I used this tool to profile my program: `https://github.com/jvm-profiling-tools/async-profiler`, and the following file is the profiler result:

as you can see, there are many stack traces which costs too much cpu like this:

```
Total: 35052025 (0.09%)  samples: 35
  [ 0] java.util.Collections$SynchronizedMap.get
  [ 1] ch.qos.logback.classic.util.LogbackMDCAdapter.get
  [ 2] org.slf4j.MDC.get
  [ 3] org.springframework.cloud.sleuth.log.Slf4jCurrentTraceContext.newScope
  [ 4] brave.Tracer.withSpanInScope
  [ 5] org.springframework.cloud.sleuth.instrument.reactor.ScopePassingSpanSubscriber.onNext
  [ 6] reactor.core.publisher.FluxIterable$IterableSubscription.slowPath
  [ 7] reactor.core.publisher.FluxIterable$IterableSubscription.request
  [ 8] org.springframework.cloud.sleuth.instrument.reactor.ScopePassingSpanSubscriber.request
  [ 9] reactor.core.publisher.FluxConcatMap$ConcatMapImmediate.onSubscribe
  [10] org.springframework.cloud.sleuth.instrument.reactor.ScopePassingSpanSubscriber.onSubscribe
......
```

another stack trace like the following costs too much cpu too:
```
Total: 40002796 (0.10%)  samples: 40
  [ 0] java.util.regex.Pattern.atom
  [ 1] java.util.regex.Pattern.sequence
  [ 2] java.util.regex.Pattern.expr
  [ 3] java.util.regex.Pattern.compile
  [ 4] java.util.regex.Pattern.<init>
  [ 5] java.util.regex.Pattern.compile
  [ 6] java.lang.String.replaceAll
  [ 7] reactor.core.Scannable.operatorName
  [ 8] reactor.core.Scannable.name
  [ 9] org.springframework.cloud.sleuth.instrument.reactor.ReactorSleuth.scopePassingSpanSubscription
  [10] org.springframework.cloud.sleuth.instrument.reactor.ReactorSleuth.lambda$null$2
  [11] org.springframework.cloud.sleuth.instrument.reactor.ReactorSleuth$$Lambda$437.1403649277.apply
...
```

And watch this stack trace, this shows that `tracer.withSpanInScope` will be invoked for many times:
image

adriancole:
I would try changing to log4j2 instead of doing any more work. the
synchronization isn't necessarily going to be in that impl.

marcingrzejszczak:
cc @bsideup @smaldini WDYT?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
```diff
@@ -154,7 +163,7 @@ private static boolean contextRefreshed(BeanFactory beanFactory) {
 				beanFactory,
 				sub,
 				sub.currentContext(),
-				scannable.name()) {
+				null) {
```
bsideup: it is used to name the spans. I suggest to pass `scannable` and call `scannable.name()` only when necessary

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/SpanSubscriptionProvider.java
```diff
@@ -57,7 +60,7 @@
 	}
 
 	SpanSubscription<T> newCoreSubscriber(Tracing tracing) {
-		return new SpanSubscriber<>(this.subscriber, this.context, tracing, this.name);
+		return new SpanSubscriber<>(this.subscriber, this.context, tracing, this.scannable.name());
```
bsideup: `SpanSubscriber` could also benefit from an on-demand `scannable.name()`

aftersss: SpanSubscriber uses `name`  in the constructor, so I think passing `scannable` to `SpanSubscriber` is meaningless.

bsideup: yes, you're right! 👍

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/log/Slf4JSpanLoggerTest.java
```diff
@@ -58,29 +59,119 @@ public void setup() {
...
+		assertMDCInfoEqualToSpanInfo(span.context());
+
+		scope.close();
+
+		assertMDCInfoNullOrEmpty();
+	}
+
+	@Test
+	public void should_set_entries_to_mdc_from_two_spans1() throws Exception {
+		Scope scope = this.slf4jCurrentTraceContext.newScope(this.span.context());
```
bsideup: just a small minor suggestion:
`Scope` implements `Closeable` and can be used with `try-with-resources` style

aftersss: done, and the test failure seems not my problem?

marcingrzejszczak:
Have you rebased against master?

aftersss:
> Have you rebased against master?

Rebase is useless.


```
java.lang.AssertionError: 

Expecting:
 <{"channel"="errorChannel"}>
to contain:
 <[class=TestBeanWithScheduledMethod, method=scheduledMethod]>
but could not find:
 <[class=TestBeanWithScheduledMethod, method=scheduledMethod]>
```

this test failure looks like that this span is from some test case of `messaging`, maybe there are some thread safe problems in the test cases, so I added @NotThreadSafe annotation to `TracingOnScheduledTests`, it is effective.

marcingrzejszczak:
Great work. Thanks a lot!

PR #1267, opened 2019-03-21
Enables Redis tracing by default.

### LINKED ISSUE #1233: Enable lettuce tracing by default
Since there is [built-in Brave tracing support in Lettuce ](https://github.com/lettuce-io/lettuce-core/tree/master/src/main/java/io/lettuce/core/tracing) , so it seems reasonable to enable the Lettuce Redis tracing by default in Spring Cloud Sleuth.

chang-chao:
Currently, only Lettcue Redis client is supported.
Closes #1233 .

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/redis/TraceRedisAutoConfiguration.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/redis/TraceRedisAutoConfiguration.java
@@ -5,7 +5,7 @@
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
  *
- *      http://www.apache.org/licenses/LICENSE-2.0
+ *      https://www.apache.org/licenses/LICENSE-2.0
  *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/redis/TraceRedisAutoConfiguration.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/redis/TraceRedisAutoConfiguration.java
@@ -75,7 +75,9 @@ class TraceLettuceClientResourcesBeanPostProcessor implements BeanPostProcessor 
 			throws BeansException {
 		if (bean instanceof ClientResources) {
 			ClientResources cr = (ClientResources) bean;
-			if (cr.tracing() == null) {
+			// tracing of ClientResources instance created by default is `disabled()`
+			if (cr.tracing() == null
+					|| cr.tracing() == io.lettuce.core.tracing.Tracing.disabled()) {
 				return cr.mutate().tracing(BraveTracing.create(this.tracing)).build();
 			}
 		}
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/redis/TraceRedisAutoConfiguration.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/redis/TraceRedisAutoConfiguration.java
@@ -79,12 +79,16 @@ class TraceLettuceClientResourcesBeanPostProcessor implements BeanPostProcessor 
 		if (bean instanceof ClientResources) {
 			ClientResources cr = (ClientResources) bean;
 			if (!cr.tracing().isEnabled()) {
-				log.debug(
-						"Lettuce ClientResources bean is auto-configured to enable tracing.");
+				if (log.isDebugEnabled()) {
+					log.debug(
+							"Lettuce ClientResources bean is auto-configured to enable tracing.");
+				}
 				return cr.mutate().tracing(BraveTracing.create(this.tracing)).build();
 			}
-			log.debug(
-					"Lettuce ClientResources bean is skipped for auto-configuration because tracing was already enabled.");
+			if (log.isDebugEnabled()) {
+				log.debug(
+						"Lettuce ClientResources bean is skipped for auto-configuration because tracing was already enabled.");
+			}
 		}
 		return bean;
 	}
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/redis/TraceRedisAutoConfiguration.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/redis/TraceRedisAutoConfiguration.java
@@ -24,7 +24,6 @@ import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.config.BeanPostProcessor;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.cloud.sleuth.autoconfig.TraceAutoConfiguration;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
@@ -43,7 +42,7 @@ import org.springframework.context.annotation.Configuration;
 public class TraceRedisAutoConfiguration {
 
 	@Configuration
-	@ConditionalOnClass(ClientResources.class)
+	@ConditionalOnBean(ClientResources.class)
 	static class LettuceConfig {
 
 		@Bean
@@ -75,9 +74,7 @@ class TraceLettuceClientResourcesBeanPostProcessor implements BeanPostProcessor 
 			throws BeansException {
 		if (bean instanceof ClientResources) {
 			ClientResources cr = (ClientResources) bean;
-			// tracing of ClientResources instance created by default is `disabled()`
-			if (cr.tracing() == null
-					|| cr.tracing() == io.lettuce.core.tracing.Tracing.disabled()) {
+			if (!cr.tracing().isEnabled()) {
 				return cr.mutate().tracing(BraveTracing.create(this.tracing)).build();
 			}
 		}
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/redis/TraceRedisAutoConfiguration.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/redis/TraceRedisAutoConfiguration.java
@@ -19,6 +19,8 @@ package org.springframework.cloud.sleuth.instrument.redis;
 import brave.Tracing;
 import io.lettuce.core.resource.ClientResources;
 import io.lettuce.core.tracing.BraveTracing;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 
 import org.springframework.beans.BeansException;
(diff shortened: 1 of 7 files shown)


marcingrzejszczak:
Looks great! Can you add a test that verifies that the processor gets executed?

chang-chao:
@marcingrzejszczak A test case was added.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/redis/TraceRedisAutoConfiguration.java
```diff
@@ -0,0 +1,85 @@
...
+ * @since 2.2.0
+ */
+@Configuration
+@OnRedisEnabled
+@ConditionalOnBean(Tracing.class)
+@AutoConfigureAfter({ TraceAutoConfiguration.class })
+public class TraceRedisAutoConfiguration {
+
+	@Configuration
+	@ConditionalOnClass(ClientResources.class)
```
mp911de: I'm a bit torn on the conditionality. `TraceLettuceClientResourcesBeanPostProcessor` conditionally configures `ClientResources`  only if no `tracing` is set. If `ClientResources.tracing` is set, `TraceLettuceClientResourcesBeanPostProcessor` just skips configuration without any external visibility that configuration was skipped.

I would propose to add another `@Condition` that introspects a `ClientResources` bean and only then the `TraceLettuceClientResourcesBeanPostProcessor` gets registered. Configuration should happen if `ClientResources.tracing().isEnabled()` returns `false`.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/redis/TraceRedisAutoConfiguration.java
```diff
@@ -0,0 +1,85 @@
...
+			throws BeansException {
+		return bean;
+	}
+
+	@Override
+	public Object postProcessAfterInitialization(Object bean, String beanName)
+			throws BeansException {
+		if (bean instanceof ClientResources) {
+			ClientResources cr = (ClientResources) bean;
+			if (cr.tracing() == null) {
```
mp911de: `tracing()` is never `null`. If tracing is disable, then `cr.tracing().isEnabled()` returns `false`.

I would remove the condition and always configure tracing. Conditionality should happen on a different level.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/redis/TraceRedisAutoConfiguration.java
```diff
@@ -0,0 +1,87 @@
...
+/**
+ * {@link org.springframework.boot.autoconfigure.EnableAutoConfiguration
+ * Auto-configuration} enables Redis span information propagation.
+ *
+ * @author Chao Chang
+ * @since 2.2.0
+ */
+@Configuration
+@OnRedisEnabled
+@ConditionalOnBean(Tracing.class)
```
mp911de: It would make sense to add also `@ConditionalOnBean(ClientResources.class)` to ensure a bean is available. One could exclude `RedisAutoConfiguration`. The condition would also report to Spring Boot's conditional reporter and improve the debugging experience.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/redis/TraceRedisAutoConfiguration.java
chang-chao: Do you mean the code like the following? 
I tried but `TraceLettuceClientResourcesBeanPostProcessor.postProcessAfterInitialization()` was never called for the ClientResources bean created in my test case.
I am not sure what I was missing.

```
public class TraceRedisAutoConfiguration {

	@Configuration
	@ConditionalOnBean(ClientResources.class)
	static class LettuceConfig {

		@Bean
		@Conditional(TracingIsDisabledInLettuceClientResourcesCondition.class)
		static TraceLettuceClientResourcesBeanPostProcessor traceLettuceClientResourcesBeanPostProcessor(
				Tracing tracing) {
			return new TraceLettuceClientResourcesBeanPostProcessor(tracing);
		}

	}

}


class TracingIsDisabledInLettuceClientResourcesCondition extends SpringBootCondition
		implements ConfigurationCondition {

	@Override
	public ConfigurationPhase getConfigurationPhase() {
		return ConfigurationPhase.REGISTER_BEAN;
	}

	@Override
	public ConditionOutcome getMatchOutcome(ConditionContext context,
			AnnotatedTypeMetadata metadata) {
		Assert.isInstanceOf(DefaultListableBeanFactory.class, context.getBeanFactory());
		DefaultListableBeanFactory listableBeanFactory = (DefaultListableBeanFactory) context
				.getBeanFactory();
		ClientResources clientResources = listableBeanFactory
				.getBean(ClientResources.class);

		if (clientResources.tracing().isEnabled()) {
			return ConditionOutcome.noMatch(
					"Tracing in the created ClientResources bean is already enabled.");
		}
		return ConditionOutcome.match();

	}

}

```

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/redis/TraceRedisAutoConfiguration.java
chang-chao: I added `@ConditionalOnBean(ClientResources.class)` to `LettuceConfig` class because this class is the configuration intended for Lettuce, while `TraceRedisAutoConfiguration` is intended for the configuration for all Redis clients, say, Jedis in the future.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/redis/TraceRedisAutoConfiguration.java
chang-chao: Now `cr.tracing().isEnabled()` is used for checking whether tracing is enabled.
As for the condition level, please see my reply  to the previous comment.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/redis/TraceRedisAutoConfiguration.java
mp911de: It's not just about a classpath dependency but also to check whether there is a `ClientResources` bean.

chang-chao: Now I added `@ConditionalOnBean(value = { Tracing.class, ClientResources.class })` to `TraceRedisAutoConfiguration` class, making  `TraceRedisAutoConfiguration` be configured only in case of there is a `ClientResources ` bean, is this what advised to do?

mp911de: Yes, that's exactly how it is supposed to work.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/redis/TraceRedisAutoConfiguration.java
mp911de: That looks in the direction how it is supposed to work. Enable debug mode as it should print why the bean was/was not configured.

chang-chao: The problem with the above code is that, when we get the `ClientResources ` bean to introspect in TracingIsDisabledInLettuceClientResourcesCondition , the bean will be created and BeanPostProcessors that created before that will be called, at that time,  `TraceLettuceClientResourcesBeanPostProcessor ` is not created yet.
After the `TraceLettuceClientResourcesBeanPostProcessor `  bean is created later, because the creation process of `ClientResources` bean has already finished,  no further `TraceLettuceClientResourcesBeanPostProcessor ` will be called against it.
Would you mind give me some hint on how the following logic can be implemented?
>add another @Condition that introspects a ClientResources bean and only then the TraceLettuceClientResourcesBeanPostProcessor gets registered.

mp911de: I see. I suppose in that case we just need to accept the fact that we cannot use a `@Condition` to introspect `ClientResources`.  I'd suggest then adding logging on `DEBUG` level to `TraceLettuceClientResourcesBeanPostProcessor` for the two cases of configuration and non-configuration. By enabling debug logging, users can at least find some indication on the log, why tracing was configured or why it was not configured by `TraceLettuceClientResourcesBeanPostProcessor`.

chang-chao: Debug logs were added.

mp911de:
Thanks for all your effort @chang-chao. This looks now good to me. @marcingrzejszczak your turn.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/redis/TraceRedisAutoConfigurationTests.java
```diff
@@ -0,0 +1,90 @@
+/*
+ * Copyright 2013-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
```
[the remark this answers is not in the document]
chang-chao: Changed

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/redis/TraceRedisAutoConfiguration.java
```diff
@@ -0,0 +1,92 @@
...
+	public Object postProcessAfterInitialization(Object bean, String beanName)
+			throws BeansException {
+		if (bean instanceof ClientResources) {
+			ClientResources cr = (ClientResources) bean;
+			if (!cr.tracing().isEnabled()) {
+				log.debug(
+						"Lettuce ClientResources bean is auto-configured to enable tracing.");
+				return cr.mutate().tracing(BraveTracing.create(this.tracing)).build();
+			}
+			log.debug(
```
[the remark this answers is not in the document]
chang-chao: Updated.

marcingrzejszczak:
Sweeeet! Now let's wait for the master to get fixed and I'll merge it.

marcingrzejszczak:
Great work @chang-chao ! Congrats!

datagitlies:
@chang-chao @marcingrzejszczak this is great! When [lettuce-core 5.2.0](https://github.com/lettuce-io/lettuce-core/milestone/51?closed=1) is released it might make sense to further enhance this code by allowing users to [include / exclude Redis command args from the span tags](https://github.com/lettuce-io/lettuce-core/pull/920) since including them can be problematic for a few reasons...

shakuzen:
@datagitlies could you make a separate issue to track that so it doesn't get lost/forgotten?

Issue #1345, opened 2019-05-02
The "nativeHeaders" Map modification in MessageHeaderPropagation can cause ConcurrentModificationException


marcingrzejszczak:
Thanks to @artembilan :

> So, my point is to create a new map and then add a new entry as a copy of an existing one. Something like what is done in the `NativeMessageHeaderAccessor`":

```java
@SuppressWarnings("unchecked")
    @Nullable
    protected Map<String, List<String>> getNativeHeaders() {
        return (Map<String, List<String>>) getHeader(NATIVE_HEADERS);
    }

    /**
     * Return a copy of the native header values or an empty map.
     */
    public Map<String, List<String>> toNativeHeaderMap() {
        Map<String, List<String>> map = getNativeHeaders();
        return (map != null ? new LinkedMultiValueMap<>(map) : Collections.emptyMap());
    }
```

> So, we would never abuse an existing map for modification which can be shared between different messages in different threads

artembilan:
The code we talk about is in the `MessageHeaderPropagation` and looks like:
```
private void doPut(MessageHeaderAccessor accessor, String key, String value) {
		accessor.setHeader(key, value);
		if (accessor instanceof NativeMessageHeaderAccessor) {
			NativeMessageHeaderAccessor nativeAccessor = (NativeMessageHeaderAccessor) accessor;
			ensureNativeHeadersAreMutable(nativeAccessor).setNativeHeader(key, value);
		}
		else {
			Object nativeHeaders = accessor
					.getHeader(NativeMessageHeaderAccessor.NATIVE_HEADERS);
			if (nativeHeaders == null) {
				nativeHeaders = new LinkedMultiValueMap<>();
				accessor.setHeader(NativeMessageHeaderAccessor.NATIVE_HEADERS,
						nativeHeaders);
			}
			if (nativeHeaders instanceof Map<?, ?>) {
				((Map) nativeHeaders).put(key, Collections.singletonList(value));
			}
		}
	}
```

We obtain a `nativeHeaders` `Map` from headers and modify it.

adriancole:
by abuse we assume it is usual for messages to be written by multiple threads and have eachother clobber eachother's map object :) Not saying it is wrong, just putting it another way.

artembilan:
> have each other clobber each other's map object 

Yes, this is really real behavior and it's OK to have it like that from the end-user perspective. We really can't control such a behavior from our side, when everything sits in the user code.
[...]

PR #1346, opened 2019-05-02
Making a copy of "nativeHeaders" and working on a copy

marcingrzejszczak:
otherwise we got ConcurrentModificationExceptions

fixes gh-1345

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/MessageHeaderPropagation.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/MessageHeaderPropagation.java
@@ -85,7 +85,9 @@ enum MessageHeaderPropagation
 						nativeHeaders = new LinkedMultiValueMap<>());
 			}
 			if (nativeHeaders instanceof Map<?, ?>) {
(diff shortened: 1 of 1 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/MessageHeaderPropagation.java
```diff
@@ -85,7 +85,9 @@ private void doPut(MessageHeaderAccessor accessor, String key, String value) {
 						nativeHeaders = new LinkedMultiValueMap<>());
 			}
 			if (nativeHeaders instanceof Map<?, ?>) {
-				((Map) nativeHeaders).put(key, Collections.singletonList(value));
+				Map<String, List<String>> copy = new LinkedMultiValueMap<>((Map<String, List<String>>) nativeHeaders);
```
artembilan: You need to have an `else` for that `if (nativeHeaders == null) {`.
Otherwise you are going to wrap a new `LinkedMultiValueMap()` into one more, almost for nothing.

marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-sleuth/commit/f54279457afdfc4702f268bb9a7440825d7b5d56

Issue #1327, opened 2019-04-03
Change RestTemplate in ZipkinRestTemplateSenderConfiguration


natrem:
I'm trying to send spans to a zipkin server secured via OAuth2.
For this purpose the provided extension `ZipkinRestTemplateCustomizer` does not fit.
Is it possible to declare the used `ZipkinRestTemplateWrapper` as a Bean with `ConditionalOnMissingBean` annotation?
I think that this would be the minimum update allowing me to override it simply.

I don't have any solution to merge quite simply the Oauth2 feature with the existing specific `ZipkinRestTemplateWrapper` so I think this support might wait (I don't use the URI resolution so I'm fine without it).

If you're ok with that proposition I'm willing to try it. Apart from the tests I did not look at it's a quite straightforward update in `ZipkinRestTemplateSenderConfiguration`.

marcingrzejszczak:
What exactly is missing currently in the `ZipkinRestTemplateCustomizer` ?

natrem:
With `ZipkinRestTemplateCustomizer` you can update the provided RestTemplate but ni my case I'd like to use an `OAuth2RestTemplate`. I don't think it's easy to plug-in with this extension mechanism.

natrem:
#1446 will allow this customization.

PR #1348, opened 2019-05-09
Support custom resttemplate

natrem:
as discussed in #1327 if I want to use a specific `RestTemplate` extension (e.g. a `OAuth2RestTempalte`) then I have to duplicate a lot of code from spring-cloud-sleuth.
This is due to many classes and Configuration classes being package private.

In this PR I add 2 things:
- the default `ZipkinRestTemplateWrapper` is now exposed as a bean with specific name (similar as Sender bean)
- since it is a bean it is subject to instrumentation by `TraceWebClientAutoConfiguration`; I added a utility class to filter the brave interceptors. This class is used in `ZipkinRestTemplateWrapper` to avoid tracing sending the spans and remains public for users wanting to replace the rest template.

I you feel like making this RestTemplate a bean is too risky then we should make several classes public to allow reuse of the default features in custom Senders.

Notes on the tests:
- I included an example of invalid extension to show the risk.
- I included an example of valid extension. this one might be dropped if you consider that the invalid test and the simple test showing the interceptors removal is enough
- these two tests are a bit slow (5s each)
- I did not manage to have `TraceWebClientAutoConfiguration` work properly in the unit test so I reproduced it's behavior in the test.

Notes on the documentation:
- I tried to update it but did not check the result at the time of this writing

### DIFF
--- a/docs/src/main/asciidoc/intro.adoc
+++ b/docs/src/main/asciidoc/intro.adoc
@@ -429,6 +429,18 @@ To do this you can use respectively `ZipkinAutoConfiguration.REPORTER_BEAN_NAME`
 include::../../../spring-cloud-sleuth-zipkin/src/test/java/org/springframework/cloud/sleuth/zipkin2/ZipkinAutoConfigurationTests.java[tags=override_default_beans,indent=0]
 ----
 
+The default web sender uses a specific `RestTemplate` that you can customize by providing a bean of type
+`ZipkinRestTemplateCustomizer`. If you wish to use a more specific RestTemplate (e.g. a OAUth2RestTemplate) 
+then you need to give your bean the specific name 
+`ZipkinAutoConfiguration.REST_TEMPLATE_BEAN_NAME`. 
+To avoid creating a 'loop of span reporting' your RestTemplate must reject the tracer interceptors. For this you can use the 
+utility class `SleuthWebClientInterceptorRemover`.
+
+[source,java]
+----
+include::../../../spring-cloud-sleuth-zipkin/src/test/java/org/springframework/cloud/sleuth/zipkin2/sender/ZipkinRestTemplateSenderConfigurationTest.java[tags=override_web_default_beans,indent=0]
+----
+
 == Additional Resources
 
 You can watch a video of https://twitter.com/reshmi9k[Reshmi Krishna] and https://twitter.com/mgrzejszczak[Marcin Grzejszczak] talking about Spring Cloud
--- a/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/sender/ZipkinRestTemplateSenderConfiguration.java
+++ b/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/sender/ZipkinRestTemplateSenderConfiguration.java
@@ -18,18 +18,21 @@ package org.springframework.cloud.sleuth.zipkin2.sender;
 
 import java.net.URI;
 import java.net.URISyntaxException;
+import java.util.List;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import zipkin2.reporter.Sender;
 
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.beans.factory.annotation.Qualifier;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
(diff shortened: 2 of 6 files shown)


spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/sender/ZipkinRestTemplateSenderConfiguration.java
```diff
@@ -34,30 +34,46 @@
...
 @ConditionalOnMissingBean(name = ZipkinAutoConfiguration.SENDER_BEAN_NAME)
 @Conditional(ZipkinSenderCondition.class)
 @EnableConfigurationProperties(ZipkinSenderProperties.class)
 class ZipkinRestTemplateSenderConfiguration {
 
+	/**
+	 * Zipkin rest template bean name. Name of the bean matters for supporting multiple tracing
+	 * systems.
+	 */
+	public static final String REST_TEMPLATE_BEAN_NAME = "zipkinRestTemplate";
```
devinsba: I think your intent is that this is public so you should move it to `ZipkinAutoConfiguration` with the other constants like this

docs/src/main/asciidoc/intro.adoc
```diff
@@ -429,6 +429,18 @@ To do this you can use respectively `ZipkinAutoConfiguration.REPORTER_BEAN_NAME`
 include::../../../spring-cloud-sleuth-zipkin/src/test/java/org/springframework/cloud/sleuth/zipkin2/ZipkinAutoConfigurationTests.java[tags=override_default_beans,indent=0]
 ----
 
+The default web sender uses a specific `RestTemplate` that you can customize by providing a bean of type
+`ZipkinRestTemplateCustomizer`. If you wish to use a more specific RestTemplate (e.g. a OAUth2RestTemplate) 
+then you need to give your bean the specific name 
+`ZipkinRestTemplateSenderConfiguration.REST_TEMPLATE_BEAN_NAME`.
```
devinsba: Update this when you move the property to the other class

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/sender/ZipkinRestTemplateSenderConfiguration.java
natrem: done

docs/src/main/asciidoc/intro.adoc
natrem: done

natrem:
While fixing the forgotten integration test (commit b4743959052254260723b057a67e5e02cf5ffb7d) I updated `AssertingRestTemplate` and moved the utility class to spring-cloud-sleuth-core module.

We could have an annotation `SleuthWebClientExcluded` to ask sleuth no to add the interceptors instead of rejecting them with the utility class but I don't think it simplifies the use case.

natrem:
Is there something I can help with to help you decide on the value of this PR for the project? Or did I miss something?

marcingrzejszczak:
Hi @natrem . Thank you for this PR. I didn't have time to look into it yet. I have some mixed feelings about `SleuthWebClientInterceptorRemover` and `SleuthWebClientExcluded`. I haven't however spent any time on suggesting a better approach.

natrem:
sorry if I seemed pushy. I understand your concern - that I share.
The main other possibility I considered was to simply make some classes public to ease creation of custom Sender `ZipkinAutoConfiguration.SENDER_BEAN_NAME`
Another possibility would be to create a class similar to ZipkinRestTemplateWrapper but using an `OAuth2RestTemplate` if there is a bean for the Oauth2 configuration (or is properties are filled)
Would you rather explore this possibility?

marcingrzejszczak:
> sorry if I seemed pushy. I understand your concern - that I share.

Don't worry, you weren't.

> Another possibility would be to create a class similar to ZipkinRestTemplateWrapper but using an OAuth2RestTemplate if there is a bean for the Oauth2 configuration (or is properties are filled)

That sounds much better IMO.

>  Would you rather explore this possibility?

Absolutely. Be my guest.

natrem:
Adaptation proposed in #1362

spencergibb:
Should this be closed then?

natrem:
I'd say that it depends on the discussion of new proposal, but most probably yes. We'll reopen if needed.

PR #1351, opened 2019-05-13
Added possibility to dynamically trace objects via aspectJ expression

tdanylchuk:
Purpose:
- ability to trace beans without applying annotations over class/method
- change tracing behavior without changing code
- trace beans from libs, where we cannot change code

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/dynamic/SleuthDynamicAspectJTracingInterceptor.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/dynamic/SleuthDynamicAspectJTracingInterceptor.java
@@ -1,0 +1,92 @@
+/*
+ * Copyright 2013-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.sleuth.dynamic;
+
+import brave.Span;
+import brave.Tracer;
+import org.aopalliance.intercept.MethodInvocation;
+
+import org.springframework.aop.IntroductionInterceptor;
+import org.springframework.aop.support.AopUtils;
+import org.springframework.cloud.sleuth.AbstractSleuthMethodInvocationProcessor;
+import org.springframework.cloud.sleuth.util.SpanNameUtil;
+
+/**
+ * Interceptor for tracing {@link MethodInvocation} objects.
+ *
+ * @author Taras Danylchuk
+ * @since 2.2.0
+ */
+public class SleuthDynamicAspectJTracingInterceptor extends
+		AbstractSleuthMethodInvocationProcessor implements IntroductionInterceptor {
+
+	private static final String SPAN_NAME_DELIMITER = "::";
+
+	private static final String METHOD_PARAMETER_PREFIX = "method_parameter_";
+
+	private final boolean isTraceParameters;
+
+	SleuthDynamicAspectJTracingInterceptor(boolean isTraceParameters) {
+		this.isTraceParameters = isTraceParameters;
+	}
+
+	@Override
+	public Object invoke(MethodInvocation invocation) throws Throwable {
+		Span span = startNewSpan(invocation);
+		try (Tracer.SpanInScope ws = tracer().withSpanInScope(span)) {
+			before(invocation, span);
+			return invocation.proceed();
+		}
+		catch (Exception ex) {
+			onFailure(span, ex);
+			throw ex;
+		}
+		finally {
+			after(span, true);
+		}
+	}
+
+	private Span startNewSpan(MethodInvocation joinPoint) {
+		Span span = tracer().nextSpan();
+		String methodName = joinPoint.getMethod().getName();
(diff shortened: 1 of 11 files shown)


tdanylchuk:
- I had idea to create aspect per expression, instead of having single expression for whole feature, but I don't know how to add dynamically add aspects to context

- I will add docs, if feature is ok and could be merged ;D

marcingrzejszczak:
Hi @tdanylchuk ! Could you please provide a little explanation of what exactly are you trying to achieve? Like, what problem are you trying to solve? Thanks!

marcingrzejszczak:
cc @adriancole @shakuzen

tdanylchuk:
I wrote ```Purpose:``` in ticket comment, to be more specific, I can add following case:
Example A: In my project I want to trace spring data related repositories with input parameters, I need to add following properties to my project:
```
spring.sleuth.dynamic.tracing.expression=target(org.springframework.data.repository.Repository+)
spring.sleuth.dynamic.tracing.traceParameters=true
```

Example B: I want to trace specific method of my own bean just for several days (code from test)
``
spring.sleuth.dynamic.tracing.expression=execution(* org.springframework.cloud.sleuth.dynamic.TestClass.testMethodA(..)) || execution(* org.springframework.cloud.sleuth.dynamic.TestClass.testMethodC(..))
```

And all calls to my spring data or other beans will be traced w\o changing the code, just adding these configs to properties file locally or remotely using spring cloud config.

Hope I was clear, let me know if I need to add more specific cases to elaborate.

File: spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/dynamic/SleuthDynamicAspectJTracingInterceptor.java [hunk 1]
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Interceptor for tracing {@link MethodInvocation} objects.
+ *
+ * @author Taras Danylchuk
+ * @since 2.2.0
+ */
+public class SleuthDynamicAspectJTracingInterceptor implements IntroductionInterceptor {
+
+	private static final String CLASS_NAME = "Class name";
```
adriancole: as tags are lookup keys and subject to indexing concerns, I think these sorts of names won't be good choices.

tdanylchuk: Class-name, method-name ?

tdanylchuk:
Well this feature is disabled by default, so only users which actually need it will use it.

Regarding bytebuddy, it's just different approach, and a lot of ppl does not like changing byte code of classes in any ways. 

The beauty of this approach is to handle next case: 
You have suspicious request traced on prod and trace does not contain full information regarding internal flows of single span, so property could be added without changing the code even temporary in order to resolve particular issue.

The implementation is very simple to support, since it is straightforward and contains just several lines of code.

WDYT?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/dynamic/SleuthDynamicAspectJTracingInterceptor.java
```diff
@@ -0,0 +1,97 @@
...
+			boolean isTraceParameters) {
+		this.tracer = tracer;
+		this.isTraceParameters = isTraceParameters;
+	}
+
+	@Override
+	public Object invoke(MethodInvocation invocation) throws Throwable {
+		ScopedSpan span = startNewSpan(invocation);
+		try {
+			return invocation.proceed();
```
robotmrv: this interceptor doesn't cover reactive case  
for e.g. see org.springframework.cloud.sleuth.annotation.ReactorSleuthMethodInvocationProcessor   
maybe it is better to make possible to reuse org.springframework.cloud.sleuth.annotation.SleuthMethodInvocationProcessor bean.   
```
Object process(MethodInvocation invocation, NewSpan newSpan,
		ContinueSpan continueSpan) throws Throwable;
```
Make it more generic and transfer metadata from NewSpan and ContinueSpan annotations to some object so other Interceptors could reuse it with their pointcuts

adriancole:
tag naming conventions are lowercase usually underscores where needed.
there are other tags in sleuth about class and method names so best to
reuse the same patterns.

adriancole:
I think you should be convincing other end users that we should be
maintaining this code. we dont add code for single users and you expect
this to be used by many. so maybe make the case and have people try it out
first? there ends up a lot to discuss in features of automatic tracing.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/dynamic/SleuthDynamicAspectJTracingInterceptor.java
tdanylchuk: I've reused existing logic from annotations feature
although it has only simple name instead of full one for class name
```invocation.getThis().getClass().getSimpleName()``` @marcingrzejszczak any reason for that?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/dynamic/SleuthDynamicAspectJTracingInterceptor.java
tdanylchuk: agree, refactored logic accordingly
currently i'm reusing existed logic from annotation feature.
As for reactive case i'm thinking to do in separate PR to keep commit and PR simpler. If it's not okay, I'll update current PR with reactive logic.

marcingrzejszczak:
My suggestion is that we wait until more people are interested in this feature. Currently there's one person. If we get more, we come back to this.

tdanylchuk:
We need to present this feature somehow, since you cannot be interested in thing you don't know :D
Although I respect this point.

marcingrzejszczak:
There wasn't much interest in the feature until now. I'm closing the PR.

PR #1362, opened 2019-05-28
Allow using Oauth2RestTemplate

natrem:
As discussed in #1327 I want to use OAuth2 to contact a Zipkin server.
After first version in #1348 and the discussion that followed I propose a second version that avoids exposing the used RestTemplate as a Bean.

Behavior is:
- by default the same as before
- if you provide a bean `OAuth2ProtectedResourceDetails` with name `ZipkinAutoConfiguration.OAUTH2_RESOURCE_BEAN_NAME` then use a new `ZipkinOAuth2RestTemplateWrapper` - similar to `ZipkinRestTemplateWrapper`
  New class will use this resource details to handle OAuth2 mechanism.

```java
		@Bean(name = ZipkinAutoConfiguration.OAUTH2_RESOURCE_BEAN_NAME)
		OAuth2ProtectedResourceDetails resourceDetails() {
			return new ClientCredentialsResourceDetails();
		}
```

This example comes from the test and is included in documentation.

This PR also contains:
- small refactoring in `ZipkinRestTemplateSenderConfiguration` to add new code and avoid duplication (or the zipkin URI resolution)
- add a `Condition`on `ActuatorSkipPatternProviderConfig`; this class triggered when adding `spring-security-oauth2-autoconfigure` dependency to zipkin module and caused errors
- exclude `ReactiveSecurityAutoConfiguration` in several _sleuth-zipkin_ tests (class triggered by adding `spring-security-oauth2-autoconfigure` dependency)

### DIFF
--- a/docs/src/main/asciidoc/intro.adoc
+++ b/docs/src/main/asciidoc/intro.adoc
@@ -429,6 +429,13 @@ To do this you can use respectively `ZipkinAutoConfiguration.REPORTER_BEAN_NAME`
 include::../../../spring-cloud-sleuth-zipkin/src/test/java/org/springframework/cloud/sleuth/zipkin2/ZipkinAutoConfigurationTests.java[tags=override_default_beans,indent=0]
 ----
 
+Default web `Sender` uses a `RestTemplate`. 
+If you want to use OAuth2 you need to provide a bean of type `OAuth2ProtectedResourceDetails` with name `ZipkinAutoConfiguration.OAUTH2_RESOURCE_BEAN_NAME` (usually a `ClientCredentialsResourceDetails`).
+
+----
+include::../../../spring-cloud-sleuth-zipkin/src/test/java/org/springframework/cloud/sleuth/zipkin2/sender/ZipkinWithOAuth2RestTemplateTests.java[tags=use_oauth2_resttemplate,indent=0]
+----
+
 == Additional Resources
 
 You can watch a video of https://twitter.com/reshmi9k[Reshmi Krishna] and https://twitter.com/mgrzejszczak[Marcin Grzejszczak] talking about Spring Cloud
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebAutoConfiguration.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebAutoConfiguration.java
@@ -106,7 +106,7 @@ public class TraceWebAutoConfiguration {
 
 	@Configuration
 	@ConditionalOnClass({ ServerProperties.class, EndpointsSupplier.class,
-			ExposableWebEndpoint.class })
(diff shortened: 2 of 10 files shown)


natrem:
I left the PR as draft in order to validate its use in real life sample application.

When using the updated module I have issues in conditions evaluation.
When I specify an `OAuth2ProtectedResourceDetails` (same as unit test) the context wouldn't load due to the two configuration classes `OAuth2RestTemplateSenderConfiguration` and `DefaultRestTemplateSenderConfiguration` create the `Sender` bean.

I know that conditions evaluation can be tricky so I suspect I missed something.
My current idea (in progress) is to create my own `Condition` class but it might be overkill. 

Any ideas on this?

natrem:
code coverage reports misses on classes I did not change. Can you confirm it's a glitch or point me to possible changes?

devinsba:
I assume it has something to do with the excluded autoconfigurations though I have no idea why

natrem:
If you're ok with it then the condition is not too intrusive and gives more clarity since there are 2 conditions aggregated

natrem:
I agree on your first points and understand your concern regarding maintenance. IMO securing Zipkin instance with OAuth2 is a valuable feature hence the discussion and effort.

A first approach would be to have the RestTemplate used to report the spans a Bean that users can override. The downside is that beans are by default traced by sleuth therefore creating a loop unless the RestTemplate explicitly rejects some `ClientHttpRequestInterceptor` (actually proposed as first solution in #1348).

In order to reduce the changes and improve reuse of existing features another possibility would be to make several classes currently defined as package private as public. This would allow users to create a custom `Sender` that would use a custom `RestTemplate` (in my case `OAuth2RestTemplate`) and still benefit from the existing features. By looking quickly at the diff I think of `ZipkinUrlExtractor`, `ZipkinUriResolver` (class I proposed to contain the behavior of method `ZipkinRestTemplateWrapper.resolvedZipkinUri`. 
The downside here is to expose more model as public. They may need to be moved to another `Configuration` classes to ease ordering of autoconfiguration.

marcingrzejszczak:
I wonder if this shouldn't be first done as a separate project. What I mean is that until now, no one has requested this feature. Maybe it's kind of specific to your company. If you create a library that does all of the above, and users get interested in it, we could consider adding support for all of that both in Zipkin and Sleuth. WDYT @natrem @adriancole

natrem:
I'm ok with that. The initial purpose of the PR was 'I think it can be a useful feature' so a separate project could address this.
In order to make this project's code simpler I'd like to reuse some classes currently not public. 
For instance the interface `ZipkinUrlExtractor` (maybe a bit more; I don't remember all the ones I saw). Would that be ok for you?

marcingrzejszczak:
With this getting merged https://github.com/spring-cloud/spring-cloud-sleuth/pull/1446 you'll be able to customize the RestTemplate instance to basically do whatever you want.

natrem:
fine with me. thanks

PR #1375, opened 2019-06-18
Decorate with Traceable Scheduled Executor service

fvlad:
Hi. I have a ScheduledExecutorService bean in context which is decorated with TraceableExecutorService. The problem is that the Scheduled part became missing and I could no longer inject this bean as ScheduledExecutorService, only as ExecutorService. I propose to adjust the decoration mechanism to check if it is scheduled or not.


marcingrzejszczak:
Hi @fvlad ! Can you create a test for that? Also can you check if it hasn't already been instrumented? We don't want to instrument `TraceableExecutorService` with another `TraceableExecutorService` instance.

fvlad:
Hi @marcingrzejszczak. Added test that checks the bean is decorated with Scheduled executor service. Also added tests that post processing tracing bean returns the same bean and changed the logic accordingly.

marcingrzejszczak:
Congrats @fvlad on your first contribution! :)

Issue #1336, opened 2019-04-16
ExecutorBeanPostProcessor creates proxies with null fields values


novy1234:
Hi,
I found a problem during creation of async executors proxies in ExecutorBeanPostProcessor. 
Namely when creating proxies for the executors they are created with null fields inside, which is a bit problematic if you use any delegating executor, e.g. DelegatingSecurityContextSchedulingTaskExecutor. 

From what I have seen, CGLib uses no-args constructor if no constructor arguments have been passed, hence the fields have null values

I use Greenwich.SR1 train/release

Can you help me with that?

marcingrzejszczak:
All CGLIB proxies have null fields. You shouldn't be using those. What is exactly your problem?

novy1234:
@marcingrzejszczak My problem is that ExecutorBeanPostProcessor creates proxy of DelegatingSecurityContextSchedulingTaskExecutor which during execution forwards call to yet another task executor. It cannot do so, because the delegate field is null

marcingrzejszczak:
ok, can you create a small sample that replicates this issue? ALso can you check the latest snapshots, cause AFAIR I've added some fixes at some point

roll57:
Hello,
I have the same issue, I notice the following log when I start my server:

> o.s.aop.framework.CglibAopProxy          : Unable to proxy interface-implementing method [public final java.util.List org.springframework.security.concurrent.DelegatingSecurityContextExecutorService.invokeAll(java.util.Collection,long,java.util.concurrent.TimeUnit) throws java.lang.InterruptedException] because it is marked as final: Consider using interface-based JDK proxies instead!

In my code I do:
`executorService.invokeAll(...)`

I tried with an executorService which have an _invokeAll_ method that is not final and it works.

marcingrzejszczak:
We can't wrap final methods with cglib. Just try no to use them if possible. My request for a sample is still active

kilink:
[DelegatingSecurityContextAsyncTaskExecutor](https://github.com/spring-projects/spring-security/blob/master/core/src/main/java/org/springframework/security/task/DelegatingSecurityContextAsyncTaskExecutor.java) has such final methods; registering one as a bean and and then trying to use it anywhere in my app, I get the same issue.

```java
@Configuration
public class AppConfig {
    @Bean
    public Executor executor(TaskExecutorBuilder: builder) {
        return new DelegatingAsyncTaskExecutor(builder.build());
    }
}
```

[...]

PR #1385, opened 2019-07-01
Fix proxying of executors with final methods

kilink:
When proxying the various Executor interfaces in ExecutorbeanPostProcessor, the decision of whether or not to use a cglib proxy was determined by checking if the bean's class is final, but final methods would also cause issues when using cglib. A common example of such an implementation is DelegatingSecurityContextScheduledExecutorService, which has final methods. (see #1336).

To resolve the issue with ThreadPoolTaskExecutor, AsyncTaskExecutor, and ExecutorService, fall back to the JDK proxy if the bean has any final methods.


marcingrzejszczak:
Hey @kilink , thanks for this PR. Looks really good! I'll backport it to 2.1.x

Issue #1382, opened 2019-06-25
WebClientTracerSubscriber.terminateSpan() throws IllegalArgumentException for non-standard status code


blake-bauman:
When using Spring Cloud Sleuth with a WebClient call, and the downstream service responds with a non-standard HTTP code (say 499), then WebClientTracerSubscriber.terminateSpan() throws an IllegalArgumentException because the status code cannot be mapped to one of the HttpStatus enum values.

The offending code is here:

```java
    void terminateSpan(@Nullable ClientResponse clientResponse,
                       @Nullable Throwable throwable) {
        if (clientResponse == null || clientResponse.statusCode() == null) {
```

From the Javadoc, ClientResponse.statusCode() will never return null, but will instead throw an IllegalArgumentException.  As suggested by https://github.com/spring-projects/spring-framework/issues/21289 clientResponse.getRawStatusCode() should be used if there's a possibility of a non-standard response code.  Our app will call something like:

```java
        webClient.get()
                 .uri(...)
                 .exchange()
                 .flatMap(response -> {
                     if (response.getRawStatusCode() == ...)
                         ...
                 })
```

However, this response handler never gets control because of the exception thrown when `terminateSpan()` calls `clientResponse.statusCode()`.  When we remove spring-cloud-sleuth, it works fine, but since we would like the OpenTracing support, we still want to use sleuth.

blake-bauman:
We are using the Greenwich.SR1 release train.

marcingrzejszczak:
Hi @blake-bauman ! Are you interested in filing a PR with a failing test and the impl that fixes it?

blake-bauman:
I wish, but at the moment, the Legal department at my company hasn't approved me for public open source code contributions.  I'm going through the vetting process, but that can take months.  I'm working on a simple project that demonstrates this.

Browsing through the code, however, it does appear that the problem is localized to this null check, the `error` validation, and the reason phrase in the exception.  walking through handleReceive(), that part appears to be ok.

adriancole:
ok so the fix for who can do this would be to try/catch and coerce IAE to
null here, with a mock test with some invalid status from the server.. who
knows, maybe 0?

[...]

PR #1387, opened 2019-07-03
Return 0 status code on exception or invalid code

marcingrzejszczak:
fixes gh-1382


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java
```diff
@@ -395,7 +395,17 @@ public String requestHeader(ClientRequest request, String name) {
 
 		@Override
 		public Integer statusCode(ClientResponse response) {
-			return response.statusCode().value();
+			return response.rawStatusCode();
```
adriancole: something like this otherwise you still end up with the exception :P

```
  @Override public Integer statusCode(HttpResponse response) {
    int result = statusCodeAsInt(response);
    return result != 0 ? result : null;
  }
```

marcingrzejszczak: Ok but `statusCode` is called only via the `statusCodeAsInt` in Brave

PR #1401, opened 2019-07-19
Add null check for sampler probability property

elefeint:
When `spring.sleuth.sampler.probability` property is missing, instantiating `ProbabilityBasedSampler` throws an NPE during an attempt to autobox a null probability.

This PR adds an assertion with a user-readable error message.


marcingrzejszczak:
Great catch @elefeint! Thanks!

Issue #1409, opened 2019-07-26
Should RateLimitingSampler be public?


elefeint:
I am trying to replicate* changes from [`SamplerAutoConfiguration`](https://github.com/spring-cloud/spring-cloud-sleuth/blob/e13f7f80489da35fbac1840451acd4252219e652/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/SamplerAutoConfiguration.java#L39) autoconfiguration logic into [spring-cloud-gcp](https://github.com/spring-cloud/spring-cloud-gcp/blob/4059007bd5fbe7f7c7c00d8472b00230a52465fb/spring-cloud-gcp-autoconfigure/src/main/java/org/springframework/cloud/gcp/autoconfigure/trace/StackdriverTraceAutoConfiguration.java#L199-L224), but [`RateLimitingSampler`](https://github.com/spring-cloud/spring-cloud-sleuth/blob/e13f7f80489da35fbac1840451acd4252219e652/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/RateLimitingSampler.java#L30) is package-private. Was it meant to be only package private, or is it okay to relax its visibility to `public`? `ProbabilityBasedSampler`, which is in the same package and is used similarly, is public already.

\* I'd rather not replicate them at all and just rely on `SamplerAutoConfiguration`, but whenever `TraceAutoConfiguration` and `SamplerAutoConfiguration` are both present, `TraceAutoConfiguration`'s [NEVER_SAMPLE](https://github.com/spring-cloud/spring-cloud-sleuth/blob/a5c4f05c2d6c15fe06b7e6ec3325162bc4b58796/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java#L122) sampler gets instantiated due to Spring Boot ordering (`TraceAutoConfiguration` has an order and `SamplerAutoConfiguration` does not, so `SamplerAutoConfiguration` is processed at the end). From the docs, NEVER_SAMPLE [seems like intended behavior](https://cloud.spring.io/spring-cloud-static/spring-cloud-sleuth/2.0.4.RELEASE/multi/multi__sampling.html#_sampling_in_spring_cloud_sleuth).

TYsewyn:
[...]

PR #1410, opened 2019-07-26
Enable sampling via properties configuration

TYsewyn:
Fixes gh-1409


marcingrzejszczak:
Fixed via https://github.com/spring-cloud/spring-cloud-gcp/pull/1774

Stmated:
So this will not be fixed in the actual spring-cloud-sleuth-core? That the recommended solution is to do as spring-cloud-gcp has done?

marcingrzejszczak:
There's nothing really to fix here so yeah, please follow what spring-cloud-gcp has done.

Issue #1411, opened 2019-07-27
Call reporter.check() during post-construct


adriancole:
AsyncReporter is by definition async, so it can lead to late errors, if it isn't checked prior to use. While the checks themselves may not be perfect, calling them is a good way to reduce time lost. One implementation is to call `asyncReporter.check()` or `sender.check()` during post-construct and checking if it is error or not. The meta-discussion is here.

https://github.com/openzipkin/zipkin-reporter-java/issues/154

Also, it seems not many people are using metrics. It may make sense that if there are no micrometer reporters configured, to pipe zipkin's reporter metrics into logs. This is making an assumption that high load sites will be using metrics, so when proper metrics are configured this eliminates the spam the console or disk risk.

PR #1413, opened 2019-07-30
Calling asyncreporter#check upon bean registration

marcingrzejszczak:
fixes gh-1411

### DIFF
--- a/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/ZipkinAutoConfiguration.java
+++ b/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/ZipkinAutoConfiguration.java
@@ -100,10 +100,19 @@ public class ZipkinAutoConfiguration {
 				.messageTimeout(zipkin.getMessageTimeout(), TimeUnit.SECONDS)
 				.metrics(reporterMetrics).build(zipkin.getEncoder());
 		CheckResult checkResult = checkResult(asyncReporter);
-		if (log.isDebugEnabled()) {
-			log.debug("Check result of the async reporter is [" + checkResult + "]");
-		}
+		logCheckResult(asyncReporter, checkResult);
 		return asyncReporter;
+	}
+
+	private void logCheckResult(AsyncReporter asyncReporter, CheckResult checkResult) {
+		if (log.isDebugEnabled() && checkResult != null && checkResult.ok()) {
+			log.debug("Check result of the [" + asyncReporter.toString() + "] is ["
+					+ checkResult + "]");
+		}
+		else if (checkResult != null && !checkResult.ok()) {
+			log.warn("Check result of the [" + asyncReporter.toString()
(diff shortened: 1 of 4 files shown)


spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/ZipkinAutoConfiguration.java
```diff
@@ -87,9 +95,32 @@
...
 			ZipkinProperties zipkin, @Qualifier(SENDER_BEAN_NAME) Sender sender) {
 		// historical constraint. Note: AsyncReporter supports memory bounds
-		return AsyncReporter.builder(sender).queuedMaxSpans(1000)
+		AsyncReporter<Span> asyncReporter = AsyncReporter.builder(sender)
+				.queuedMaxSpans(1000)
 				.messageTimeout(zipkin.getMessageTimeout(), TimeUnit.SECONDS)
 				.metrics(reporterMetrics).build(zipkin.getEncoder());
+		CheckResult checkResult = checkResult(asyncReporter);
+		if (log.isDebugEnabled()) {
+			log.debug("Check result of the async reporter is [" + checkResult + "]");
```
adriancole: consider adding the toString of reporter instead of "async reporter"

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/ZipkinAutoConfiguration.java
```diff
@@ -87,9 +95,32 @@
...
+	}
+
+	private CheckResult checkResult(AsyncReporter<Span> asyncReporter) {
+		ExecutorService executor = Executors.newSingleThreadExecutor();
+		Callable<CheckResult> task = asyncReporter::check;
+		Future<CheckResult> future = executor.submit(task);
+		try {
+			return future.get(1, TimeUnit.SECONDS);
+		} catch (Exception ex) {
+			if (log.isDebugEnabled()) {
```
adriancole: if this is only at startup, I think people talking about this will want it to be something higher than debug, like warn

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/sender/ZipkinRestTemplateSenderConfiguration.java
```diff
@@ -145,6 +147,8 @@ ZipkinLoadBalancer noOpLoadBalancer(final ZipkinProperties zipkinProperties) {
 
 	private static final Log log = LogFactory.getLog(ZipkinRestTemplateWrapper.class);
 
+	private static final int DEFAULT_TIMEOUT = 500;
```
shakuzen: Is this different than before? I wonder if users in slow networks will start to get timeouts because of this.

Issue #1421, opened 2019-08-14
Add redaction support to extra fields config


adriancole:
Some people want to add "extra fields" but at process scope, and not propagated downstream. Currently, the approach to implementing this is to add them as "redacted" https://github.com/openzipkin/brave/tree/master/brave#redacting-fields

This serves use cases like log context synchronization for totally trusted applications. What I mean by trusted is that applications should be as careful about what to put in extra fields as they are about MDC, as anything in the classpath can read them.

Incidentally, this was added for sleuth, requested by @alonbl but I don't think we hooked it together, yet. hence the issue.

https://github.com/openzipkin/brave/tree/master/brave#redacting-fields
https://github.com/openzipkin/brave/pull/931

adriancole:
You can now use the `spring.sleuth.local-keys` property to pass a list keys that will be propagated locally but will not be propagated over the wire.

alonbl:
Thank you all! Much appreciated.

PR #1424, opened 2019-08-16
Adds "extra fields" but at process scope

marcingrzejszczak:
fixes gh-1421

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-sleuth.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-sleuth.adoc
@@ -448,8 +448,8 @@ String countryCode = ExtraFieldPropagation.get(span.context(), "x-country-code")
 IMPORTANT: A difference from previous versions of Sleuth is that, with Brave, you must pass the list of baggage keys.
 There are the following properties to achieve this.
 With the `spring.sleuth.baggage-keys`, you set keys that get prefixed with `baggage-` for HTTP calls and `baggage_` for messaging.
-You can also use the `spring.sleuth.propagation-keys` property to pass a list of prefixed keys that are propagated without any prefix.
-You can also use the `spring.sleuth.redacted-keys` property to pass a list keys that will be available locally but will not be propagated over the wire.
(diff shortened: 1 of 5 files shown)


docs/src/main/asciidoc/spring-cloud-sleuth.adoc
```diff
@@ -446,9 +446,10 @@ String countryCode = ExtraFieldPropagation.get(span.context(), "x-country-code")
 ```
 
 IMPORTANT: A difference from previous versions of Sleuth is that, with Brave, you must pass the list of baggage keys.
-There are two properties to achieve this.
+There are the following properties to achieve this.
 With the `spring.sleuth.baggage-keys`, you set keys that get prefixed with `baggage-` for HTTP calls and `baggage_` for messaging.
-You can also use the `spring.sleuth.propagation-keys` property to pass a list of prefixed keys that are whitelisted without any prefix.
+You can also use the `spring.sleuth.propagation-keys` property to pass a list of prefixed keys that are propagated without any prefix.
+You can also use the `spring.sleuth.redacted-keys` property to pass a list keys that will be available locally but will not be propagated over the wire.
```
adriancole: should we call this local-keys? cc @shakuzen @devinsba @jeqo

adriancole: if we call this local-keys, we should clarify "will be available locally" as "will be propagated locally"

shakuzen: Yea, I would prefer terminology other than redacted, since it is but one use case and implies perhaps more than this is (e.g. making something available in logging that's "redacted" seems off to me). `local-keys` as suggested sounds good to me.
While we're polishing the documentation here, perhaps the description for `propagation-keys` could be updated to something like "keys that are propagated to remote services without any prefix".

adriancole: we can also change brave with deprecation (we'd never be able to remove that method before a major, but at least it could clarify)

marcingrzejszczak:
Suggestions make sense and got applied.

PR #1428, opened 2019-08-21
Added support for quartz instrumentation

ammmze:
Added instrumentation support for quartz jobs. This allows jobs to be wrapped in a tracing span so requests made from a quartz job can be correlated.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TraceQuartzAutoConfiguration.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TraceQuartzAutoConfiguration.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2013-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.sleuth.instrument.quartz;
+
+import brave.Tracing;
+import org.quartz.Scheduler;
+
+import org.springframework.beans.factory.InitializingBean;
+import org.springframework.boot.autoconfigure.AutoConfigureAfter;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.autoconfigure.quartz.QuartzAutoConfiguration;
+import org.springframework.cloud.sleuth.autoconfig.TraceAutoConfiguration;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * {@link org.springframework.boot.autoconfigure.EnableAutoConfiguration
+ * Auto-configuration} enables Quartz span information propagation.
+ *
+ * @author Branden Cash
+ * @since 2.2.0
+ */
+@Configuration
+@ConditionalOnBean({ Tracing.class, Scheduler.class })
+@AutoConfigureAfter({ TraceAutoConfiguration.class, QuartzAutoConfiguration.class })
+@ConditionalOnProperty(value = "spring.sleuth.quartz.enabled", matchIfMissing = true)
+public class TraceQuartzAutoConfiguration implements InitializingBean {
+
+	private Scheduler scheduler;
+
+	private Tracing tracing;
+
+	public TraceQuartzAutoConfiguration(Scheduler scheduler, Tracing tracing) {
+		this.scheduler = scheduler;
+		this.tracing = tracing;
+	}
+
+	@Bean
+	public TracingJobListener tracingJobListener() {
+		return new TracingJobListener(tracing);
+	}
+
+	@Override
+	public void afterPropertiesSet() throws Exception {
+		scheduler.getListenerManager().addTriggerListener(tracingJobListener());
+		scheduler.getListenerManager().addJobListener(tracingJobListener());
+	}
+
+}
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
@@ -1,0 +1,120 @@
+/*
+ * Copyright 2013-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.sleuth.instrument.quartz;
+
+import brave.Span;
+import brave.Tracer.SpanInScope;
+import brave.Tracing;
+import brave.propagation.Propagation.Getter;
+import brave.propagation.TraceContextOrSamplingFlags;
+import org.quartz.JobDataMap;
+import org.quartz.JobExecutionContext;
+import org.quartz.JobExecutionException;
+import org.quartz.JobListener;
+import org.quartz.Trigger;
+import org.quartz.Trigger.CompletedExecutionInstruction;
+import org.quartz.TriggerListener;
+
+/**
+ * {@link org.quartz.JobListener JobListener} that will wrap a span around quartz jobs
+ * when they start and finish.
+ *
+ * @author Branden Cash
+ * @since 2.2.0
+ */
+class TracingJobListener implements JobListener, TriggerListener {
+
+	static final String TRIGGER_TAG_KEY = "quartz.trigger";
+
+	static final String CONTEXT_SPAN_KEY = Span.class.getName();
+
+	static final String CONTEXT_SPAN_IN_SCOPE_KEY = SpanInScope.class.getName();
+
+	private static final Getter<JobDataMap, String> GETTER = (carrier, key) -> {
+		Object value = carrier.get(key);
+		if (value instanceof String) {
+			return (String) value;
+		}
+		return null;
+	};
+
+	private final Tracing tracing;
+
+	TracingJobListener(Tracing tracing) {
+		this.tracing = tracing;
+	}
+
+	@Override
+	public String getName() {
+		return getClass().getName();
+	}
+
+	@Override
+	public void triggerFired(Trigger trigger, JobExecutionContext context) {
+		TraceContextOrSamplingFlags extracted = tracing.propagation().extractor(GETTER)
+				.extract(context.getMergedJobDataMap());
+		Span span = tracing.tracer().nextSpan(extracted)
+				.name(context.getTrigger().getJobKey().toString())
+				.tag(TRIGGER_TAG_KEY, context.getTrigger().getKey().toString());
+		context.put(CONTEXT_SPAN_KEY, span);
+		context.put(CONTEXT_SPAN_IN_SCOPE_KEY,
+				tracing.tracer().withSpanInScope(span.start()));
+	}
+
+	@Override
+	public boolean vetoJobExecution(Trigger trigger, JobExecutionContext context) {
+		return false;
+	}
+
+	@Override
+	public void triggerMisfired(Trigger trigger) {
+
+	}
+
+	@Override
+	public void triggerComplete(Trigger trigger, JobExecutionContext context,
+			CompletedExecutionInstruction triggerInstructionCode) {
+		closeTrace(context);
+	}
+
+	@Override
+	public void jobToBeExecuted(JobExecutionContext context) {
+
+	}
+
+	@Override
+	public void jobExecutionVetoed(JobExecutionContext context) {
+		closeTrace(context);
+	}
+
+	@Override
+	public void jobWasExecuted(JobExecutionContext context,
+			JobExecutionException jobException) {
+	}
+
+	private void closeTrace(JobExecutionContext context) {
+		Object spanInScope = context.get(CONTEXT_SPAN_IN_SCOPE_KEY);
+		Object span = context.get(CONTEXT_SPAN_KEY);
+		if (spanInScope instanceof SpanInScope) {
+			((SpanInScope) spanInScope).close();
+		}
+		if (span instanceof Span) {
+			((Span) span).finish();
+		}
+	}
+
+}
--- a/spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/quartz/TraceQuartzAutoConfigurationTest.java
+++ b/spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/quartz/TraceQuartzAutoConfigurationTest.java
@@ -1,0 +1,139 @@
+/*
+ * Copyright 2013-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.sleuth.instrument.quartz;
+
+import brave.Tracer;
+import brave.Tracing;
+import org.junit.Test;
+import org.quartz.ListenerManager;
+import org.quartz.Scheduler;
+
+import org.springframework.boot.autoconfigure.AutoConfigurations;
+import org.springframework.boot.autoconfigure.AutoConfigureBefore;
+import org.springframework.boot.test.context.runner.ApplicationContextRunner;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
+import static org.mockito.Mockito.when;
+
+/**
+ * @author Branden Cash
+ */
+public class TraceQuartzAutoConfigurationTest {
+
+	private final ApplicationContextRunner contextRunner = new ApplicationContextRunner()
+			.withConfiguration(AutoConfigurations.of(SchedulerConfig.class,
+					TracingConfig.class, TraceQuartzAutoConfiguration.class));
+
+	@Test
+	public void should_create_job_listener_bean_when_all_conditions_are_met() {
+		// when
+		this.contextRunner.run(context -> {
+			// expect
+			assertThat(context).hasSingleBean(TracingJobListener.class);
+		});
+	}
+
+	@Test
+	public void should_add_listener_to_trigger_listeners_when_conditions_are_met() {
+		// when
+		this.contextRunner.run(context -> {
+			// expect
+			verify(context.getBean(Scheduler.class).getListenerManager())
+					.addTriggerListener(context.getBean(TracingJobListener.class));
+		});
+	}
+
+	@Test
+	public void should_add_listener_job_listeners_when_conditions_are_met() {
+		// when
+		this.contextRunner.run(context -> {
+			// expect
+			verify(context.getBean(Scheduler.class).getListenerManager())
+					.addJobListener(context.getBean(TracingJobListener.class));
+		});
+	}
+
+	@Test
+	public void should_not_create_listener_bean_when_tracing_bean_is_not_present() {
+		new ApplicationContextRunner().withConfiguration(AutoConfigurations
+				// given
+				.of(SchedulerConfig.class, TraceQuartzAutoConfiguration.class))
+
+				// when
+				.run(context -> {
+					// expect
+					assertThat(context).doesNotHaveBean(TracingJobListener.class);
+				});
+	}
+
+	@Test
+	public void should_not_create_listener_when_scheduler_bean_is_not_present() {
+		new ApplicationContextRunner().withConfiguration(AutoConfigurations
+				// given
+				.of(TracingConfig.class, TraceQuartzAutoConfiguration.class))
+
+				// when
+				.run(context -> {
+					// expect
+					assertThat(context).doesNotHaveBean(TracingJobListener.class);
+				});
+	}
+
+	@Test
+	public void should_not_create_listener_when_sleuth_is_disabled() {
+		// when
+		this.contextRunner.withPropertyValues("spring.sleuth.quartz.enabled=false")
+				// expect
+				.run(context -> {
+					assertThat(context).doesNotHaveBean(TracingJobListener.class);
+				});
+	}
+
+	@Configuration
+	@AutoConfigureBefore(TraceQuartzAutoConfiguration.class)
+	public static class SchedulerConfig {
+
+		@Bean
+		public Scheduler scheduler() throws Exception {
+			Scheduler scheduler = mock(Scheduler.class);
+			when(scheduler.getListenerManager()).thenReturn(mock(ListenerManager.class));
+			return scheduler;
+		}
+
+	}
+
+	@Configuration
+	@AutoConfigureBefore(TraceQuartzAutoConfiguration.class)
+	public static class TracingConfig {
+
+		@Bean
+		public Tracing tracing() {
+			return mock(Tracing.class);
+		}
+
+		@Bean
+		public Tracer tracer() {
+			return mock(Tracer.class);
+		}
+
+	}
+
+}
--- a/spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListenerTest.java
+++ b/spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListenerTest.java
@@ -1,0 +1,355 @@
+/*
+ * Copyright 2013-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.sleuth.instrument.quartz;
+
+import java.util.ArrayDeque;
+import java.util.HashMap;
+import java.util.Properties;
+import java.util.Queue;
+import java.util.concurrent.CompletableFuture;
+
+import brave.Tracer.SpanInScope;
+import brave.Tracing;
+import brave.propagation.Propagation.Setter;
+import brave.propagation.StrictScopeDecorator;
+import brave.propagation.ThreadLocalCurrentTraceContext;
+import org.junit.After;
+import org.junit.Before;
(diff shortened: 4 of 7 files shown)


pivotal-issuemaster:
@ammmze Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=1428)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=1428) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ammmze Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=1428)!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
```diff
@@ -0,0 +1,76 @@
...
+		return getClass().getName();
+	}
+
+	@Override
+	public void jobToBeExecuted(JobExecutionContext context) {
+		String name = String.format("Job: %s; Trigger: %s",
+				context.getTrigger().getJobKey(), context.getTrigger().getKey());
+		Span span = tracer.nextSpan().name(name);
+		jobSpans.put(context.getFireInstanceId(), span);
+		tracer.withSpanInScope(span.start());
```
devinsba: You should attach the scope object to the context in order to be able to close it

From the javadocs:
```
you must call close to end the scope, or risk leaking resources associated with the scope.
```

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
```diff
@@ -0,0 +1,76 @@
...
+	public String getName() {
+		return getClass().getName();
+	}
+
+	@Override
+	public void jobToBeExecuted(JobExecutionContext context) {
+		String name = String.format("Job: %s; Trigger: %s",
+				context.getTrigger().getJobKey(), context.getTrigger().getKey());
+		Span span = tracer.nextSpan().name(name);
+		jobSpans.put(context.getFireInstanceId(), span);
```
devinsba: Instead of using a map in this class I would use the `put` method on the `JobExecutionContext` to attach the objects you need later to the quartz context

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
```diff
@@ -0,0 +1,76 @@
...
+/**
+ * {@link org.quartz.JobListener JobListener} that will wrap a span around quartz jobs
+ * when they start and finish.
+ *
+ * @author Branden Cash
+ */
+@Component
+public class TracingJobListener implements JobListener {
+
+	private final Map<String, Span> jobSpans = new HashMap<>();
```
devinsba: If keeping the map you will need to use the Concurrent variant to ensure thread safety

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
adriancole: agree. map is dodgiest option, can accidentally leak etc. use as last resort

ammmze: Very good suggestion :D thank you

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
ammmze: Good suggestion, but I like putting it in the `JobExecutionContext` way better

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
ammmze: Ah...yes...I overlooked the try with resources in the examples I was looking at [here](https://cloud.spring.io/spring-cloud-sleuth/reference/html/#_span_lifecycle). Thank you

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListenerTest.java
```diff
@@ -0,0 +1,165 @@
...
+	private JobExecutionContext context;
+
+	@Mock
+	private JobExecutionException jobException;
+
+	@Mock
+	private Span span;
+
+	@Mock
+	private Closeable spanInScope;
```
ammmze: This is less than ideal ... the `SpanInScope` can't be mocked because its a final class and the constructor is package private. So I went up to the interface I care about, but unfortunately had to the same in `TracingJobListener`. I really don't want to have to shape the code around what is possible in tests. Got any suggestions? Hopefully without wiring up the whole application context.

ammmze: It looks like I may be able to get an actual Tracing/Tracer using something like this:

```
Tracing tracing = Tracing.newBuilder()
		.currentTraceContext(ThreadLocalCurrentTraceContext.newBuilder()
				.addScopeDecorator(StrictScopeDecorator.create()).build())
		.sampler(Sampler.ALWAYS_SAMPLE).spanReporter(this.reporter).build();

Tracer tracer = this.tracing.tracer();
```

However, i'm still not sure how I'd be able to verify that the `SpanInScope` ever has `close` invoked.

ammmze: Got it figured out by creating a real Tracing object and injecting a mock CurrentTraceContext

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
ammmze: Code has been updated. Please review and resolve the conversation if the concern is resolved.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
ammmze: Code has been updated. Please review and resolve the conversation if the concern is resolved.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
```diff
@@ -0,0 +1,84 @@
...
+	}
+
+	@Override
+	public String getName() {
+		return getClass().getName();
+	}
+
+	@Override
+	public void jobToBeExecuted(JobExecutionContext context) {
+		String name = String.format("Job: %s; Trigger: %s",
```
adriancole: examples of what these names would be? I am worried about unlimited cardinality which would blow up people's span name drop-down

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListenerTest.java
```diff
@@ -0,0 +1,190 @@
...
+import static org.springframework.cloud.sleuth.instrument.quartz.TracingJobListener.CONTEXT_SPAN_IN_SCOPE_KEY;
+import static org.springframework.cloud.sleuth.instrument.quartz.TracingJobListener.CONTEXT_SPAN_KEY;
+
+/**
+ * @author Branden Cash
+ */
+@RunWith(MockitoJUnitRunner.class)
+public class TracingJobListenerTest {
+
+	private TracingJobListener listener;
```
adriancole: please just test normally and not with mocks (at least the tracer part). StrictScopeDecorator will verify scopes are closed. This will let us see things like span names without a lot of ceremony

```
Tracing tracing = Tracing.newBuilder()
      .spanReporter(spans::add)
      .currentTraceContext(ThreadLocalCurrentTraceContext.newBuilder()
        .addScopeDecorator(StrictScopeDecorator.create())
        .build())
```

how you define spans depends on whether this uses threads or not. If not, then `List<Span>`. Otherwise a blocking queue..

```java
  /** See brave.http.ITHttp for rationale on using a concurrent blocking queue */
  BlockingQueue<Span> spans = new LinkedBlockingQueue<>();


  /** Call this to block until a span was reported */
  Span takeSpan() throws InterruptedException {
    Span result = spans.poll(3, TimeUnit.SECONDS);
    assertThat(result)
      .withFailMessage("Span was not reported")
      .isNotNull();
    return result;
  }
```

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
ammmze: Gotcha ... I wasn't really sure where this was used. Where I work we haven't got a UI around sleuth yet. Currently we just use it to correlate log messages in sumo logic. But some examples could be:

* `Job: DEFAULT.memberIndexingJob; Trigger: DEFAULT.MT_4b7k81ajfts4g`
* `Job: DEFAULT.memberIndexingJob; Trigger: DEFAULT.memberIndexingScheduledTrigger`
* `Job: INDEXING.memberIndexingJob; Trigger: INDEXING.memberIndexingScheduledTrigger`

Maybe just job key alone would be sufficient? I just don't have real world experience where this is used to make that call.

ammmze: Oh...looks like you can add tags to the span? I'm not sure which would be more useful where, but maybe the job key as the name and tag it with the trigger key? Or vice versa?

adriancole: job could possibly be a decent choice. if that also could be unlimited, we sometimes look for a "command name". trigger looks like a bad thing for a name. in this case a tag like `quartz.trigger` if in fact this is something clarifying.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListenerTest.java
ammmze: Thank you! This was very helpful. I've updated the test. Please review when you get a chance.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
ammmze: I've updated it to use the job key as the name and tagged it with the trigger key. Please let me know if there is something different you'd rather see. I'm completely open to other names/tags.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListenerTest.java
```diff
@@ -0,0 +1,135 @@
...
+		requireNoSpan();
+	}
+
+	@Test
+	public void jobWasExecuted_When_ContextContainsSpan_Then_ExpectProperSpanName() {
+		listener.jobToBeExecuted(context);
+		listener.jobWasExecuted(context, null);
+		Span span = takeSpan();
+		assertThat(span.name()).isEqualToIgnoringCase(JOB_KEY.toString());
+		assertThat(span.tags().get("Trigger")).isEqualToIgnoringCase(TRIGGER_KEY.toString());
```
adriancole: literally `quartz.trigger` is a better name. mixed case in lookup keys tend to be problematic

ammmze:
Hmm...I hadn't really considered what happens for manually triggering a job. I may need to also add some aspect stuff to be able to correlate anonymous triggers with their respective job by adding information to the JobDataMap.

Here is the trigger that runs when the app starts up. This makes sense that you would only see the backend because its just a headless job. The job is really the only thing happening. Realistically this is probably the majority of the use cases, but I know we usually expose some management endpoints so we can manually trigger a job if needed.
Trigger that runs on app startup

Here's what you end up seeing currently if you manually trigger a job from an api endpoint. The result is that the point where we trigger the job never gets correlated with the job itself.

We see the call that does the `scheduler.triggerJob(new JobKey("ExampleJob"));
the call that triggered the anonymous trigger

But the actual job once its picked up is completely separate.
anonymous trigger from hitting api endpoint

Unfortunately, quartz doesn't provide a way to hook into this process, so I think we'd probably need to do some aspect stuff around the triggerJob methods to add some data to the `JobDataMap` and then in the listener we would pick up that info and resume the trace.

adriancole:
oh i didn't mean a manually triggered job. I meant a scheduled one..
frontend just being a scheduler

ammmze:
Sounds good 👍 I can save adding support for manually triggered jobs for a later change

ammmze:
Doing some more experimentation, it looks like if I do a combination `TriggerListener` and `JobListener`, I can get a little bit wider scope for the span that gets created. So that if there are other listeners that an app registers, hopefully most of those would get encompassed within the span. I'll be pushing an update to cover that in a bit.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TraceQuartzAutoConfiguration.java
```diff
@@ -0,0 +1,65 @@
...
+@AutoConfigureAfter({ TraceAutoConfiguration.class, QuartzAutoConfiguration.class })
+@ConditionalOnProperty(value = "spring.sleuth.quartz.enabled", matchIfMissing = true)
+public class TraceQuartzAutoConfiguration implements InitializingBean {
+
+	private Scheduler scheduler;
+
+	private Tracing tracing;
+
+	@Autowired
+	public TraceQuartzAutoConfiguration(Scheduler scheduler, Tracing tracing) {
```
marcingrzejszczak: I don't think we actually need the `@Autowired`. Also, throught the project AFAIR we're setting the beans via fields in the auto configurations.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
```diff
@@ -0,0 +1,126 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/**
+ * {@link org.quartz.JobListener JobListener} that will wrap a span around quartz jobs
+ * when they start and finish.
+ *
+ * @author Branden Cash
+ */
+@Component
```
marcingrzejszczak: You don't need `@Component` since you've created the bean via the `@Bean` annotation.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
```diff
@@ -0,0 +1,126 @@
...
+import org.quartz.TriggerListener;
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.stereotype.Component;
+
+/**
+ * {@link org.quartz.JobListener JobListener} that will wrap a span around quartz jobs
+ * when they start and finish.
+ *
+ * @author Branden Cash
```
marcingrzejszczak: Please add the `@since` javadoc info.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TraceQuartzAutoConfiguration.java
```diff
@@ -0,0 +1,65 @@
...
+import org.springframework.boot.autoconfigure.quartz.QuartzAutoConfiguration;
+import org.springframework.cloud.sleuth.autoconfig.TraceAutoConfiguration;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * {@link org.springframework.boot.autoconfigure.EnableAutoConfiguration
+ * Auto-configuration} enables Quartz span information propagation.
+ *
+ * @author Branden Cash
```
marcingrzejszczak: Please add the `@since` javadoc.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/quartz/TraceQuartzAutoConfigurationTest.java
```diff
@@ -0,0 +1,123 @@
...
+ * @author Branden Cash
+ */
+public class TraceQuartzAutoConfigurationTest {
+
+	private final ApplicationContextRunner contextRunner = new ApplicationContextRunner()
+			.withConfiguration(AutoConfigurations.of(SchedulerConfig.class,
+					TracingConfig.class, TraceQuartzAutoConfiguration.class));
+
+	@Test
+	public void run_When_AllConditionalsTrue_Expect_TracingJobListenerIsCreated() {
```
marcingrzejszczak: can we have the `should_do_sth_when_sth_happens()` method naming?

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListenerTest.java
```diff
@@ -0,0 +1,299 @@
...
+		tracing = Tracing.newBuilder().spanReporter(spans::add)
+				.currentTraceContext(ThreadLocalCurrentTraceContext.newBuilder()
+						.addScopeDecorator(StrictScopeDecorator.create()).build())
+				.build();
+
+		listener = new TracingJobListener(tracing);
+		completableJob = new CompleteableTriggerListener();
+
+		scheduler = createScheduler(getClass().getSimpleName(), 1);
+		scheduler.addJob(newJob(ExceptionalJob.class).withIdentity(EXCEPTIONAL_JOB_KEY)
```
marcingrzejszczak: can we extract this to a separate method?

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListenerTest.java
```diff
@@ -0,0 +1,299 @@
...
+		scheduler.start();
+	}
+
+	@After
+	public void tearDown() throws Exception {
+		scheduler.shutdown(true);
+	}
+
+	@Test
+	public void getName_When_AllCases_Expect_ToReturnClassName() {
```
marcingrzejszczak: can we have the `should_do_sth_when_sth_happened()` method naming?

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListenerTest.java
```diff
@@ -0,0 +1,299 @@
...
+			throws Exception {
+		scheduler
+				.scheduleJob(newTrigger().forJob(EXCEPTIONAL_JOB_KEY).startNow().build());
+		completableJob.join();
+		takeSpan();
+	}
+
+	@Test
+	public void scheduleJob_When_JobIsVetoed_Expect_SpanToComplete() throws Exception {
+		scheduler.getListenerManager().addTriggerListener(new VetoJobTriggerListener());
```
marcingrzejszczak: Throughout all of these tests can we have the structure as follows?

```
// given
input conditions

// when 
a single line where sth triggers the behavior we want to assert

// then
assert the outcome
```

You can even leave the `//given, // when, // then` comments in the test code

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
marcingrzejszczak: 2.2.0

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
```diff
@@ -0,0 +1,125 @@
...
+ * @since 2.2.0
+ */
+public class TracingJobListener implements JobListener, TriggerListener {
+
+	static final String TRIGGER_TAG_KEY = "quartz.trigger";
+
+	static final String CONTEXT_SPAN_KEY = "X-TracingSpan";
```
adriancole: note: in brave, we usually use the class name for these instead of pseudo X-Headers

ex. Span.class.getName() etc

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
```diff
@@ -0,0 +1,125 @@
...
+
+	static final String CONTEXT_SPAN_IN_SCOPE_KEY = "X-TracingSpanInScope";
+
+	private static final Getter<JobDataMap, String> GETTER = (carrier, key) -> {
+		if (carrier != null) {
```
adriancole: carrier should never be null, please instead don't pass null to injector or extractor

adriancole:
the way I do it usually is github searches.. like "quartz brave" or "quartz zipkin" or a search for the dependencies used. Look for any github issues and see who raised them, or worked on it. Usually can find at least one person this way to ping

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
```diff
@@ -0,0 +1,123 @@
...
+
+	static final String TRIGGER_TAG_KEY = "quartz.trigger";
+
+	static final String CONTEXT_SPAN_KEY = Span.class.getName();
+
+	static final String CONTEXT_SPAN_IN_SCOPE_KEY = SpanInScope.class.getName();
+
+	private static final Getter<JobDataMap, String> GETTER = (carrier, key) -> {
+		Object value = carrier.get(key);
+		if (value instanceof String) {
```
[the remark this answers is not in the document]
adriancole: instanceof should prevent that. good suggestion of a unit test!

mark that field package private and give a whirl

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListenerTest.java
```diff
@@ -0,0 +1,330 @@
...
+		// expect
+		requireNoSpan();
+	}
+
+	@Test
+	public void should_have_parent_and_child_span_when_trigger_contains_span_info()
+			throws Exception {
+		// given
+		brave.Span span = tracing.tracer().nextSpan();
+		JobDataMap data = new JobDataMap();
```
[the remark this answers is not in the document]
adriancole: suppose the real question is if JobDataMap is type bound consistently from the job. if that's the case, then we can't stick strings in it. need some real world stuff I guess. excellent discovery

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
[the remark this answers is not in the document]
adriancole: fwiw instanceof doubles as a null check, so if the map is good otherwise, lgtm

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
```diff
@@ -0,0 +1,123 @@
...
+			return (String) value;
+		}
+		return null;
+	};
+
+	private Tracing tracing;
```
marcingrzejszczak: We can make this final I guess

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
```diff
@@ -0,0 +1,123 @@
...
+
+	@Autowired
```
marcingrzejszczak: `@Autowired` is unnecessary here

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
```diff
@@ -0,0 +1,123 @@
...
+import org.springframework.beans.factory.annotation.Autowired;
+
+/**
+ * {@link org.quartz.JobListener JobListener} that will wrap a span around quartz jobs
+ * when they start and finish.
+ *
+ * @author Branden Cash
+ * @since 2.2.0
+ */
+public class TracingJobListener implements JobListener, TriggerListener {
```
marcingrzejszczak: Do we need this to be public? I guess we could make this package scoped

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListenerTest.java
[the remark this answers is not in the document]
marcingrzejszczak: So I understand that doing the `instanceof String` is good enough for this usecase?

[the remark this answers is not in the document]
ammmze: Yea ... I've added a test case to cover the concern where the `JobDataMap` is constructed with a map that does not have a compatible typing. As expected, the code still works since `JobDataMap` internally is using a `Map<String, Object>` and all values from the map you give in the constructor to `JobDataMap` are simply copied over.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
ammmze: Good suggestion. Latest commit makes this change.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
ammmze: 👍  Latest commit makes this change.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/quartz/TracingJobListener.java
ammmze: Good suggestion. Latest commit makes this change.

ammmze:
Thanks guys for all the suggestions! I'm very happy about all the feedback on here. I've made the suggested changes. Let me know if we're missing anything else.

marcingrzejszczak:
The build is broken after the recent changes - https://circleci.com/gh/spring-cloud/spring-cloud-sleuth/5346?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

ammmze:
Yea...looking into it. The build worked locally when I did a `mvn verify`. But after doing a `mvn -U clean verify` it broke. So either 1) there was something cached that allowed it to bypass something, or something in the updated snapshot packages that may have gotten updated caused it to break.

ammmze:
It looks like `TraceWebClientBeanPostProcessor` needs some updates in response to some changes in spring-webflux [here](https://github.com/spring-projects/spring-framework/commit/f5640cbfe04bedec12aa00bd60e5fb7f63aee96b#diff-cf5eb857e736cebb0f0bccd73fadd616) and [here](https://github.com/spring-projects/spring-framework/commit/37398c669ca3af857487d963f3645e6d1cb678e5#diff-cf5eb857e736cebb0f0bccd73fadd616)

The anonymous `ClientResponse` class that is getting created in there needs to implement the new methods: `releaseBody` and `toBodilessEntity`

ammmze:
@marcingrzejszczak @adriancole Any opposition to me squashing my commits to keep history clean? Not sure what your practice is here.

adriancole:
personally I always squash. the code is still on your fork so you can do
anything :)

marcingrzejszczak:
I've fixed the master branch build, please rebase and I'll merge it with squash.

ammmze:
Thank you! I've rebased and squashed my commits.

And thanks again to everyone to shared feedback on this.

marcingrzejszczak:
Great work @ammmze , thank you so much for this PR!

Issue #1431, opened 2019-08-26
Different modes for the HystrixConcurrencyStrategy


worldtiki:
Context: I'm using Hystrix to wrap a number of http clients. Each of these clients is instrumented and generates tracing information.

Problem: For each network call there are 2 spans, one generated by the SleuthHystrixConcurrencyStrategy and another by the instrumented http client.
This leads to a duplication of spans and unnecessary nesting when visualising the traces.

Proposal: Include an option to configure SleuthHystrixConcurrencyStrategy so that users can decide if they want to create spans for each Hystrix command execution or just have the SleuthHystrixConcurrencyStrategy to propagate the trace context between threads.

I'm happy to open a pr if there is interest in this.

adriancole:
interesting. pass-through basically. between disabled and enabled

worldtiki:
Exactly.

I think my story is similar to many others, specially for monolith apps. Assuming all network clients are wrapped with Hystrix, it's easy to start capturing spans with the HystrixConcurrencyStrategy.
However at some point the many network clients will be instrumented which means that the Hystrix spans stop providing useful info and become mostly noise.

PR #1433, opened 2019-08-26
Add passthrough mode to HystrixConcurrencyStrategy

worldtiki:
Draft proposal for #1431.

Passthrough is disabled by default which means that the current behaviour will not change with this change.
When `spring.sleuth.hystrix.strategy.passthrough` is enabled, the HystrixConcurrencyStrategy will just pass the trace context to the Hystrix execution thread instead of using TraceCallables.
This way we propagate tracing information but don't create . spans for each execution.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/hystrix/SleuthHystrixConcurrencyStrategyProperties.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/hystrix/SleuthHystrixConcurrencyStrategyProperties.java
@@ -26,8 +26,16 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 @ConfigurationProperties("spring.sleuth.hystrix.strategy")
 public class SleuthHystrixConcurrencyStrategyProperties {
 
+	/**
+	 * Enable custom HystrixConcurrencyStrategy that wraps all Callable instances into
+	 * their Sleuth representative - the TraceCallable.
+	 */
 	private boolean enabled = true;
 
+	/**
+	 * When enabled the tracing information is passed to the Hystrix execution threads but
+	 * spans are not created for each execution.
+	 */
 	private boolean passthrough = false;
 
 	public boolean isEnabled() {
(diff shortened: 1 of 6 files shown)


worldtiki:
An enum would probably be a better fit, e.g.:
`spring.sleuth.hystrix.strategy`=`passthrough` / `active` / `disabled` / `...`
But it would likely involve breaking changes.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/hystrix/SleuthHystrixConcurrencyStrategyProperties.java
```diff
@@ -0,0 +1,49 @@
...
+ * Sleuth Hystrix settings.
+ *
+ * @author Daniel Albuquerque
+ */
+@ConfigurationProperties("spring.sleuth.hystrix.strategy")
+public class SleuthHystrixConcurrencyStrategyProperties {
+
+	private boolean enabled = true;
+
+	private boolean passthrough = false;
```
marcingrzejszczak: Can you add a javadoc?

spring-cloud-sleuth-core/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -30,6 +30,12 @@
       "description": "Enable custom HystrixConcurrencyStrategy that wraps all Callable instances into their Sleuth representative - the TraceCallable.",
       "defaultValue": true
     },
+    {
```
marcingrzejszczak: Since you're adding a `ConfigurationProperty` you don't need to provide it here. It's enough to pass it in the javadocs of the class.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/hystrix/SleuthHystrixConcurrencyStrategyProperties.java
worldtiki: Of course

marcingrzejszczak:
Congrats @worldtiki on this PR! Great work :)

PR #1436, opened 2019-08-31
WIP: customizer all the things

adriancole:
This adds hooks for new customizers that will be in Brave 5.7. As I'll
be offline for a day or two, if anyone wants to play with this they can.

We could just weave-in and document this, or we could retrofit some of
our existing stuff to configure itself via customizers instead (ex MDC
integration).

cc @devinsba

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
@@ -24,9 +24,12 @@ import brave.CurrentSpanCustomizer;
 import brave.ErrorParser;
 import brave.Tracer;
 import brave.Tracing;
+import brave.TracingCustomizer;
 import brave.handler.FinishedSpanHandler;
 import brave.propagation.B3Propagation;
 import brave.propagation.CurrentTraceContext;
+import brave.propagation.CurrentTraceContextCustomizer;
+import brave.propagation.ExtraFieldCustomizer;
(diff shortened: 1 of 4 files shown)


marcingrzejszczak:
Lgtm

I'd refactor what we have to use the customizers I guess

narayaruna:
@marcingrzejszczak Netflix needs this. Could you include it in 2.1.3. Thanks.

@adriancole fyi

shakuzen:
Brave 5.7 is being released now.

adriancole:
this is now on a stable release. needs tests and docs.. not sure if you have time for it @marcingrzejszczak but I'm offline until monday. hoping to get this in (also 2.1) to unlock the secondary sampling thing nara's hoping to roll out.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
```diff
@@ -107,6 +119,9 @@ Tracing tracing(
 		for (FinishedSpanHandler finishedSpanHandlerFactory : this.finishedSpanHandlers) {
 			builder.addFinishedSpanHandler(finishedSpanHandlerFactory);
 		}
+		for (TracingCustomizer customizer : this.tracingCustomizers) {
```
adriancole: ps without tests this is subtle. it is important these apply last as they can rejig the reporter config, something needed for secondary sampling to work. ex we want to flip default reporter to noop but reuse the normal reporter as a finished span handler. this stuff works when customizes hit last before build

adriancole:
bravo!

PR #1437, opened 2019-09-05
Fix TraceableExecutorService

koozz:
TraceableExecutorService swallows Exception because `execute` submits a Runnable and ignores the Future's result. It should call the delegates `execute` method instead of the `submit`.


marcingrzejszczak:
:scream: thanks for the great catch!

koozz:
> scream thanks for the great catch!

Credit goes to: @tdons

PR #1447, opened 2019-09-18
Fix ThreadPoolTaskScheduler proxy mechanism

nkavtur:
Purpose:
ThreadPoolTaskExecutor is not instrumented properly. Only methods inherited from AsyncTaskExecutor are intercepted, but TaskScheduler ones are lost. Added additional check to fix the issue.


marcingrzejszczak:
Great catch @nkavtur , thanks a lot

PR #1456, opened 2019-10-02
Updates to latest Sampling infrastructure

adriancole:
Brave recently switched to an interface model for higher level sampling
like HTTP. So, `HttpSampler` -> `SamplingFunction<HttpRequest>`. Don't
worry because `HttpSampler` was retrofitted as a `SamplingFunction`.

To ensure bean compatibility, this deprecates the existing `@ClientSampler` and `@ServerSampler` annotations in favor of `@HttpClientSampler` and `@HttpServerSampler`.

More interestingly, this weaves in support for `RpcTracing`, which *finally*
introduces RPC sampling the same way. Specifically, this adds `SamplingFunction<RpcRequest>` under the same conventions as HTTP.
Those wanting to configure can use the new `@RpcClientSampler` and
`@RpcServerSampler` annotations.

Most immediately, this can be used here in gRPC and Dubbo, as
autoconfiguration exists. It also works with any autoconfiguration that
isn't here, such as Armeria.

Ex. Here's a sampler that traces 100 "GetUserToken" requests per second. This
doesn't start new traces for requests to the health check service. Other
requests will use a global rate provided by the tracing component.

```java
import static brave.rpc.RpcRequestMatchers.methodEquals;
import static brave.rpc.RpcRequestMatchers.serviceEquals;
import static brave.sampler.Matchers.and;

import org.springframework.cloud.sleuth.instrument.rpc.RpcServerSampler;

--snip--
@Bean(name = RpcServerSampler.NAME)
SamplerFunction<RpcRequest> myRpcSampler() {
  Matcher<RpcRequest> userAuth = and(
      serviceEquals("users.UserService"),
      methodEquals("GetUserToken")
  );
  return RpcRuleSampler.newBuilder()
      .putRule(serviceEquals("grpc.health.v1.Health"), Sampler.NEVER_SAMPLE)
      .putRule(userAuth, RateLimitingSampler.create(100)).build();
}
```

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-sleuth.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-sleuth.adoc
@@ -841,7 +841,7 @@ Sleuth will search for beans of those types and automatically apply customizatio
 If a customization of client / server parsing of the HTTP related spans is required,
 just register a bean of type `brave.http.HttpClientParser` or
 `brave.http.HttpServerParser`. If client /server sampling is required, just
-register a bean of type `brave.http.HttpSampler` and name the bean
+register a bean of type `brave.sampler.SamplerFunction<HttpRequest>` and name the bean
  `sleuthClientSampler` for client sampler and `sleuthServerSampler` for server sampler.
  For your convenience the `@ClientSampler` and `@ServerSampler`
  annotations can be used to inject the proper beans or to
@@ -852,7 +852,7 @@ https://github.com/openzipkin/brave/tree/master/instrumentation/http#sampling-po
 
 If you want to completely rewrite the `HttpTracing` bean you can use the `SkipPatternProvider`
 interface to retrieve the URL `Pattern` for spans that should be not sampled. Below you can see
-an example of usage of `SkipPatternProvider` inside a server side, `HttpSampler`.
+an example of usage of `SkipPatternProvider` inside a server side, `Sampler<HttpRequest>`.
 
 [source,java]
 ----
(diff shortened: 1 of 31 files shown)


adriancole:
notifying...

@saturnism @elefeint as you might be interested in that you can now assign sample rates to gRPC
@chickenlj @geercode we now switch to recommend Apache Dubbo (noting that our tests don't pass until 2.7.4)
@trustin @kojilin @anuraaga about armeria we may want to update also (both HTTP and RPC invokers)
@narayaruna I know Netflix has been waiting for consolidated sampler support and this does it!

Thanks also to @jeqo who did the release of Brave 5.8 which is ideally synced now.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/rpc/ServerSampler.java
```diff
@@ -0,0 +1,50 @@
...
+ * @since 2.2.0
+ * @see Qualifier
+ */
+@Target({ ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER, ElementType.TYPE,
+		ElementType.ANNOTATION_TYPE })
+@Retention(RetentionPolicy.RUNTIME)
+@Inherited
+@Documented
+@Qualifier(ServerSampler.NAME)
+public @interface ServerSampler {
```
adriancole: @marcingrzejszczak @devinsba @minwoox @llinder So, historically, the annotation was ServerSampler for http. It may be a good idea to name the new one RpcServerSampler. If so, and I am asleep.. just change it here and also for client :P

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHttpAutoConfiguration.java
```diff
@@ -165,7 +167,7 @@ HttpSampler sleuthClientSampler(SleuthWebProperties sleuthWebProperties) {
 			return leftDecision;
 		}
 		// Neither are null and at least one is true
-		return leftDecision && rightDecision;
+		return rightDecision;
```
adriancole: intellij did this :)

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
```diff
@@ -841,7 +841,7 @@ Sleuth will search for beans of those types and automatically apply customizatio
 If a customization of client / server parsing of the HTTP related spans is required,
 just register a bean of type `brave.http.HttpClientParser` or
 `brave.http.HttpServerParser`. If client /server sampling is required, just
-register a bean of type `brave.http.HttpSampler` and name the bean
+register a bean of type `brave.sampler.SamplerFunction<HttpRequest>` and name the bean
```
marcingrzejszczak: Isn't this a breaking change?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/rpc/TraceRpcAutoConfiguration.java
```diff
@@ -0,0 +1,75 @@
...
+import org.springframework.cloud.sleuth.autoconfig.TraceAutoConfiguration;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.lang.Nullable;
+
+/**
+ * {@link org.springframework.boot.autoconfigure.EnableAutoConfiguration
+ * Auto-configuration} related to RPC based communication.
+ *
+ * @author Marcin Grzejszczak
```
marcingrzejszczak: LOL - I'm producing code even if I don't write it

docs/src/main/asciidoc/spring-cloud-sleuth.adoc
adriancole: I suppose you are right. because if a 3rd party is expecting the type `brave.http.HttpSampler`, then it doesn't matter that it implements `brave.sampler.SamplerFunction<HttpRequest>`. The scenario I'm
thinking is if someone is manually calling HttpTracing.builder?

One way out is to deprecate and rename ServerSampler -> HttpServerSampler
only check null on ServerSampler and use the old types as needed.

I can raise a change for that.

adriancole:
OK I switched the approach and backfilled tests to show this is not a breaking change, rather a deprecation of the existing annotations.

I will be away flying so if any small changes, please go ahead. I hope the big changes are in!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/rpc/ServerSampler.java
adriancole: did this

adriancole:
woot thanks!

PR #1464, opened 2019-10-10
Adds the possibility to opt-out of executor decoration

adriancole:
Recently, `AsyncAutoConfiguration` auto-instruments executors. This
causes spans named 'async' with no tag or other metadata to help
understand what is happening. It was somewhat difficult troubleshooting
this because the control is by bean name, but for example the "async"
span doesn't include the bean name.

For starters, this allows people to opt-out of automatically spanning
executors via `spring.sleuth.async.enabled=false`

See https://github.com/spring-cloud/spring-cloud-gcp/issues/1978
cc @meltsufin @saturnism

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/AsyncAutoConfiguration.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/AsyncAutoConfiguration.java
@@ -18,6 +18,7 @@ package org.springframework.cloud.sleuth.instrument.async;
 
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.cloud.sleuth.instrument.scheduling.SleuthSchedulingProperties;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 
@@ -29,8 +30,8 @@ import org.springframework.context.annotation.Configuration;
  * @since 2.1.0
  */
 @Configuration
-@EnableConfigurationProperties(SleuthAsyncProperties.class)
-@ConditionalOnProperty(value = "spring.sleuth.async.enabled", matchIfMissing = true)
+@ConditionalOnProperty(value = "spring.sleuth.scheduled.enabled", matchIfMissing = true)
+@EnableConfigurationProperties({SleuthAsyncProperties.class, SleuthSchedulingProperties.class})
 public class AsyncAutoConfiguration {
 
 	@Bean
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/AsyncAutoConfiguration.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/AsyncAutoConfiguration.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.sleuth.instrument.async;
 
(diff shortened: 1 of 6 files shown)


adriancole:
added this to help pinpoint the concern later https://github.com/spring-cloud/spring-cloud-sleuth/issues/1465

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/AsyncAutoConfiguration.java
```diff
@@ -29,6 +30,7 @@
  */
 @Configuration
 @EnableConfigurationProperties(SleuthAsyncProperties.class)
+@ConditionalOnProperty(value = "spring.sleuth.async.enabled", matchIfMissing = true)
```
marcingrzejszczak: Why not reuse `org.springframework.cloud.sleuth.instrument.scheduling.SleuthSchedulingProperties` and `spring.sleuth.scheduled.enabled` ?

adriancole: this feature itself is confusing really, but the properties related to what you need to stop are under the async namespace, and this feature itself, not necessarily the parent feature, is what needs to be stopped.

adriancole: for example, if you look at the class https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/SleuthAsyncProperties.java it has only one property which allows you to cherry-pick what to ignore. this is effectively saying `*`. Since this is recent, I don't know the full impact of disabling the upstream stuff. If you think it is the same, then it is curious why have the "async" property here instead of in SleuthSchedulingProperties

adriancole:
one thing discussed is there is tension contributing to the exclude list.. if there was a way, we could avoid this (though the feature still should be possible to disable, possibly shouldn't be enabled by default).

Think of gcp, for example.. they don't want to interfere with app config. the "disabled" thing is mostly to note that there should be a way to stop beans from automatically becoming instrumented.

In gcp instead disable, it would be nice to add a bean name to a skip config. However, there's also no clean way to do that. For example, the app defines a skip list.. how does a third party add to that list at runtime similar to how customizers work?

Another way could be to introduce an annotation that is like `@DisableTracing` when features are added that instrument everything, such as executors.

Some solution to this could ease pressure on this feature and ones similar, though this PR is mainly just to allow people to stop spamming their zipkin if they only recently upgraded and now they are getting "async" spans everywhere.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/AsyncAutoConfiguration.java
adriancole: I switched, though now the test is broke locally

adriancole:
broke with cause

```
Caused by: org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.springframework.cloud.sleuth.instrument.async.SleuthAsyncProperties' available
```

saturnism:
how abt the option of opt out by default? and enable w/ annotation?
i.e., existing code that's already producing thread pool related beans didn't expect to be traced. if we opt out by default then there is no issues. those who needs to be traced can be opt'ed in.

marcingrzejszczak:
> how abt the option of opt out by default? and enable w/ annotation?

this would be a breaking change

PR #1475, opened 2019-10-21
Service name for the remote redis endpoint.

### LINKED ISSUE #1406: Ability to provide a custom name for redis spans
https://github.com/lettuce-io/lettuce-core/pull/866 adds an enhancement where we can override the service name for calls to Redis (default is "redis").

An example of where we may want to do this is on setups where there are multiple installations of Redis. Using the default name would mean that in a service graph visualization, different services interacting with separate Redis would appear as if they are interacting with the same one.

worldtiki:
Fixes #1406


tests/spring-cloud-sleuth-instrumentation-lettuce-tests/src/test/java/org/springframework/cloud/sleuth/instrument/redis/TraceRedisAutoConfigurationTests.java
```diff
@@ -62,10 +62,19 @@ ClientResources clientResources() {
 			return clientResources;
 		}
 
+		@Bean
+		TraceRedisProperties traceRedisProperties() {
+			TraceRedisProperties traceRedisProperties = new TraceRedisProperties();
+			traceRedisProperties.setEnabled(true);
+			traceRedisProperties.setRemoteServiceName("redis-foo");
```
adriancole: I think there should be a test that this actually ends up with a remote service name. It is currently implicit via looking at the lettuce change.

worldtiki: I agree but I'm not sure how to do that.
I thought about spinning up an embedded redis and doing an end to end test, but that doesn't seem to be common in this project.
I would love if this was using test containers but I don't think it's up to me to decide when to introduce new libs.

adriancole: I suppose how easy this is depends on how the lettuce instrumentation works. If it works as I'd expect it, you could make a connection to an intentionally not available redis cluster and expect a span with an error tag. If the instrumentation only works when the redis host is, that's probably a bug, but anyway worth a thought!

worldtiki: That wont work unfortunately :(
The client first needs to establish a connection before we can start using it and before we're able to trace commands.

adriancole: to quote lyrics from Angie.. "You can't say we never tried...." :)

marcingrzejszczak:
Thanks for the contribution @worldtiki !

PR #1477, opened 2019-10-23
Use Micrometer metrics instead of in memory one

marcingrzejszczak:
Since I'm no metrics expert I need feedback from such as experts as @shakuzen @adriancole @msmsimondean and the rest of the experts interested in this topic ;)

I've added an initial implementation that uses `counter`s and `gauge`s from micrometer for metrics. I've passed some default values for metric names but I'm not sure if those are the best.

This would actually start defaulting to the Micrometer version of a metric reporter since it's pretty much always on the classpath. If someone doesn't have a `meterRegistry` bean, then the in memory option will be taken into account.

I don't think we can have an option that there is no micrometer on the classpath since it comes with boot.

UPDATE: As @spencergibb mentioned, micrometer comes as an dependency with an actuator, however actuator doesn't always have to be there. As @adriancole mentioned there already is a zipkin micrometer reporter that we can reuse.

### DIFF
--- a/spring-cloud-sleuth-core/pom.xml
+++ b/spring-cloud-sleuth-core/pom.xml
@@ -24,6 +24,9 @@
 	<packaging>jar</packaging>
 	<name>Spring Cloud Sleuth Core</name>
 	<description>Spring Cloud Sleuth Core</description>
+	<properties>
+		<spring-security-oauth2.version>2.2.0.RELEASE</spring-security-oauth2.version>
+	</properties>
 
 	<parent>
 		<groupId>org.springframework.cloud</groupId>
@@ -131,7 +134,7 @@
 		<dependency>
 			<groupId>org.springframework.security.oauth</groupId>
 			<artifactId>spring-security-oauth2</artifactId>
-			<version>2.2.0.RELEASE</version>
+			<version>${spring-security-oauth2.version}</version>
 			<optional>true</optional>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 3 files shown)


msmsimondean:
@marcingrzejszczak would it be worth using the Noop metrics reporter if `meterRegistry` bean is not registered.  What do you think @adriancole?

spencergibb:
Micrometer is optional in boot, it is pulled in with the actuator starter.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
```diff
@@ -283,4 +274,106 @@ public String toString() {
...
+		@ConditionalOnMissingBean
+		ReporterMetrics sleuthReporterMetrics() {
+			return new InMemoryReporterMetrics();
+		}
+
+	}
+
+	@Configuration
+	@ConditionalOnBean(MeterRegistry.class)
+	static class TraceMicrometerMetricsConfiguration {
```
spencergibb: This should be a toplevel auto configuration conditional on a micrometer class

marcingrzejszczak:
> @marcingrzejszczak would it be worth using the Noop metrics reporter if meterRegistry bean is not registered. What do you think @adriancole?

that's already the case

> Micrometer is optional in boot, it is pulled in with the actuator starter.

@spencergibb I wasn't sure of that, initially I did exactly what you suggested (with class check) and then I removed it. I'll bring it back.

adriancole:
is there a reason we aren't using this? if so, maybe we mention https://github.com/openzipkin/zipkin-reporter-java/tree/master/metrics-micrometer

marcingrzejszczak:
Most likely cause I didn't know that it's a thing

spring-cloud-sleuth-core/pom.xml
```diff
@@ -286,20 +289,28 @@
 			<artifactId>brave-instrumentation-grpc</artifactId>
 			<optional>true</optional>
 		</dependency>
+		<dependency>
+			<groupId>io.zipkin.reporter2</groupId>
+			<artifactId>zipkin-reporter-metrics-micrometer</artifactId>
+			<exclusions>
+				<exclusion>
+					<groupId>io.micrometer</groupId>
```
adriancole: @shakuzen reminds me.. we probably should have had this set to "provided" scope in upstream anyway

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
```diff
@@ -283,4 +282,40 @@ public String toString() {
 
 	}
 
+	@Configuration
+	@ConditionalOnMissingClass("io.micrometer.core.instrument.MeterRegistry")
+	static class TraceMetricsNoOpConfiguration {
```
adriancole: should this be

```suggestion
	static class TraceMetricsInMemoryConfiguration {
```

if no-op is desired, that would be different and cheaper.

Issue #1125, opened 2018-11-06
Manually created load-balanced Feign Clients - Sleuth Instrumentation


tf-haotri-pham:
Hi,
I have a Spring Boot 1.5.17 application that also includes the **spring-cloud-starter-zipkin** for trace ID generation. The application further uses Feign for any outbound REST calls, however the [Feign clients are created manually](https://cloud.spring.io/spring-cloud-netflix/multi/multi_spring-cloud-feign.html#_creating_feign_clients_manually) rather than via _@FeignClient_ annotation. Ribbon is included and I build the Feign clients with target URL as 
```java
Feign.builder()... .target(MyFeignClient.class, "http://my-service");
```
for which I have _my-service.ribbon.listOfServers_ set in my _application.properties_ file.

I have observed that the REST calls via the manually created (Ribbon load-balanced) Feign clients **do not include the trace information HTTP headers** (such as _x-b3-tracid_, etc.) and so I looked at the instrumentation config beans under [TraceFeignClientAutoConfiguration.java](https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/TraceFeignClientAutoConfiguration.java) to find out what I would need to do additionally for such manually created clients.

During debugging, I could however see that [TraceFeignAspect.java](https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/TraceFeignAspect.java) does wrap the instantiated feign clients in respective _TraceLoadBalancerFeignClient_ or _TraceFeignClient_ instances, but these are never invoked ([code line 55](https://github.com/spring-cloud/spring-cloud-sleuth/blob/3c25fc1a11b8d39e15bbb5c12b1225db582e72df/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/TraceFeignAspect.java#L55)):
```java
if (bean != wrappedBean) {
    return executeTraceFeignClient(bean, pjp);
}
```

I wanted to ask, if the line was meant to be
```java
return executeTraceFeignClient(wrappedBean, pjp);
```
[...]

PR #1490, opened 2019-11-20
Issue 1125 test 2 1 x fix partial

colejnic:
Not meant for merge; for showing issues relating to https://github.com/spring-cloud/spring-cloud-sleuth/issues/1125


marcingrzejszczak:
ping @colejnic :arrow_up:

colejnic:
@marcingrzejszczak  I'm hoping to address this soon (early next week)! got pulled away for a while

colejnic:
@marcingrzejszczak I did some cleanup; there's still a spot I left commented in TraceFeignObjectWrapper where I'm not sure what should happen if we fail to modify the delegate via reflection.
Let me know if there's more you had in mind!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/TraceFeignObjectWrapper.java
```diff
@@ -57,12 +60,31 @@ Object wrap(Object bean) {
...
+				}
+				else {
+					LoadBalancerFeignClient client = ((LoadBalancerFeignClient) bean);
+					try {
+						Field delegate = LoadBalancerFeignClient.class.getDeclaredField("delegate");
+						delegate.setAccessible(true);
+						delegate.set(client, new TraceFeignObjectWrapper(this.beanFactory).wrap(client.getDelegate()));
+					}
+					catch (NoSuchFieldException | IllegalArgumentException | IllegalAccessException | SecurityException e) {
+						// not sure the best option here
```
marcingrzejszczak: I think that ignoring this is a good idea. We can try to just return a wrapped instance or return a non wrapped one with a warning message. I think that we shouldn't break the application startup due to tracing. In case of any exception I could log a warning message and fallback to only wrapping the delegate as you suggested

PR #1493, opened 2019-11-22
close span after unexpected exception thrown

addozhang:
fix issue 1492


marcingrzejszczak:
We don't support that version of Sleuth anymore, please upgrade.

PR #1505, opened 2019-12-10
Spring Cloud Circuit Breaker tracing

marcingrzejszczak:
First take on the integration with `spring-cloud-circuitbreaker`.

We could provide options such as `TraceSupplier` and `TraceFunction` to manually wrap any commands that are executed via a circuit breaker. If Spring beans are involved we will wrap the `CircuitBreakerFactory` that the generated circuit breaker implementation always wraps the supplier and function in a trace representation.

### DIFF
--- a/docs/src/main/asciidoc/spring-cloud-sleuth.adoc
+++ b/docs/src/main/asciidoc/spring-cloud-sleuth.adoc
@@ -1083,6 +1083,11 @@ include::{project-root}/spring-cloud-sleuth-core/src/test/java/org/springframewo
 
 That way, you ensure that a new span is created and closed for each execution.
 
+=== Spring Cloud CircuitBreaker
+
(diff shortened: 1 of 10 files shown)


marcingrzejszczak:
The problem I see is that unlike what we have with Hystrix, we can't use any metadata - e.g. command key, group etc.

ryanjbaxter:
Is the limitation on metadata due to something in Spring Cloud CircuitBreaker or resilience4j?

What about using Spring Cloud CircuitBreaker with Hystrix?  Or any other implementation?

marcingrzejszczak:
From the point of view of the current  spring cloud circuit breaker api we have no access to meta data.

marcingrzejszczak:
Maybe some insight from @shakuzen would be helpful? I mean, the current abstraction of Spring Cloud CircuitBreaker only allows you to pass callbacks for running the task and for fallback. Maybe we should allow those callbacks to contain some meta-data? Or have another method that allows to pass meta-data?

PR #1513, opened 2020-01-06
Report HTTP pattern and handler function name for functional WebFlux routes.

csabakos:
This pull request improves on the instrumentation of WebFlux requests when the routes are defined using `RouterFunction` rather than in a `@Controller`. In particular:
- Before this patch, the operation name used to be just the HTTP verb like `get`, but now it will include the pattern like: `get /api/books/{id}`. This brings parity with the `@Controller` case.
- The `mvc.controller` tag used to be unset but now it will be set to the name of the router function, such as `BooksRouterService$$Lambda$123/123456789`. While this is not perfect, at least it points you to the right file to look at.


marcingrzejszczak:
Hi! Thanks for the pull request. Is this related to a pull request of an issue? What problem are you trying to solve?

csabakos:
@marcingrzejszczak Sorry for the lack of explanation. I added a description to the initial comment. I hope that clarifies it, but please let me know if there are further questions.

marcingrzejszczak:
Thanks for the clarification and for the pull request :)

PR #1522, opened 2020-01-13
Use actual log timestamp as logstash timestamp

rafaelrenanpacheco:
Based on the grok example from the docs, a Filebeat + Logstash stack will produce the following:
- a `@timestamp` field based on the time that filebeat sends the log to logstash, as timestamp type.
- a `timestamp` field from the Spring Boot log, as string type.

In order to use the Spring Boot timestamp as the real Logstash `@timestamp` I use the following filters after the grok pattern:

```
  date {
    match => ["timestamp", "ISO8601"]
  }
  mutate {
    remove_field => ["timestamp"]
  }
```

Using these filters will result in `@timestamp` being the Spring Boot's timestamp, as a timestamp type, and will remove the now duplicated (string type) `timestamp`.

### DIFF
--- a/docs/src/main/asciidoc/intro.adoc
+++ b/docs/src/main/asciidoc/intro.adoc
@@ -158,19 +158,31 @@ If you want to use https://www.elastic.co/guide/en/logstash/current/index.html[L
 
 [source]
 filter {
-       # pattern matching logback pattern
-       grok {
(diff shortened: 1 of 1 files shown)


marcingrzejszczak:
Very good point, thanks!

PR #1525, opened 2020-01-17
Reduces overhead when using reactor by avoiding redundant context syncs

adriancole:
This switches to pass around and use `TraceContext` instead of `Span` in
Reactor's context to take advantage of `maybeScope` which is far cheaper
than `withSpanInScope` when the span is already current. This also kills
some extra scope checks and any other accidental calls noticed.

The performance improvement here, I can't quantify as haven't done any
JMH on it.

cc @anuraaga

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingSpanSubscriber.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingSpanSubscriber.java
@@ -18,9 +18,8 @@ package org.springframework.cloud.sleuth.instrument.reactor;
 
 import javax.annotation.Nullable;
 
-import brave.Span;
-import brave.Tracing;
 import brave.propagation.CurrentTraceContext;
+import brave.propagation.CurrentTraceContext.Scope;
 import brave.propagation.TraceContext;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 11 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingSpanSubscriber.java
```diff
@@ -46,69 +45,62 @@
...
-		this.currentTraceContext = tracing.currentTraceContext();
-
-		this.traceContext = root == null ? null : root.context();
-		this.context = ctx != null && root != null ? ctx.put(Span.class, root)
+		this.currentTraceContext = currentTraceContext;
+		this.parent = parent;
+		this.context = ctx != null && parent != null ? ctx.put(TraceContext.class, parent)
 				: ctx != null ? ctx : Context.empty();
 		if (log.isTraceEnabled()) {
-			log.trace("Root span [" + root + "], context [" + this.context + "]");
```
adriancole: not root

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/SpanSubscriber.java
```diff
@@ -44,35 +46,40 @@
...
 
+	private final CurrentTraceContext currentTraceContext;
+
 	private Subscription s;
 
 	SpanSubscriber(Subscriber<? super T> subscriber, Context ctx, Tracing tracing,
 			String name) {
 		this.subscriber = subscriber;
 		this.tracer = tracing.tracer();
-		Span root = ctx.getOrDefault(Span.class, this.tracer.currentSpan());
```
adriancole: this would have checked the context MDC etc even if the context was in reactor

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/SpanSubscriber.java
```diff
@@ -44,35 +46,40 @@
...
 		}
-		this.rootSpan = root;
+		this.parent = parent;
 		if (log.isTraceEnabled()) {
-			log.trace("Stored context root span [{}]", this.rootSpan);
+			log.trace("Stored context parent span [{}]", this.parent);
 		}
-		this.span = root != null ? this.tracer
-				.nextSpan(TraceContextOrSamplingFlags.create(root.context())).name(name)
-				: this.tracer.nextSpan().name(name);
```
adriancole: nextSpan makes a get call to the current context, which is not needed when we know we are in a new root scenario.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/SpanSubscriber.java
```diff
@@ -81,7 +88,7 @@ public void onSubscribe(Subscription subscription) {
 			log.trace("On subscribe");
 		}
 		this.s = subscription;
-		try (Tracer.SpanInScope ws = this.tracer.withSpanInScope(this.span)) {
+		try (Scope ws = this.currentTraceContext.maybeScope(this.span.context())) {
```
adriancole: if the span was already in scope, this would have synced MDC etc again for no reason

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/SpanSubscriber.java
```diff
@@ -94,15 +101,15 @@ public void request(long n) {
 		if (log.isTraceEnabled()) {
 			log.trace("Request");
 		}
-		try (Tracer.SpanInScope ws = this.tracer.withSpanInScope(this.span)) {
```
adriancole: if the span was already in scope, this would have synced MDC etc again for no reason

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/SpanSubscriber.java
```diff
@@ -154,10 +161,10 @@ void cleanup() {
 			if (log.isTraceEnabled()) {
 				log.trace("Span closed");
 			}
-			if (this.rootSpan != null) {
-				this.rootSpan.finish();
+			if (this.parent != null) {
+				this.tracer.toSpan(parent).finish(); // TODO: why are we closing this?
```
adriancole: ack this class is deprecated, so the TODO maybe not important

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebFilter.java
```diff
@@ -192,19 +191,17 @@ public void subscribe(CoreSubscriber<? super Void> subscriber) {
...
 			}
 			return context;
 		}
 
 		private Span findOrCreateSpan(Context c) {
 			Span span;
-			if (c.hasKey(Span.class)) {
-				Span parent = c.get(Span.class);
-				span = this.tracer
-						.nextSpan(TraceContextOrSamplingFlags.create(parent.context()))
```
adriancole: not hugely important, but this is more expensive than newChild

PR #1530, opened 2020-01-21
Cleanups of build including upgrade of Brave

adriancole:
In efforts to dig deep into benchmarks, I noticed we were a bit out of
date on Brave. Then, noticed some other artifacts could be bumped
safely. Finally, a find/replace on http/https was over zealous and
tripped up XML.

One change to test expectations and Brave is explained here:
https://github.com/openzipkin/brave/blob/master/instrumentation/RATIONALE.md#calling-spanfinish-while-the-context-is-in-scope

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -888,7 +888,6 @@ Checkstyle rules are *disabled by default*. To add checkstyle to your project ju
                 <artifactId>spring-javaformat-maven-plugin</artifactId>
             </plugin>
             <plugin> <5>
-                <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-checkstyle-plugin</artifactId>
             </plugin>
         </plugins>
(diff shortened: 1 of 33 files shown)


adriancole:
fyi some folks have fixed some context related issues in brave.. I'm in a process of making a consistent version to weave in here at some point, maybe if lucky by friday. anyway this merged is good progress as there is already some performance work in the version of brave mentioned here

marcingrzejszczak:
If there are no breaking changes we could apply that change towards 2.2.x and merge it to master?

adriancole:
yeah sure can. this too. it isn't breaking changes. one thing concretely is that some people were accidentally making client spans child of client span due to callback nesting. this should be fixed and tested in the next version of brave (by backfilling tests to prove the correct parent is forced onto client callbacks)

PR #1532, opened 2020-01-22
Moves client instrumentation off deprecated Brave code

adriancole:
this will make things allocate less as less adapters are used. I will do the server ones tomorrow in a different PR

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/HttpClientBeanPostProcessor.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/HttpClientBeanPostProcessor.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.sleuth.instrument.web.client;
 
+import java.util.List;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.BiConsumer;
 import java.util.function.BiFunction;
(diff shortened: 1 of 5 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/HttpClientBeanPostProcessor.java
```diff
@@ -259,34 +234,68 @@ protected void handle(HttpClientResponse httpClientResponse,
...
+		final HttpClientResponse delegate;
+
+		WrappedHttpClientResponse(HttpClientResponse delegate) {
+			this.delegate = delegate;
 		}
 
 		@Override
-		public String requestHeader(HttpClientRequest request, String name) {
-			Object result = request.requestHeaders().get(name);
-			return result != null ? result.toString() : "";
```
adriancole: ps this was not good to return empty strings before. I'm not sure how that practice started...

marcingrzejszczak: I don't remember, maybe a guard against npe?

adriancole: brave always checks null, so maybe if someone was using this incorrectly perhaps. the impact of "" is that it will invoke a parser for no reason, adding overhead

PR #1533, opened 2020-01-22
Ilford updates

spencergibb:
Bumps versions to 3.0.0 and removes ribbon, zuul, hystrix and s-c-aws.

Not sure if tests pass (probably don't), but I'd want someone else to finish the work.


adriancole:
I can take this over.. I'll probably rebase etc. thanks for the start @spencergibb !

spencergibb:
Thanks! It compiles. Welcome back!

adriancole:
In ribbon version, `WebClientDiscoveryExceptionTests.testFeignInterfaceWithException()` raises `com.netflix.client.ClientException: Load balancer does not have available server for client: exceptionservice` Without ribbon, we get something slightly different: there's no registration for the service "exceptionservice", and feign actually returns a synthetic 503 instead of raising an exception.

In a separate PR, I think probably `TraceFeignBlockingLoadBalancerClient` should catch this 503 (even in the 2.2 line), so that when there's nothing registered there's at least a span suggesting what happened. We could either look at the 503 alone, or poke at response.request() to see if there were trace headers (there aren't when feign returns the synthetic response). we probably need a new test.

To make  `WebClientDiscoveryExceptionTests.testFeignInterfaceWithException()`, which doesn't have ribbon now, to fail.. somehow we need to register "exceptionservice" I guess, and assume that if it were and no servers were there, the impl would throw? That or maybe it shouldn't really throw, and that's just a legacy.. that 503 is fine?

cc @OlgaMaciaszek for commentary/advise as I'm not sure intent vs behavior here

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/TraceFeignBlockingLoadBalancerClient.java
```diff
@@ -80,10 +77,13 @@ public Response execute(Request request, Request.Options options) throws IOExcep
 			if (LOG.isDebugEnabled()) {
 				LOG.debug("Exception thrown", e);
 			}
-			if (e instanceof IOException || e.getCause() != null
-					&& e.getCause() instanceof ClientException
-					&& ((ClientException) e.getCause())
-							.getErrorType() == ClientException.ErrorType.GENERAL) {
+			if (e instanceof IOException/*
+										 * || e.getCause() != null && e.getCause()
```
OlgaMaciaszek: Why are we commenting this out?

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/annotation/SleuthSpanCreatorAspectFluxTests.java
```diff
@@ -27,7 +27,7 @@
 import brave.Tracer;
 import brave.propagation.TraceContext;
 import brave.sampler.Sampler;
-import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang.StringUtils;
```
OlgaMaciaszek: Why go back to an older version of the lib?

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/TraceRestTemplateInterceptorTests.java
```diff
@@ -29,7 +29,7 @@
 import brave.propagation.ThreadLocalCurrentTraceContext;
 import brave.sampler.Sampler;
 import brave.spring.web.TracingClientHttpRequestInterceptor;
-import org.apache.commons.lang3.StringUtils;
+import org.apache.commons.lang.StringUtils;
```
OlgaMaciaszek: Why change from `lang` to `lang3`?

tests/spring-cloud-sleuth-instrumentation-feign-tests/src/test/java/org/springframework/cloud/sleuth/instrument/feign/issues/issue1125/ManuallyCreatedLoadBalancerFeignClientTests.java
```diff
@@ -91,7 +84,7 @@ public void should_reuse_custom_feign_client() {
 	@Test
 	public void my_client_called() {
 		this.myNameRemote.get();
-		then(this.myClient.wasCalled()).isTrue();
+		// then(this.myClient.wasCalled()).isTrue();
```
OlgaMaciaszek: Maybe add TODOs to add a client  verification?

tests/spring-cloud-sleuth-instrumentation-feign-tests/src/test/java/org/springframework/cloud/sleuth/instrument/feign/issues/issue1125delegates/ManuallyCreatedDelegateLoadBalancerFeignClientTests.java
```diff
@@ -89,10 +82,10 @@ public void open() {
 	public void should_reuse_custom_feign_client() {
 		String response = this.myNameRemote.get();
 
-		then(this.myClient.wasCalled()).isTrue();
+		// then(this.myClient.wasCalled()).isTrue();
```
OlgaMaciaszek: Maybe add TODOs to add client verification?

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/TraceRestTemplateInterceptorTests.java
spencergibb: I think it was a transitive dependency. I don't think we should be using commons lang at all.

adriancole:
taking self off while I focus on reactor. I don't want to block others in case they want to chip in. we are pro non-blocking

marcingrzejszczak:
I'm taking this over

marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-sleuth/commit/3813cf9dd47f98db0e075412abbffbd9d6742974

PR #1541, opened 2020-01-31
Refactors some of the reactor code

adriancole:
This starts a process to understand what the reactor code is doing and
why.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/LazySpanSubscriber.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/LazySpanSubscriber.java
@@ -16,7 +16,6 @@
 
 package org.springframework.cloud.sleuth.instrument.reactor;
 
-import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Supplier;
 
 import org.reactivestreams.Subscription;
@@ -29,7 +28,7 @@ import reactor.util.context.Context;
  * @author Marcin Grzejszczak
  * @since 2.0.0
  */
-final class LazySpanSubscriber<T> extends AtomicBoolean implements SpanSubscription<T> {
+final class LazySpanSubscriber<T> implements SpanSubscription<T> {
 
 	private final Supplier<SpanSubscription<T>> supplier;
 
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/LazySpanSubscriber.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/LazySpanSubscriber.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.sleuth.instrument.reactor;
 
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.function.Supplier;
 
 import org.reactivestreams.Subscription;
@@ -28,7 +29,8 @@ import reactor.util.context.Context;
  * @author Marcin Grzejszczak
  * @since 2.0.0
  */
-final class LazySpanSubscriber<T> implements SpanSubscription<T> {
+// TODO: why are we extending AtomicBoolean and not actually using its methods?
+final class LazySpanSubscriber<T> extends AtomicBoolean implements SpanSubscription<T> {
 
 	private final Supplier<SpanSubscription<T>> supplier;
 
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
@@ -18,7 +18,6 @@ package org.springframework.cloud.sleuth.instrument.reactor;
 
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
-import java.util.function.BooleanSupplier;
 import java.util.function.Function;
 
 import brave.Tracing;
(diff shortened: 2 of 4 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/LazySpanSubscriber.java
```diff
@@ -29,7 +28,7 @@
  * @author Marcin Grzejszczak
  * @since 2.0.0
  */
-final class LazySpanSubscriber<T> extends AtomicBoolean implements SpanSubscription<T> {
+final class LazySpanSubscriber<T> implements SpanSubscription<T> {
```
marcingrzejszczak: That was a suggestion by @smaldini and this is how the reactor team does it cc @simonbasle

adriancole: ok.. maybe we can add a comment once we figure out why it helps

adriancole: I think some files use compareAndSet, just this one doesn't so the AtomicBoolean is a distraction. that's why I removed it.

I guess if we are supposed to do some only once callbacks, we should use the atomic methods ya?

adriancole: I don't want to be blocked on this as there's more important things. I replaced with a TODO to explain wtf.

adriancole:
took out the questionable thing. going to merge so that the next refactor is readable

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
```diff
@@ -67,43 +66,37 @@ private ReactorSleuth() {
...
+			// production of values in a trace context. This prevents excessive overhead
+			// when using constant results such as Flux/Mono #just, #empty, #error
 			if (p instanceof Fuseable.ScalarCallable) {
 				return sub;
 			}
-			Scannable scannable = Scannable.from(p);
-			// rest of the logic unchanged...
-			if (isActive.getAsBoolean()) {
+
+			if (beanFactory instanceof ConfigurableApplicationContext
```
robotmrv: @adriancole   
before change `instanceof` was calculated only once  
now it is calculated on each `Publisher.subscribe()` and it could be a lot of times  
As far as I understand reactor team trying to avoid "expensive instanceof check" see reactor/reactor-core#1882

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/LazySpanSubscriber.java
robotmrv: as far as I understand, this pattern reduces allocation of extra field in case when subscriber has `boolean` state  
and if you do not use `AtomicBoolean` methods it does not make sense to apply it.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
adriancole: how is coersing a lambda beanFactory::isActive to boolean supplier memoizing?

adriancole: oh I see it is specifically the instanceof you are trying to cache.. anyway the code is a bit messy and this moved to a place where it is cached anyway. I'll ping you on the next PR

adriancole: actually the logic is a bit strange to me, I guess I messed up because when not an instanceof ConfigurableApplicationContext it should return true (not sure when spring boot will be used without a ConfigurableApplicationContext, though..)

adriancole: I'll revert the logic error, but I think the instanceof part is suspicious. the pull request @bsideup did changed things besides instanceof and focusing on indexof check where there are many places in this code that call getBean still.. seems to draw attention to something that suggests everything has similar attention. anyway will revert this and hope we get to a point where instanceof is limiting us

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/LazySpanSubscriber.java
robotmrv: considering allocations  
maybe it reasonable to merge `SpanSubscriptionProvider` and `LazySpanSubscriber` so you do not allocate extra object?

adriancole:
agreed I'm trying now to see if we even need the extra type, if we rejig
things slightly

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
```diff
@@ -67,43 +66,37 @@ private ReactorSleuth() {
 			log.trace("Scope passing operator [" + beanFactory + "]");
 		}
 
-		// Adapt if lazy bean factory
-		BooleanSupplier isActive = beanFactory instanceof ConfigurableApplicationContext
-				? ((ConfigurableApplicationContext) beanFactory)::isActive : () -> true;
-
 		return Operators.liftPublisher((p, sub) -> {
-			// if Flux/Mono #just, #empty, #error
+			// While supply of scalar types may be deferred, we don't currently scope
```
simonbasle: Side note: the contract expressed in the `ScalarCallable` javadoc should be made more explicit, but yeah a `ScalarCallable` should resolve immediately without latency, to a terminating `Publisher`.

There's not a lot of them in core: just, empty, error, and errorSupplied. The only outlier is `errorSupplied`, since the Throwable is not pre-computed but lazily generated. But that's still considered an instant termination for all intent and purposes.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/LazySpanSubscriber.java
```diff
@@ -29,6 +29,7 @@
  * @author Marcin Grzejszczak
  * @since 2.0.0
  */
+// TODO: why are we extending AtomicBoolean and not actually using its methods?
```
simonbasle: Indeed, if the CAS methods are not used then extending `AtomicBoolean` is not useful. Have you looked at `LazySpanSubscriber` instances being passed around / used as `AtomicBoolean`s maybe?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
adriancole: thanks.. updating in the next PR!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/LazySpanSubscriber.java
adriancole: we think this was copy/paste error and nothing more. good idea (checked.)

adriancole:
much larger follow-up here https://github.com/spring-cloud/spring-cloud-sleuth/pull/1542

PR #1542, opened 2020-01-31
Extracts LazyBean and backports reactor code to use it

adriancole:
There are numerous places where code defensively guards access to
BeanFactory methods. This centralizes the code, first using in reactor
as that's the more performance sensitive.

This also weaves in feedback from #1541 given by @simonbasle and @robotmrv

More unit tests are needed, but raised to ensure the approach is fine.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
@@ -16,8 +16,6 @@
 
 package org.springframework.cloud.sleuth.instrument.reactor;
 
-import java.util.Map;
-import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Function;
 
 import brave.Tracing;
@@ -70,10 +68,14 @@ public abstract class ReactorSleuth {
 			log.trace("Scope passing operator [" + springContext + "]");
 		}
 
+		// keep a reference outside the lambda so that any caching will be visible to
+		// all publishers
+		LazyBean<CurrentTraceContext> lazyCurrentTraceContext = new LazyBean<>(
+				springContext, CurrentTraceContext.class);
+
 		return Operators.liftPublisher((p, sub) -> {
-			// While supply of scalar types may be deferred, we don't currently scope
-			// production of values in a trace context. This prevents excessive overhead
-			// when using constant results such as Flux/Mono #just, #empty, #error
+			// We don't scope scalar results as they happen in an instant. This prevents
+			// excessive overhead when using Flux/Mono #just, #empty, #error, etc.
 			if (p instanceof Fuseable.ScalarCallable) {
 				return sub;
 			}
@@ -88,13 +90,36 @@ public abstract class ReactorSleuth {
 				return sub;
 			}
 
+			Context context = sub.currentContext();
+
 			if (log.isTraceEnabled()) {
-				log.trace("Spring Context [" + springContext
-						+ "] Creating a scope passing span subscriber with Reactor Context "
-						+ "[" + sub.currentContext() + "] and name [" + name(sub) + "]");
+				log.trace("Spring context [" + springContext + "], Reactor context ["
+						+ context + "], name [" + name(sub) + "]");
 			}
 
-			return scopePassingSpanSubscription(springContext, sub);
+			// Try to get the current trace context bean, lenient when there are problems
+			CurrentTraceContext currentTraceContext = lazyCurrentTraceContext.get();
+			if (currentTraceContext == null) {
+				if (log.isTraceEnabled()) {
+					log.trace("Spring Context [" + springContext
+							+ "] did not return a CurrentTraceContext. Reactor Context is ["
+							+ sub.currentContext() + "] and name is [" + name(sub) + "]");
+				}
+				assert false; // should never happen, but don't break.
+				return sub;
+			}
+
+			TraceContext parent = traceContext(context, currentTraceContext);
+			if (parent == null) {
+				return sub; // no need to scope a null parent
+			}
+
+			if (log.isTraceEnabled()) {
+				log.trace("Creating a scope passing span subscriber with Reactor Context "
+						+ "[" + context + "] and name [" + name(sub) + "]");
+			}
+			return new ScopePassingSpanSubscriber<>(sub, context, currentTraceContext,
+					parent);
 		});
 	}
 
@@ -102,25 +127,14 @@ public abstract class ReactorSleuth {
 		return Scannable.from(sub).name();
 	}
 
-	private static Map<ConfigurableApplicationContext, CurrentTraceContext> CACHE = new ConcurrentHashMap<>();
-
-	static <T> CoreSubscriber<? super T> scopePassingSpanSubscription(
-			ConfigurableApplicationContext springContext, CoreSubscriber<? super T> sub) {
-		CurrentTraceContext currentTraceContext = CACHE.computeIfAbsent(springContext,
-				springContext1 -> springContext1.getBean(CurrentTraceContext.class));
-		Context context = sub.currentContext();
-
-		TraceContext parent = context.getOrDefault(TraceContext.class, null);
-		if (parent == null) {
-			parent = currentTraceContext.get();
+	/**
+	 * Like {@link CurrentTraceContext#get()}, except it first checks the reactor context.
+	 */
+	static TraceContext traceContext(Context context, CurrentTraceContext fallback) {
+		if (context.hasKey(TraceContext.class)) {
+			return context.get(TraceContext.class);
 		}
-		if (parent != null) {
-			return new ScopePassingSpanSubscriber<>(sub, context, currentTraceContext,
-					parent);
-		}
-		else {
-			return sub; // no need to trace
-		}
+		return fallback.get();
 	}
 
 }
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/LazyBean.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/LazyBean.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2013-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.sleuth.instrument.reactor;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.lang.Nullable;
+
+/**
+ * Avoids calling the expensive {@link ConfigurableApplicationContext#getBean(Class)} many
+ * times or throwing an exception.
+ */
+final class LazyBean<T> {
+
+	// spring-jcl uses commons-logging, so do we.
+	private static final Log log = LogFactory.getLog(LazyBean.class);
+
+	final ConfigurableApplicationContext springContext;
+
+	final Class<T> requiredType;
+
+	T value;
+
+	LazyBean(ConfigurableApplicationContext springContext, Class<T> requiredType) {
+		this.springContext = springContext;
+		this.requiredType = requiredType;
+	}
+
+	/**
+	 * Attempts to provision from the underlying bean factory, if not already provisioned.
+	 * @return the bean value or null if there was an exception getting it.
+	 */
+	@Nullable
+	T get() {
+		if (this.value != null) {
+			return this.value;
+		}
+
+		try {
+			this.value = springContext.getBean(requiredType);
+		}
+		catch (Exception ex) {
+			if (log.isDebugEnabled()) {
+				log.debug("Spring context [" + springContext + "] error getting ["
+						+ requiredType + "].", ex);
+			}
+		}
+		return this.value;
+	}
+
+}
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingSpanSubscriber.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingSpanSubscriber.java
@@ -54,8 +54,7 @@ final class ScopePassingSpanSubscriber<T> implements SpanSubscription<T>, Scanna
 		this.subscriber = subscriber;
 		this.currentTraceContext = currentTraceContext;
 		this.parent = parent;
-		this.context = ctx != null && parent != null ? ctx.put(TraceContext.class, parent)
-				: ctx != null ? ctx : Context.empty();
+		this.context = parent != null ? ctx.put(TraceContext.class, parent) : ctx;
 		if (log.isTraceEnabled()) {
 			log.trace("Parent span [" + parent + "], context [" + this.context + "]");
 		}
@@ -81,7 +80,6 @@ final class ScopePassingSpanSubscriber<T> implements SpanSubscription<T>, Scanna
 		try (Scope scope = this.currentTraceContext.maybeScope(this.parent)) {
 			this.s.cancel();
 		}
-
 	}
 
 	@Override
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/SpanSubscriber.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/SpanSubscriber.java
@@ -109,7 +109,7 @@ final class SpanSubscriber<T> extends AtomicBoolean implements SpanSubscription<
 			// no additional cleaning is required cause we operate on scopes
 			if (log.isTraceEnabled()) {
 				log.trace("Request after cleaning. Current span [{}]",
-						this.currentTraceContext.get());
+						this.span.context());
 			}
 		}
 	}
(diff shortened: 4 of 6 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
```diff
@@ -59,65 +57,69 @@ private ReactorSleuth() {
...
-			return new ScopePassingSpanSubscriber<>(sub, context, currentTraceContext,
-					parent);
-		}
-		else {
-			return sub; // no need to trace
+	/**
+	 * Like {@link CurrentTraceContext#get()}, except it first checks the reactor context.
+	 */
+	// TODO: context is nullable likely for tests, but this is unverified
+	static TraceContext traceContext(@Nullable Context context,
```
adriancole: Will we ever expect `subscription.currentContext()` to return null? If not, I really want to clean this up..

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
```diff
@@ -59,65 +57,69 @@ private ReactorSleuth() {
...
 		if (log.isTraceEnabled()) {
 			log.trace("Scope passing operator [" + beanFactory + "]");
 		}
 
+		// keep a reference outside the lambda so that any caching will be visible to
+		// all publishers
+		LazyCurrentTraceContext lazyCurrentTraceContext = new LazyCurrentTraceContext(
+				beanFactory);
+
 		return Operators.liftPublisher((p, sub) -> {
```
adriancole: note that the main difference between this and normal spring stuff, is usually there's an object to catch the reference of the bean we want to cache. since this is a lambda, there's no stash. I've extracted the stashing logic into `LazyCurrentTraceContext` although most of it is defensive ceremony around bean factories. the main advantage is we can pass the same type whether or not it is lazy.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/LazyCurrentTraceContext.java
```diff
@@ -0,0 +1,126 @@
...
+		}
+		else if (log.isTraceEnabled()) {
+			log.trace("Spring context [" + beanFactory + "] not yet refreshed: "
+					+ "Will use NoOp CurrentTraceContext.");
+		}
+		return this.currentTraceContext;
+	}
+
+	boolean isActive() {
+		// TODO: when if ever would boot provision something besides
```
adriancole: this still baffles me.. if anyone knows the answer

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/LazyCurrentTraceContext.java
```diff
@@ -0,0 +1,126 @@
...
+	final boolean instanceofConfigurableApplicationContext;
+
+	CurrentTraceContext currentTraceContext;
+
+	LazyCurrentTraceContext(BeanFactory beanFactory) {
+		this.beanFactory = beanFactory;
+		// There is an assumption that instanceof is limiting performance, though not
+		// verified here.
+		// https://github.com/reactor/reactor-core/pull/1882
+		this.instanceofConfigurableApplicationContext = beanFactory instanceof ConfigurableApplicationContext;
```
adriancole: I kept this special case though I think it is over optimization. I don't think we've done enough work on this code to start scrubbing like this.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/LazyScopePassingSpanSubscriber.java
```diff
@@ -29,48 +29,67 @@
...
-		return this.supplier.get().currentContext();
+		return spanSubscription().currentContext();
+	}
+
+	SpanSubscription<T> spanSubscription() {
+		if (this.spanSubscription == null) {
+			TraceContext parent = ReactorSleuth.traceContext(this.context,
+					this.currentTraceContext);
+
+			this.spanSubscription = new ScopePassingSpanSubscriber<>(this.subscriber,
```
adriancole: TODO: we can doublecheck lazycurrentTraceContext.unwrap here to straight pipe future operations

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/LazyScopePassingSpanSubscriber.java
```diff
@@ -29,48 +29,67 @@
  * @author Marcin Grzejszczak
  * @since 2.0.0
  */
-// TODO: why are we extending AtomicBoolean and not actually using its methods?
-final class LazySpanSubscriber<T> extends AtomicBoolean implements SpanSubscription<T> {
+final class LazyScopePassingSpanSubscriber<T> implements SpanSubscription<T> {
```
robotmrv: interface name is `SpanSubscription`  implementation name is `LazyScopePassingSpanSubscriber` looks strange.  
`SpanSubscription`  is both `Subscription` and `CoreSubscriber` probably better name would be `LazyScopePassingSpanOperator`

adriancole:
agree better name thanks

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/LazyScopePassingSpanSubscriber.java
robotmrv: looked at reactor-core    
`Subscription` + `CoreSubscriber` called `InnerOperator` but its implementations most of the time called `*Subscriber` so it is ok   
interface name is confusing

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/LazyCurrentTraceContext.java
```diff
@@ -0,0 +1,126 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.sleuth.instrument.reactor;
+
+import brave.propagation.CurrentTraceContext;
+import brave.propagation.TraceContext;
+import org.apache.commons.logging.Log;
```
adriancole: ps what's the latest with logging libraries? I moved this from the other file, but not sure what we should be using. As it surprises people I might make a comment above the imports.

spencergibb: Spring still uses commons logging https://github.com/spring-projects/spring-framework/tree/master/spring-jcl/src/main/java/org/apache/commons/logging

adriancole: thanks. I asked someone to review and they stopped surprised about this..  now I know how to answer.  cheers!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/LazyScopePassingSpanSubscriber.java
anuraaga: Maybe worth a comment that this object is safe to allocate multiple times so we don't worry about thread safety (at least assuming that's correct ;)

adriancole: I don't know if it is, the class was formerly allocated each time :D maybe someone else can chime in on this one.

adriancole: there's no obvious state concern, I guess I mean I'm unaware of a subtle one.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/LazyScopePassingSpanSubscriber.java
adriancole: roger. will leave it for now. thx

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/LazyScopePassingSpanSubscriber.java
adriancole: actually I think it is less distracting to assume this should be provisioned only once.. like a normal lazy. I tried hunting down the various interfaces and there could be different threads invoking callbacks as far as I can tell.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/LazyCurrentTraceContext.java
robotmrv: looks like you are right, it is over optimization as it is not a hot path

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
robotmrv: reactor-core uses `@NonNullApi` on a package level 
https://github.com/reactor/reactor-core/blob/master/reactor-core/src/main/java/reactor/core/package-info.java#L22-L23
``` 
@NonNullApi
package reactor.core;
```

and `CoreSubscriber#currentContext` does not have `@Nullable` annotation on return type 
https://github.com/reactor/reactor-core/blob/master/reactor-core/src/main/java/reactor/core/CoreSubscriber.java#L44  
```java
	default Context currentContext(){
		return Context.empty();
	}
```
So it must be safe assumption that context should not be null

adriancole:
@robotmrv you're a real champ. thanks for all the research!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/LazyCurrentTraceContext.java
```diff
@@ -0,0 +1,119 @@
...
+		}
+		else if (log.isTraceEnabled()) {
+			log.trace("Spring context [" + beanFactory + "] not yet refreshed: "
+					+ "Will use NoOp CurrentTraceContext.");
+		}
+		return this.currentTraceContext;
+	}
+
+	boolean isActive() {
+		// TODO: Will Spring Boot ever return something besides ConfigurableApplicationContext?
```
adriancole: @wilkinsona @philwebb can you answer this one for me? I'd like to erase this question not just here, but everywhere.. with an assertion and blind cast, ideally, or a comment if it is actually expected.

wilkinsona: I don't think a `BeanFactory` will ever be a `ConfigurableApplicationContext`, at least not with Framework's various implementations of those two interfaces. Perhaps you could inject the context rather than the bean factory?

Much of Boot assumes that the context will be a `ConfigurableApplicationContext` (rooted in SpringApplication's requirement for it to be so). It is possible to use some pieces of Boot with a context that isn't a `ConfigurableApplicationContext`, but I would be surprised if many if any people do so. That said we do still have some `instanceof` checks in Boot's code. Some of those deal with parent contexts where you can be less sure of the type and I don't think you could reasonably assume it will be a `ConfigurableApplicationContext`.

adriancole: it looks like in sleuth a BeanDefinitionRegistryPostProcessor injects ConfigurableApplicationContext and passes it here as a BeanFactory https://github.com/spring-cloud/spring-cloud-sleuth/blob/76b7563feb58bbf31104a5c7dae080dcea469c06/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/TraceReactorAutoConfiguration.java#L141

However, this one injects a BeanFactory.. https://github.com/spring-cloud/spring-cloud-sleuth/blob/d08e106100c1e9957926517394097047a6884ea1/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java#L149

I'll try to see if it will still work if all things just ask for the `ConfigurableApplicationContext`. While it will be a signature change, I'm not sure who would be using this utility besides us.. at worst, I guess we can do instanceof and overload it.

adriancole: since we injected both the narrowed type, ConfigurableApplicationContext, and also the wider type BeanFactory, it seems safe to assume we can narrow this to ConfigurableApplicationContext and still support what we did before. Since this is simpler, I'll do that in the next commit. Folks who want to deal with BF edge cases will want to ask about that. I think this was BF solely due to mocking anyway.

much appreciated background, @wilkinsona, I've copied the relevant bits into a comment.

adriancole:
when SpanSubscriber was deprecated, code saying it was testing SpanSubscriber didn't actually touch it anymore. I fixed that to focus on what is executed, which is purely scope related and not span lifecycle

tests/spring-cloud-sleuth-instrumentation-reactor-tests/src/test/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingSpanSubscriberSpringBootTests.java
```diff
@@ -0,0 +1,204 @@
...
+import static org.assertj.core.api.BDDAssertions.then;
+
+/**
+ * Like {@link ScopePassingSpanSubscriberTests}, except this tests wiring with spring boot
+ * config.
+ */
+@RunWith(SpringRunner.class)
+@SpringBootTest(classes = ScopePassingSpanSubscriberSpringBootTests.Config.class,
+		webEnvironment = SpringBootTest.WebEnvironment.NONE)
+public class ScopePassingSpanSubscriberSpringBootTests {
```
adriancole: This was formerly named SpanSubscriberTests, and it used the tracer. The actual code coverage showed it only touched ScopePassingSpanSubscriber, and that the tracer code was a distraction as the only thing done is related to scoping of spans, not start/stop of them.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
```diff
@@ -55,69 +52,77 @@ private ReactorSleuth() {
...
 			}
 			if (log.isTraceEnabled()) {
-				log.trace("Spring Context [" + beanFactory
-						+ "] is not yet refreshed, falling back to lazy span subscriber. Reactor Context is ["
+				log.trace("Falling back to lazy span subscriber. Reactor Context is ["
 						+ sub.currentContext() + "] and name is [" + name(sub) + "]");
 			}
-			return new LazySpanSubscriber<>(
-					new SpanSubscriptionProvider<>(beanFactory, sub));
+			return new LazyScopePassingSpanSubscriber<>(sub, context,
```
adriancole: now that we no longer inject beanfactory, we never get here. the code is dead running all the tests as far as I can tell with breakpoints.. even in the case of spring-cloud-context. because the ConfigurableApplicationContext is never not active anymore. It could be a massive simplification to remove the if statement and this special case...

adriancole:
pulled the spring context related code into a separate PR so that it can be applied to other branches. I verified nothing on github uses the `ReactorSleuth` type except sleuth itself.

adriancole:
https://github.com/spring-cloud/spring-cloud-sleuth/pull/1545 was the PR I was talking about

adriancole:
ok this code is a lot smaller now. Mostly, we now know we won't get into a spot where we have a non-refreshed context in the middle of a lift. even if we did, we'll just not trace that specific flow vs have a lot of expensive code around it.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
```diff
@@ -88,39 +90,51 @@ private ReactorSleuth() {
...
 
 	static String name(CoreSubscriber<?> sub) {
 		return Scannable.from(sub).name();
 	}
 
-	private static Map<ConfigurableApplicationContext, CurrentTraceContext> CACHE = new ConcurrentHashMap<>();
-
-	static <T> CoreSubscriber<? super T> scopePassingSpanSubscription(
-			ConfigurableApplicationContext springContext, CoreSubscriber<? super T> sub) {
-		CurrentTraceContext currentTraceContext = CACHE.computeIfAbsent(springContext,
```
adriancole: @robotmrv so the lazybean thing eliminates the need for this cache as far as I can tell.

adriancole:
Thanks for all the feedback, folks. I'll ping this PR with the next one.. there will be a few more on the topic of reactor refactor.

adriancole:
PS my aim is to document all the top-level types and why they are used. then isolate the more important code issues are raised about (ex FlatMapTests) in a JMH benchmark that doesn't use spring boot (more isolated). Along with the existing webflux benchmarks, we should be able to tell if we've progressed materially or if this has all been idle refactoring :)

adriancole:
https://github.com/spring-cloud/spring-cloud-sleuth/pull/1546

adriancole:
https://github.com/spring-cloud/spring-cloud-sleuth/pull/1548

adriancole:
https://github.com/spring-cloud/spring-cloud-sleuth/pull/1551 < reducing redundant getBean when using reactor-netty http client

adriancole:
#1553 < avoids double-wrapping: @simonbasle or @robotmrv would appreciate a look on this one..

adriancole:
#1554 < HttpClient bugs meet torchlight

PR #1545, opened 2020-02-03
Uses ConfigurableApplicationContext consistently in Reactor

adriancole:
Before, we used ConfigurableApplicationContext or BeanFactory eventhough
we already had a reference to ConfigurableApplicationContext. This uses
the latter consistently, avoiding a state condition that caused more
code.

This also corrects some misnamed tests and adjusts them to verify only
what they are responsible for.


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
```diff
@@ -55,15 +54,20 @@ private ReactorSleuth() {
...
 	 * @return a new lazy span operator pointcut
 	 */
-	@SuppressWarnings("unchecked")
+	// Much of Boot assumes that the Spring context will be a
+	// ConfigurableApplicationContext, rooted in SpringApplication's
+	// requirement for it to be so. Previous versions of Reactor
+	// instrumentation injected both BeanFactory and also
+	// ConfigurableApplicationContext. This chooses the more narrow
+	// signature as it is simpler than explaining instanceof checks.
 	public static <T> Function<? super Publisher<T>, ? extends Publisher<T>> scopePassingSpanOperator(
```
marcingrzejszczak: That's a breaking change. Maybe we should deprecate this one and create a new one?

adriancole: I looked on github, nothing is using it except us.. deprecating would cause a lot of debt

adriancole: ex https://github.com/search?q=reactorsleuth&type=Code only copies of sleuth

marcingrzejszczak: :+1:

adriancole:
@simonbasle I cannot reproduce the build break locally, but circleci is failing like this https://circleci.com/gh/spring-cloud/spring-cloud-sleuth/5816

> java.lang.UnsupportedOperationException: Although QueueSubscription extends Queue it is purely internal and only guarantees support for poll/clear/size/isEmpty. Instances shouldn't be used/exposed as Queue outside of Reactor operators.
	at org.springframework.cloud.sleuth.instrument.reactor.ScopePassingSpanSubscriberTests.should_not_trace_scalar_flows(ScopePassingSpanSubscriberTests.java:144)

That line is `transformer.apply(Mono.just(1)).subscribe(assertNoSpanSubscriber);`

I noticed that the `SpanSubscription` was retrofitted to be a `QueueSubscription` as a part of https://github.com/spring-cloud/spring-cloud-sleuth/pull/1008

As I'm not seeing the stack trace, only the pre-canned reactor exception message.. I'm wondering if you have hints either what could be doing this, or how to see the actual exception.

adriancole:
lulz cc @bsideup I think assertj is being clever, and it detects that the subscription is a collection as it extends Queue, then invokes iterator in its toSTring, which reactor forbids.. will look carefully next https://github.com/joel-costigliola/assertj-core/blob/master/src/main/java/org/assertj/core/presentation/StandardRepresentation.java#L189

still odd I can't reproduce locally..

simonbasle:
Yes that is typically the sort of reason this happens in tests.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
```diff
@@ -74,37 +78,36 @@ private ReactorSleuth() {
...
-	private static Map<BeanFactory, CurrentTraceContext> CACHE = new ConcurrentHashMap<>();
+	private static Map<ConfigurableApplicationContext, CurrentTraceContext> CACHE = new ConcurrentHashMap<>();
 
 	static <T> CoreSubscriber<? super T> scopePassingSpanSubscription(
-			BeanFactory beanFactory, CoreSubscriber<? super T> sub) {
-		CurrentTraceContext currentTraceContext = CACHE.computeIfAbsent(beanFactory,
-				beanFactory1 -> beanFactory1.getBean(CurrentTraceContext.class));
+			ConfigurableApplicationContext springContext, CoreSubscriber<? super T> sub) {
+		CurrentTraceContext currentTraceContext = CACHE.computeIfAbsent(springContext,
+				springContext1 -> springContext1.getBean(CurrentTraceContext.class));
```
robotmrv: `springContext1 -> springContext1.getBean(CurrentTraceContext.class)` could be extracted to the field to reduce allocations (it is created on each assembly or on each last operator)

adriancole: thanks, this is killed in the other pull request.

adriancole:
just squashed and force pushed to make sure the build pass happens more than once

simonbasle:
> lulz cc @bsideup I think assertj is being clever, and it detects that the subscription is a collection as it extends Queue, then invokes iterator in its toSTring, which reactor forbids.. 

We have had the case with logging frameworks before, so we protect against it explicitly in the `log()` operator (see [SignalLogger](https://github.com/reactor/reactor-core/blob/master/reactor-core/src/main/java/reactor/core/publisher/SignalLogger.java#L203))

adriancole:
thanks for the review and tips, @simonbasle and @robotmrv!

PR #1548, opened 2020-02-05
Fixes null bugs on WebClient canceled request

adriancole:
There are several places where we didn't check null correctly. This
addresses a couple of them.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java
@@ -18,6 +18,7 @@ package org.springframework.cloud.sleuth.instrument.web.client;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.concurrent.CancellationException;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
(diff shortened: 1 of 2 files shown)


violetagg:
> @violetagg do you know how to trigger `doOnRequestError`? 

Check these tests

https://github.com/reactor/reactor-netty/blob/master/src/test/java/reactor/netty/http/client/HttpClientTest.java#L1110

https://github.com/reactor/reactor-netty/blob/master/src/test/java/reactor/netty/http/client/HttpClientTest.java#L2012

adriancole:
@violetagg any ideas on last commit?

adriancole:
oh I see httpClient and webClient don't share this handler.. only httpClient does

adriancole:
I think I can proceed from here.. thanks for the hint @violetagg!

adriancole:
I was confused as multiple different instrumentation are being tested in a class named "WebClientTests" I'll separate these into separate files in the next commit, so that the next person as easily confused as me doesn't get confused.

adriancole:
moved the `doOnRequestError` failure to https://github.com/spring-cloud/spring-cloud-sleuth/pull/1552

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java
```diff
@@ -374,32 +383,31 @@ void terminateSpanOnCancel() {
...
 								+ this.span + "]");
 					}
-					handleReceive(this.span, clientResponse, throwable);
+					handleReceive(this.span, null, error);
 					return;
 				}
 				int statusCode = clientResponse.rawStatusCode();
-				boolean error = statusCode >= 400;
-				if (error) {
+				boolean isHttpError = statusCode >= 400;
```
adriancole: ps this isn't correct either, but I'm not changing this now. later we should redo most of this instrumentation

PR #1553, opened 2020-02-06
Reduces the amount of wrapping in reactive commands

adriancole:
This opts-out of wrapping when the reactor subscription is already
wrapped with the same trace context. I noticed this in some AspectJ
unit tests and webflux scenarios.

If I understand correctly, this should reduce the stack size, yet
keep the same propagation functionality as we have now.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
@@ -115,10 +115,19 @@ public abstract class ReactorSleuth {
 				return sub; // no need to scope a null parent
 			}
 
+			// Handle scenarios such as Mono.defer
+			if (sub instanceof ScopePassingSpanSubscriber) {
+				ScopePassingSpanSubscriber<?> scopePassing = (ScopePassingSpanSubscriber) sub;
+				if (scopePassing.parent.equals(parent)) {
+					return sub; // don't double-wrap
+				}
+			}
+
 			if (log.isTraceEnabled()) {
 				log.trace("Creating a scope passing span subscriber with Reactor Context "
 						+ "[" + context + "] and name [" + name(sub) + "]");
 			}
+
 			return new ScopePassingSpanSubscriber<>(sub, context, currentTraceContext,
 					parent);
 		});
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingSpanSubscriber.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingSpanSubscriber.java
@@ -37,17 +37,17 @@ import reactor.util.context.Context;
  */
 final class ScopePassingSpanSubscriber<T> implements SpanSubscription<T>, Scannable {
 
-	private static final Log log = LogFactory.getLog(ScopePassingSpanSubscriber.class);
+	static final Log log = LogFactory.getLog(ScopePassingSpanSubscriber.class);
 
-	private final Subscriber<? super T> subscriber;
+	final Subscriber<? super T> subscriber;
 
-	private final Context context;
+	final Context context;
 
-	private final CurrentTraceContext currentTraceContext;
+	final CurrentTraceContext currentTraceContext;
 
-	private final TraceContext parent;
+	final TraceContext parent;
 
-	private Subscription s;
+	Subscription s;
 
 	ScopePassingSpanSubscriber(Subscriber<? super T> subscriber, Context ctx,
 			CurrentTraceContext currentTraceContext, @Nullable TraceContext parent) {
--- a/tests/spring-cloud-sleuth-instrumentation-reactor-tests/src/test/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingSpanSubscriberTests.java
+++ b/tests/spring-cloud-sleuth-instrumentation-reactor-tests/src/test/java/org/springframework/cloud/sleuth/instrument/reactor/ScopePassingSpanSubscriberTests.java
@@ -166,6 +166,46 @@ public class ScopePassingSpanSubscriberTests {
 	}
 
 	@Test
+	public void should_not_double_wrap() {
+		springContext.registerBean(CurrentTraceContext.class, () -> currentTraceContext);
+		springContext.refresh();
+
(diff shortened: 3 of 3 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
```diff
@@ -115,10 +115,19 @@ private ReactorSleuth() {
 				return sub; // no need to scope a null parent
 			}
 
+			// Handle scenarios such as Mono.defer
+			if (sub instanceof ScopePassingSpanSubscriber) {
+				ScopePassingSpanSubscriber<?> scopePassing = (ScopePassingSpanSubscriber) sub;
+				if (scopePassing.parent.equals(parent)) {
+					return sub; // don't double-wrap
```
adriancole: the way I figured out defer did this, was I changed this line to throw an exception, then ran all tests. Defer ended up one of the patterns in the stack, though there are likely others.

robotmrv: nice idea.  
Unfortunately it works only if there is only one lift function in `onEachOperator` hook.

robotmrv:
I have another simple idea to reduce allocations.  
https://github.com/spring-cloud/spring-cloud-sleuth/pull/1553/files#diff-dd4ad4e07b03810c1ea9a914ba81f2f9R80-R82

```java
if (p instanceof Fuseable.ScalarCallable) {
	return p;
}
```
as far as publisher is known at assembly time this check could be moved outside of `Operators.liftPublisher()`   
like this  
```java
public static <T> Function<? super Publisher<T>, ? extends Publisher<T>> scopePassingSpanOperator(
			ConfigurableApplicationContext springContext) {
...
	Function<? super Publisher<T>, ? extends Publisher<T>> lift = Operators.liftPublisher((p, sub) -> {
	....
	});
	return p -> {
		if (p instanceof Fuseable.ScalarCallable) {
			return p;
		}
		return lift.apply(p);
	};
}
```
this will remove unnecessary `MonoLift` creation and all logic from `Operators.liftPublisher`

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/ReactorSleuth.java
simonbasle: to clarify, the optimization is only _guaranteed_ to work if there is only one lift function. having several sleuth lift functions chained together would also work, or if sleuth is the last lift function in the chain. this is good enough coverage to make it worth applying the optim, I guess

robotmrv:
If to go further  
it would be nice if `reactor-core` publishers have some kind of attribute to determine whether operator is synchronous. Like new scannable attribute `synchronous` or `async`  

It would allow to optimize for `onEachOperator` lift function to not create scope if it was already created for synchronous chain of operators  
for chains like this  
```java
Mono.just(1)
        .publishOn(Schedulers.immediate())//need to scope here (+)
        .flatMap(it -> Mono.just(it))//need to scope here (+)
        .filter(it -> true)//do not scope (-)
        .map(it -> it)//do not scope (-)
        .doOnNext(it -> {})//do not scope (-)
        .block();//need to scope last operator (+)
```
and lift function could look like this  
```java
Hooks.onEachOperator(pub -> {
    boolean isAlreadyScoped = false;
    boolean isSynchronous = true;
    Scannable currentPub = Scannable.from(pub);
    while (isSynchronous && !isAlreadyScoped && currentPub != null) {
        isAlreadyScoped = isSleuthScoped(currentPub);
        isSynchronous = isSynchronous(currentPub);
        currentPub = currentPub.scan(Scannable.Attr.PARENT);
    }
    if (isAlreadyScoped && isSynchronous) {
        return pub;
    }
    return lift.apply(pub);
});
```
`isSleuthScoped(currentPub)` could be implemented by adding to MonoLift/FluxLift... attribute like  `lifterName` .  

This changes could reduce allocations even more.  
But without changes in `reactor-core` it could not be done... 
@simonbasle  WDYT

adriancole:
still looking for an answer here.. meanwhile just a small thing I noticed which is that reactor context methods return a new instance even if the ref associated with a key is the same instance as prior. This code guards against that. https://github.com/spring-cloud/spring-cloud-sleuth/pull/1558

simonbasle:
this is an interesting idea... it would need to be taken with a grain of salt / made very clear the result only concerns one stage of the pipeline, so that a result of `SYNC` isn't interpretable as "the whole operator pipeline runs synchronously" just because the last step is a synchronous operator (like a `filter`)

robotmrv:
> it would need to be taken with a grain of salt / made very clear the result only concerns one stage of the pipeline, so that a result of `SYNC` isn't interpretable as "the whole operator pipeline runs synchronously" just because the last step is a synchronous operator (like a `filter`)

Yes, `SYNC` is the property of an operator not of the whole pipeline. This is why the whole chain of operators is checked for `SYNC`.  
@simonbasle , is there any chance that it could be added as new feature in `reactor-core`?  
Should It be submitted as a feature request?

simonbasle:
maybe all the other avenues of optimization can be explored before implementing this in core. this can certainly be discussed in a feature request in the meantime though...

robotmrv:
> maybe all the other avenues of optimization can be explored before implementing this in core. this can certainly be discussed in a feature request in the meantime though...

@simonbasle  @adriancole   
I've created the feature request https://github.com/reactor/reactor-core/issues/2058 in `reactor-core` for further discussion  

> maybe all the other avenues of optimization can be explored before implementing this in core

@simonbasle 
`ScopePassingSpanSubscriber` implements `Fuseable.QueueSubscription` but always returns `Fuseable.NONE` on `requestFusion()`.  
Does it mean that all `Fuseable` optimizations will not be applied for the whole operators chain in case of `onEachHook` enabled?  
If so, what was the intention of this? Just to simplify implementation?  
If `ScopePassingSpanSubscriber` could be `Fused` it could enhance performance

simonbasle:
it has to implement the `QueueSubscription` interface because operators that see a `Fuseable` source generally cast their `Subscription` to `QueueSubscription`.
However, for simplification, the `ScopePassingSpanSubscriber` doesn't really support fusion and opts out of it by always returning `NONE` when negotiating fusion. If the implementation can be made fully-fusion compatible, it could indeed help negate some performance hit.

marcingrzejszczak:
I'd close this cause with the 3.0.0 release we will have the `MANUAL` instrumentation type option where we will have almost no wrapping, That should speed things up ;)

PR #1554, opened 2020-02-07
ReactorNettyHttpClientBraveTests

adriancole:
During other change, I noticed a lot of bugs in the instrumentation of
reactor-netty `HttpClient`. This uses Brave's integration tests to show
what works and what doesn't.

### DIFF
--- a/spring-cloud-sleuth-core/pom.xml
+++ b/spring-cloud-sleuth-core/pom.xml
@@ -336,6 +336,11 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
+			<groupId>io.zipkin.brave</groupId>
+			<artifactId>brave-instrumentation-http-tests</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
 			<groupId>com.netflix.archaius</groupId>
 			<artifactId>archaius-core</artifactId>
 			<scope>test</scope>
--- a/spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/client/ReactorNettyHttpClientBraveTests.java
+++ b/spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/client/ReactorNettyHttpClientBraveTests.java
@@ -1,0 +1,174 @@
+/*
+ * Copyright 2013-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.sleuth.instrument.web.client;
+
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.atomic.AtomicReference;
+
+import brave.http.HttpTracing;
+import brave.test.http.ITHttpAsyncClient;
+import io.netty.channel.ChannelOption;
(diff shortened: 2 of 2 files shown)


adriancole:
rebased

adriancole:
It looks like further work is needed, possibly the lifecycle hooks aren't at the right place. For example, `.doOnRequest()` sees a POST as a GET.. maybe we have to parse later than this.. like `.doAfterRequest()`

cc @violettagg

adriancole:
@violetagg I noticed a few things that if cleared should allow us to fixup the client instrumentation.

doOnRequest doesn't have the final requestline so this interferes with parsing and sampling. We could do parsing late, like doAfterRequest, but sampling need to be able to see the requestline.

it seems the retry on connect failure isn't configurable, and disabling redirect follow-up still results in a redirect follow-up. We may need to handle multiple callbacks to onRequest and make a new span each attempt. However, the doOnRequest problem remains.

Also important is that the test framework assumes retries and redirect follow-up can be turned off. we should to be able to do this. It seems single-retry on TCP connect failure might be hard coded?

adriancole:
@violetagg @smaldini particularly this one from above.. we need some clarity on. Will we be able to expect the resolved request (ex including http method and path) on `doOnRequest()`? If not, we may need to do some serious hacking..

violetagg:
> @violetagg @smaldini particularly this one from above.. we need some clarity on. Will we be able to expect the resolved request (ex including http method and path) on `doOnRequest()`? If not, we may need to do some serious hacking..

Can you point in the code where you cannot obtain that information?
We provide `HttpClientRequest` - https://projectreactor.io/docs/netty/release/api/reactor/netty/http/client/HttpClient.html#doOnRequest-java.util.function.BiConsumer-
from that request you can get this information
https://projectreactor.io/docs/netty/release/api/reactor/netty/http/client/HttpClientRequest.html

adriancole:
It isn't lack of the object, it is that for example, the request returns
GET on a POST request. I verified the server receives a POST. the
onAfterRequest hook has the correct data, but the onRequest has incorrect.

violetagg:
> It isn't lack of the object, it is that for example, the request returns GET on a POST request. I verified the server receives a POST. the onAfterRequest hook has the correct data, but the onRequest has incorrect.
> […](#)

That's a problem. Can you create an issue?

violetagg:
> it seems the retry on connect failure isn't configurable

We will retry one ONLY in cases of of `Connection reset by peer`
https://github.com/reactor/reactor-netty/blob/master/src/main/java/reactor/netty/http/client/HttpClientConnect.java#L666

> , and disabling redirect follow-up still results in a redirect follow-up. 

Please create an issue about this. By default auto-redirect is not enable and it should NOT redirect.

adriancole:
I was wrong about the redirect problem: https://github.com/reactor/reactor-netty/pull/996

adriancole:
about opting out of connection fail retry: https://github.com/reactor/reactor-netty/issues/995

adriancole:
about parse failures, which are behind a few of the test fails here, not just POST https://github.com/reactor/reactor-netty/pull/994

adriancole:
more of the code problems in this repo are around context propagation and scoping. I'll look into these while the parsing stuff happens in reactor-netty

adriancole:
almost everything is done or nearly done here at the reactor-netty `HttpClient` abstraction. We will need a release of reactor-netty to fix a couple things.

adriancole:
I've not pushed a code update, but will soon

adriancole:
Besides the noted reactor stuff, scope propagation isn't integrated at HttpClient, yet. I'm trying to figure out how to do this now..

adriancole:
going to now check against WebClient to ensure nothing regressed there (due to missing tests)

adriancole:
WebClient is similarly messed up.. will need to do 2 things
* have `WebClient` opt-out except scope propagation when reactor instrumentation is present
* Fix `WebClient` instrumentation when reactor instrumentation is absent (it is messed up also)

adriancole:
There are numerous bugs with associated issues raised in reactor for investigation. Many are fixed with what's currently in master. I'm going to merge this as it helps identify what's left to do without constantly rebasing.

adriancole:
check to see if the build passes twice in a row. it always passes on clean cache, but this time is the real test

PR #1562, opened 2020-02-14
Fixes WebClient propagation errors and adds Brave Tests

adriancole:
In many places, the trace context of callbacks was accidentally set to
the client span, not the invocation context. I noticed a hack trying to
work around this. This code fixes all the problems around context. It
also removes some sporadic logging, which was only applied to a few
hooks.

Finally, this adds Brave tests which would have caught the problems
earlier. Notably, there is still more work to do as this will not help
with duplicate instrumentation, which is normal when reactor-netty is
the WebClient's HTTP connector.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java
@@ -16,18 +16,16 @@
 
 package org.springframework.cloud.sleuth.instrument.web.client;
 
-import java.util.Collections;
 import java.util.List;
 import java.util.concurrent.CancellationException;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
 import brave.Span;
-import brave.Tracer;
-import brave.Tracing;
 import brave.http.HttpClientHandler;
 import brave.http.HttpTracing;
-import brave.propagation.Propagation;
+import brave.propagation.CurrentTraceContext;
+import brave.propagation.CurrentTraceContext.Scope;
 import brave.propagation.TraceContext;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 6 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java
```diff
@@ -166,80 +139,55 @@ public static ExchangeFilterFunction create(
...
+	}
+
+	CurrentTraceContext currentTraceContext() {
+		if (this.currentTraceContext == null) {
+			this.currentTraceContext = httpTracing.get().tracing().currentTraceContext();
 		}
-		MonoWebClientTrace trace = new MonoWebClientTrace(next, wrapper.buildRequest(),
-				this, span);
-		// TODO: investigate why this commit leaks a scope:
-		// 8f5bcdabd7af23df443e771432eb85597f3b3076
```
adriancole: this fixed one part but not the others.. the proper fix is this PR

adriancole:
@simonbasle @robotmrv I noticed in #1126 we started handling both onNext and onComplete, to finish a span. The code that handles onComplete when onNext isn't called is dead code. At least in our tests it isn't invoked.

I would like to remove the onComplete without onNext edge case as it isn't tested and also it is very confusing. Is there any reason why we should be so defensive?

adriancole:
build fail is likely the ever annoying hook registration leak thing, best I can guess. Not repeatable locally of course.. I'll try to use the exact commands circle uses..

adriancole:
I used the same test command as circleci locally and passed..

adriancole:
always when a flake it passes when build with no cache..

simonbasle:
> @simonbasle @robotmrv I noticed in #1126 we started handling both onNext and onComplete, to finish a span. The code that handles onComplete when onNext isn't called is dead code. At least in our tests it isn't invoked.
> 
> I would like to remove the onComplete without onNext edge case as it isn't tested and also it is very confusing. Is there any reason why we should be so defensive?

When you look at the `Flux` or `Mono` level, not the `Subscriber` level, pipelines which end results appear empty are not that uncommon. Any reactive method that returns a `Mono<Void>` for instance...

That said, Sleuth works by wrapping `Subscriber`s, which are the steps in one activation of the pipeline. The cases where such a step is empty are probably fewer. For instance, the `Subscriber` generated by `ignoreElements()`, despite producing a `Flux` that appears empty to the end consumer, still sees `onNext()` calls from its source. It just doesn't propagate them. If there is another operator after that, its `Subscriber` will only see an `onComplete()`.

Another way of triggering this code path is by having an empty source:
```java
Flux.empty()
    .map(i -> i)
    .filter(i -> true)
    .flatMap(i -> Mono.error(new RuntimeException("won't be thrown because no onNext"));
```

Hope this helps improving the tests/understanding when `onComplete` might matter...

It is probably true that the vast majority of _`Subscriber`s_ will not `onComplete()` without an `onNext()`, but when one considers the whole sequence (ie. the `Flux` or `Mono`) this cannot be considered an edge case. All Mon

adriancole:
@marcingrzejszczak PS since these tests are junit 4 we may need the junit-vintage-engine when you update to 5?

adriancole:
Thanks @simonbasle moved your comment here so it isn't lost https://github.com/spring-cloud/spring-cloud-sleuth/issues/1570

PR #1571, opened 2020-02-24
Adds test to show impact of duplication instrumentation

adriancole:
This shows tracing both `WebClient` and `HttpClient`, which doesn't happen by default, luckily!

However, it could easily happen if someone follows our directions to customize the backend.

Ex. if the following separated the HttpClient bean from the connector one, it would lead to duplicate instrumentation:
https://github.com/spring-projects/spring-boot/blob/8190b8eafbc00258197e89376c92d081207688aa/spring-boot-project/spring-boot-docs/src/main/java/org/springframework/boot/docs/web/reactive/function/client/ReactorNettyClientCustomizationExample.java

follow-up to https://github.com/spring-cloud/spring-cloud-sleuth/pull/1554

### DIFF
--- a/tests/spring-cloud-sleuth-instrumentation-reactor-tests/src/test/java/org/springframework/cloud/sleuth/instrument/web/client/WebClientUsingInstrumentedReactorNettyHttpClientBraveTests.java
+++ b/tests/spring-cloud-sleuth-instrumentation-reactor-tests/src/test/java/org/springframework/cloud/sleuth/instrument/web/client/WebClientUsingInstrumentedReactorNettyHttpClientBraveTests.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2013-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.sleuth.instrument.web.client;
+
+import brave.http.HttpTracing;
+import reactor.netty.http.client.HttpClient;
+
+import org.springframework.beans.factory.config.BeanPostProcessor;
+import org.springframework.context.annotation.AnnotationConfigApplicationContext;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+import org.springframework.http.client.reactive.ClientHttpConnector;
+import org.springframework.http.client.reactive.ReactorClientHttpConnector;
+import org.springframework.web.reactive.function.client.WebClient;
+
+/**
+ * This runs Brave's integration tests with duplicate instrumentation. This is not a
+ * situation that happens by default. However, if people follow instructions to customize
+ * Reactor, it can.
+ *
+ * <p>
+ * Ex.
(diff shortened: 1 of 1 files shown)


adriancole:
@marcingrzejszczak @simonbasle @violettagg so the plan here is to share state between the webclient and the reactor client instrumentation.

I think before we were looking if headers were already set, but maybe there's a way to do better. I'm not sure though as it seems like reactor context might not be visible from WebClient abstraction.. Any ideas on some non-http header way to share state between WebClient and HttpClient?

adriancole:
PS let's not release sleuth 2.2.x again until this is fixed, unless emergency. cc @OlgaMaciaszek

adriancole:
easy would be to read the request attributes webclient can set somehow from reactor-netty httpclient abstraction..

spencergibb:
We have a release scheduled Monday

adriancole:
ok will sort this out one way or another by somebody's monday! thanks for
the heads up!

adriancole:
@violetagg @smaldini can one of you please cut reactor-netty by monday? the passive parsing bugs are already sorted out in master

adriancole:
I currently see no way to access attributes from `org.springframework.web.reactive.function.client.ClientRequest` at the reactor-netty abstraction in any way. Tabling that idea.

adriancole:
ps on 
> @violetagg @smaldini can one of you please cut reactor-netty by monday? the massive parsing bugs are already sorted out in master

seems reactor-netty was released, just sleuth is behind on spring boot dep. I'll fix that offline

adriancole:
Between reactor-netty's HttpClient and WebClient's view of the client span, both have the same parent. WebClient's span overwrites reactor-netty's HttpClient trace headers, and ends up on the wire. In other words, WebClient's span wins. That said two spans are still written to zipkin, and log sync is confused, so even i we know this... it is still bad!

Specifically, this overwriting happens during `org.springframework.http.client.reactive.AbstractClientHttpRequest.doCommit()` when `ReactorClientHttpRequest.applyHeaders()` is invoked.. This transfers the trace headers from the WebClient layer to the reactor-netty layer, nuking the headers already there.

adriancole:
The main problem here is that WebClient and HttpClient can be initialized statically, or via injector, or via an injected builder. In most cases, the user is free to swap the client connector used by WebClient and we wouldn't know which is in use until runtime, plus there's no accessor to tell which is in use. The other main issue is the opposite order of hydrating an http client request. Eventhough the WebClient abstraction is higher than HttpClient, the latter is built first then the former applied into it.

This creates a very tricky situation which is compounded by the fact that attributes are not shareable between the abstractions:

Ex. 

`org.springframework.web.reactive.function.client.ClientRequest` has attributes

`org.springframework.http.client.reactive.ClientHttpRequest` does not

`reactor.netty.http.client.HttpClientRequest` also does not

adriancole:
My current thinking is to not fight the order. WebClient applies its request after reactory-netty HttpClient creates the underlying on. The only way out I can tell is to have WebClient be very lazy in its instrumentation. That or playing with synthetic headers..

Maybe, defer `WebClient` tracing until or later.. `ClientRequest.writeTo()`, checking `ClientHttpRequest` based on some value created with a custom ClientHttpConnector perhaps..

adriancole:
revised https://github.com/spring-cloud/spring-cloud-sleuth/pull/1571#issuecomment-593188223 as I had to precedence wrong

adriancole:
ok I'm officially taking this off the critical path :) I verified that we are saved by the bell, because we statically instantiate reactor http client when used in WebClient

https://github.com/spring-projects/spring-boot/blob/0f567c879d9a710ac2e4068b57c88335daeaf460/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/function/client/ClientHttpConnectorConfiguration.java#L51-L61

To fix this problem properly, we really need ReactorClientHttpConnector adjusted so that webclient attributes are sharable. For example, they could be re-exported in the reactor context, or the original webclient request could be set as a context property, or some other way.

If an attribute were sharable then we could opt-out of tracing at the reactor layer seeing it upstream, or visa versa. without this, there are not many good options left, as there's no thread locality and the webclient request goes through 2 transforms before being invoked.

Until then, it is way too hairy to fix this, at least in the ways I've become aware of in the last several days.

adriancole:
closing this for now until someone asks for it. the default config won't do this

Issue #1570, opened 2020-02-24
Backfill test about reactor edge case


adriancole:
From https://github.com/spring-cloud/spring-cloud-sleuth/pull/1562#issuecomment-586573093

> @simonbasle @robotmrv I noticed in #1126 we started handling both onNext and onComplete, to finish a span. The code that handles onComplete when onNext isn't called is dead code. At least in our tests it isn't invoked.
> 
> I would like to remove the onComplete without onNext edge case as it isn't tested and also it is very confusing. Is there any reason why we should be so defensive?

When you look at the `Flux` or `Mono` level, not the `Subscriber` level, pipelines which end results appear empty are not that uncommon. Any reactive method that returns a `Mono<Void>` for instance...

That said, Sleuth works by wrapping `Subscriber`s, which are the steps in one activation of the pipeline. The cases where such a step is empty are probably fewer. For instance, the `Subscriber` generated by `ignoreElements()`, despite producing a `Flux` that appears empty to the end consumer, still sees `onNext()` calls from its source. It just doesn't propagate them. If there is another operator after that, its `Subscriber` will only see an `onComplete()`.

Another way of triggering this code path is by having an empty source:
```java
Flux.empty()
    .map(i -> i)
    .filter(i -> true)
    .flatMap(i -> Mono.error(new RuntimeException("won't be thrown because no onNext"));
```

Hope this helps improving the tests/understanding when `onComplete` might matter...

It is probably true that the vast majority of _`Subscriber`s_ will not `onComplete()` without an `onNext()`, but when one considers the whole sequence (ie. the `Flux` or `Mono`) this cannot be considered an edge case. All Mon

adriancole:
We also need a test case for onNext code that might issue a cancel after its first element. https://github.com/spring-cloud/spring-cloud-sleuth/pull/1572#pullrequestreview-364012718

adriancole:
another note about empty source https://github.com/spring-cloud/spring-cloud-sleuth/pull/1572#discussion_r383790741

adriancole:
TODO: also make a test that shows if possible spans are not started until the mono representing the call is subscribed to. There was a bug around this recently.

olssonoskar:
Scope decoration seems to be order dependent if using Project reactors recommended approach to blocking calls. It seems that everything below .subscribeOn(Schedulers.boundedElastic()) in the chain will have MDC decorated while for everything above it will be empty.

[...]

PR #1572, opened 2020-02-25
Ensures spring context is closed in brave tests

adriancole:
This also normalizes how we subscribe to request monos.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/HttpClientBeanPostProcessor.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/HttpClientBeanPostProcessor.java
@@ -182,6 +182,7 @@ class HttpClientBeanPostProcessor implements BeanPostProcessor {
 
 		@Override
 		public void accept(HttpClientResponse response, Connection connection) {
+			// TODO: is there a way to read the request at response time?
 			handle(response.currentContext(), response, null);
 		}
 
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java
@@ -220,6 +220,7 @@ final class TraceExchangeFilterFunction implements ExchangeFilterFunction {
 
 		final CurrentTraceContext currentTraceContext;
 
+		// TODO: this isn't implemented correctly. error and success could both be called
 		boolean done;
 
 		WebClientTracerSubscriber(CoreSubscriber<? super ClientResponse> actual,
@@ -251,7 +252,8 @@ final class TraceExchangeFilterFunction implements ExchangeFilterFunction {
 					try (Scope scope = currentTraceContext.maybeScope(parent)) {
 						subscription.cancel();
 					}
-					finally {
+					finally { // TODO: this is probably incorrect as cancel happens
+								// routinely in unary subscription.
 						if (log.isDebugEnabled()) {
 							log.debug("Subscription was cancelled. Will close the span ["
 									+ clientSpan + "]");
@@ -274,6 +276,7 @@ final class TraceExchangeFilterFunction implements ExchangeFilterFunction {
 								.build());
 			}
 			finally {
+				// TODO: is there a way to read the request at response time?
 				handleReceive(response, null);
 			}
 		}
--- a/tests/spring-cloud-sleuth-instrumentation-reactor-tests/src/test/java/org/springframework/cloud/sleuth/instrument/web/client/ReactorNettyHttpClientBraveTests.java
+++ b/tests/spring-cloud-sleuth-instrumentation-reactor-tests/src/test/java/org/springframework/cloud/sleuth/instrument/web/client/ReactorNettyHttpClientBraveTests.java
@@ -23,16 +23,12 @@ import brave.http.HttpTracing;
 import brave.test.http.ITHttpAsyncClient;
 import io.netty.channel.ChannelOption;
 import io.netty.handler.timeout.ReadTimeoutHandler;
-import org.junit.After;
-import org.junit.Before;
 import org.junit.Ignore;
 import org.junit.Test;
 import org.reactivestreams.Subscription;
 import reactor.core.CoreSubscriber;
-import reactor.core.publisher.Hooks;
 import reactor.core.publisher.Mono;
 import reactor.core.publisher.Operators;
-import reactor.core.scheduler.Schedulers;
 import reactor.netty.ByteBufFlux;
 import reactor.netty.http.client.HttpClient;
 import reactor.netty.http.client.HttpClientResponse;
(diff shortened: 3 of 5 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/HttpClientBeanPostProcessor.java
```diff
@@ -182,6 +182,7 @@ static void parseConnectionAddress(Connection connection, Span span) {
 
 		@Override
 		public void accept(HttpClientResponse response, Connection connection) {
+			// TODO: is there a way to read the request at response time?
```
adriancole: @violetagg is there any way to do this at the moment or some future release of reactor-netty? The intent is to see the request that actually caused the response. The original request may have been rewritten due to redirects etc.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java
```diff
@@ -274,6 +276,7 @@ public void onNext(ClientResponse response) {
 								.build());
 			}
 			finally {
+				// TODO: is there a way to read the request at response time?
```
adriancole: @rstoyanchev @bclozel  is there any way to do this at the moment or some future release of reactor-netty? The intent is to see the request that actually caused the response. The original request may have been rewritten due to redirects etc.

adriancole: updated the copy/paste :P

tests/spring-cloud-sleuth-instrumentation-reactor-tests/src/test/java/org/springframework/cloud/sleuth/instrument/web/client/WebClientBraveTests.java
```diff
@@ -107,14 +99,12 @@ public void onSubscribe(Subscription s) {
 			}
 
 			@Override
-			public void onNext(ClientResponse t) {
-				Subscription s = ref.getAndSet(null);
-				if (s != null) {
+			public void onNext(ClientResponse clientResponse) {
+				if (ref.getAndSet(null) != null) {
 					callback.onSuccess(null);
-					s.cancel();
```
adriancole: the other test doesn't call cancel, but I noticed sometimes this is called. This causes a problem really, if we assume that cancel is an error. For example, a success response might not happen due to this cancel depending on order cc @simonbasle @robotmrv

tests/spring-cloud-sleuth-instrumentation-reactor-tests/src/test/java/org/springframework/cloud/sleuth/instrument/web/client/WebClientBraveTests.java
```diff
@@ -128,7 +118,8 @@ public void onError(Throwable t) {
 			@Override
 			public void onComplete() {
 				if (ref.getAndSet(null) != null) {
-					callback.onSuccess(null);
+					callback.onError(
```
simonbasle: isn't that path testing the "empty source" case? ie. if the `ref` wasn't set to `null` previously, it means that we completed directly, aka. the source is empty. this is not an error. note that a well-formed source MUST guarantee `onNext`-before-`onComplete` order.

tests/spring-cloud-sleuth-instrumentation-reactor-tests/src/test/java/org/springframework/cloud/sleuth/instrument/web/client/WebClientBraveTests.java
simonbasle: `s` is the object that produced the `ClientResponse`, the `Subscription` generated by the `request` Mono, and if we're in `onNext` it means that it already produced it... so `s.cancel()` doesn't bring much _for that particular case_.

if the test could run on an arbitrary source, then `s.cancel()` would make sense to signal we're only interested in the first incoming element.

tests/spring-cloud-sleuth-instrumentation-reactor-tests/src/test/java/org/springframework/cloud/sleuth/instrument/web/client/WebClientBraveTests.java
```diff
@@ -107,14 +99,12 @@ public void onSubscribe(Subscription s) {
...
-				Subscription s = ref.getAndSet(null);
-				if (s != null) {
+			public void onNext(ClientResponse clientResponse) {
+				if (ref.getAndSet(null) != null) {
 					callback.onSuccess(null);
-					s.cancel();
 				}
 				else {
-					Operators.onNextDropped(t, currentContext());
+					Operators.onNextDropped(clientResponse, currentContext());
```
simonbasle: again, with a `Mono` source this should never happen... maybe make it fail the test rather than `onNextDropped` (which by default only logs, IIRC).

tests/spring-cloud-sleuth-instrumentation-reactor-tests/src/test/java/org/springframework/cloud/sleuth/instrument/web/client/WebClientBraveTests.java
simonbasle: again, since that is dedicated to a specific test, having such an early completion could indeed be considered an error (ie. we expect exactly `onNext(ClientResponse)` then `onComplete()`).

adriancole: we have so few tests about these edge cases.. https://github.com/spring-cloud/spring-cloud-sleuth/issues/1570

main thing is that the two types of integration tests here, WebClient and HttpClient should have a coherent means to do the subscription dance, or a clear reason why they are different. This PR so far attempts to get them to do the same thing.. then we can start documenting all the odd possibilities therein

tests/spring-cloud-sleuth-instrumentation-reactor-tests/src/test/java/org/springframework/cloud/sleuth/instrument/web/client/WebClientBraveTests.java
adriancole: sounds good

adriancole:
I'll do my best to parse the existing mentions as well the one I've not gotten to (#1570) tomorrow. appreciate the responses so far and keep them coming. Unlike webflux, I'm ok with multiple responses prior to onComplete.

tests/spring-cloud-sleuth-instrumentation-reactor-tests/src/test/java/org/springframework/cloud/sleuth/instrument/web/client/WebClientBraveTests.java
adriancole: thanks.. moved the thing about arbitrary source here https://github.com/spring-cloud/spring-cloud-sleuth/issues/1570#issuecomment-591247178

tests/spring-cloud-sleuth-instrumentation-reactor-tests/src/test/java/org/springframework/cloud/sleuth/instrument/web/client/WebClientBraveTests.java
adriancole: empty source, we'll handle explicitly here. https://github.com/spring-cloud/spring-cloud-sleuth/issues/1570#issuecomment-591247178

thanks for the reminder. this particular one is test code which isn't written to also handle empty source, plus there's no empty source test, yet

tests/spring-cloud-sleuth-instrumentation-reactor-tests/src/test/java/org/springframework/cloud/sleuth/instrument/web/client/ReactorNettyHttpClientBraveTests.java
```diff
@@ -82,13 +66,15 @@ static HttpClient testHttpClient() {
 	}
 
 	@Override
-	protected void closeClient(HttpClient client) {
-		// HttpClient is not Closeable
+	protected void closeClient(AnnotationConfigApplicationContext context) {
+		context.close(); // ensures shutdown hooks fire
```
adriancole: reactor hooks are a PreDestroy thing.. we formerly needed hacks because this test leaked spring contexts.

adriancole:
any more comments after are welcome

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java
rstoyanchev: That would be an `ExchangeFilterFunction` which this [processor is using](https://github.com/spring-cloud/spring-cloud-sleuth/blob/0227c050985ca656306e50eb8ab0c69792b89420/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java#L139-L149). What you do from there, extending `Mono<ClientResponse>` and wrapping its `Subscriber` is not an expected way to use the API. Nevertheless the request seems to be passed into the former but not into the latter, but probably could be.

adriancole: thanks for the hint! added it to the todo bucket :) https://github.com/spring-cloud/spring-cloud-sleuth/pull/1573/files#diff-bb63d22a058c70307dc3b4e017e8b0b4R98

PR #1575, opened 2020-02-28
Adds wiring for HttpRequestParser and HttpResponseParser

adriancole:
The default span data policy for HTTP requests is described in Brave:
https://github.com/openzipkin/brave/tree/master/instrumentation/http#span-data-policy

To add different data to the span, you need to register a bean of type
`brave.http.HttpRequestParser` or `brave.http.HttpResponseParser` based on when
the data is collected.

The bean names correspond to the request or response side, and whether it is
a client or server. For example, `sleuthHttpClientRequestParser` changes what
is collected before a client request is sent to the server.

For your convenience `@HttpClientRequestParser`, `@HttpClientResponseParser`
and corresponding server annotations can be used to inject the proper beans
or to  reference the bean names via their static String `NAME` fields.

Here's an example adding the HTTP url in addition to defaults:
```java
@Configuration
class Config {
	@Bean(name = { HttpClientRequestParser.NAME, HttpServerRequestParser.NAME })
	HttpRequestParser sleuthHttpServerRequestParser() {
		return (req, context, span) -> {
			HttpRequestParser.DEFAULT.parse(req, context, span);
			String url = req.url();
			if (url != null) {
				span.tag("http.url", url);
			}
		};
	}
}
```


spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/client/feign/FeignRetriesTests.java
```diff
@@ -69,8 +68,7 @@
 					.addScopeDecorator(StrictScopeDecorator.create()).build())
 			.spanReporter(this.reporter).build();
 
-	HttpTracing httpTracing = HttpTracing.newBuilder(this.tracing)
-			.clientParser(SleuthHttpParserAccessor.getClient()).build();
```
adriancole: here and everywhere I removed the legacy parser from unrelated tests

adriancole:
review after merge welcome, but I think the i's are dotted

PR #1576, opened 2020-03-01
Fixes WebClient which started a client span prior to a subscription

adriancole:
In looking at underlying HttpClient mechanics, I noticed the reactor
call doesn't happen until subscribe. Before this change, we started the
client span at the ExchangeFilterFunction, not at subscribe time.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java
@@ -138,14 +138,7 @@ final class TraceExchangeFilterFunction implements ExchangeFilterFunction {
 
 	@Override
 	public Mono<ClientResponse> filter(ClientRequest request, ExchangeFunction next) {
-		HttpClientRequest wrapper = new HttpClientRequest(request);
-		TraceContext parent = currentTraceContext().get();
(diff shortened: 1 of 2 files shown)


adriancole:
@simonbasle @violetagg before I go back and fixup test cases, does what I just did make sense with your understanding of WebClient and HttpClient?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java
```diff
@@ -177,27 +170,30 @@ CurrentTraceContext currentTraceContext() {
...
 
-			this.next.exchange(request).subscribe(new WebClientTracerSubscriber(
-					subscriber, context, parent, span, this));
+			HttpClientRequest wrapper = new HttpClientRequest(request);
+			Span span = handler.handleSendWithParent(wrapper, parent);
+			if (log.isDebugEnabled()) {
+				log.debug("HttpClientHandler::handleSend: " + span);
+			}
+
+			this.next.exchange(wrapper.buildRequest()).subscribe(
```
adriancole: it is my understanding that this `.exchange()` starts the request, so we should do the `handleSend()` stuff here as that's right before the request is made. correct?

adriancole:
I've looked at other code and decided that this doesn't require second eyes. there's no intent to stat a client request before the mono is subscribed to, and so we shouldn't do similar for the client span.

PR #1577, opened 2020-03-02
Attempts to align feign code with master

adriancole:
This tries to pick in the relevant changes made in 3813cf9dd47f98db0e075412abbffbd9d6742974

Notably, there's one glitch, `TraceFeignAspect` still passes bean, not
wrappedBean. The latter trips out `ManuallyCreatedLoadBalancerFeignClientTests`
as the nonexistenturl raises a hard error in Ribbon. I *think* this is a bug and
tests need to just adjust for that, but need a second opinion. Also, I'm not
entirely sure the intent of using the broken url then asserting against a
success result..

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/TraceFeignAspect.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/TraceFeignAspect.java
@@ -53,6 +53,8 @@ class TraceFeignAspect {
 			log.debug("Executing feign client via TraceFeignAspect");
 		}
 		if (bean != wrappedBean) {
+			// NOTE: in master(3813cf9dd47f98db0e075412abbffbd9d6742974),
+			// this is executeTraceFeignClient(wrappedBean, pjp)
 			return executeTraceFeignClient(bean, pjp);
 		}
 		return pjp.proceed();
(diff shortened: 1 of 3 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/TraceFeignAspect.java
```diff
@@ -53,6 +53,8 @@ public Object feignClientWasCalled(final ProceedingJoinPoint pjp) throws Throwab
 			log.debug("Executing feign client via TraceFeignAspect");
 		}
 		if (bean != wrappedBean) {
+			// NOTE: in master(3813cf9dd47f98db0e075412abbffbd9d6742974),
+			// this is executeTraceFeignClient(wrappedBean, pjp)
 			return executeTraceFeignClient(bean, pjp);
```
marcingrzejszczak: This should be `wrappedBean` I guess

adriancole: yep just it blows up tests if we fix it:)

marcingrzejszczak: Can't we align those tests as we did in master?

adriancole: master swapped out ribbon so it changes the nature of the tests I think. that's why I paused here

marcingrzejszczak: Ok I see. Let's merge it as it is then and I'll look at this once I have spare time. It won't make things worse obviously ;)

adriancole:
cool. if master needs a release today I didnt merge from 2.2 latest. I wont be able to until later

marcingrzejszczak:
I don't think master is released today

PR #1578, opened 2020-03-04
Adds base integration test for reactor clients and adds cancelation tests

adriancole:
in order to test typical concerns common to all reactor based clients, this makes a base type. Afterwards, we can do things like canceled subscriptions and make sure both WebClient and Reactor's HttpClient (both of which are pinned to reactor's types) get tested.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java
@@ -18,6 +18,8 @@ package org.springframework.cloud.sleuth.instrument.web.client;
 
 import java.util.List;
 import java.util.concurrent.CancellationException;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.function.Function;
 
@@ -112,13 +114,6 @@ final class TraceExchangeFilterFunction implements ExchangeFilterFunction {
 
 	private static final Log log = LogFactory.getLog(TraceExchangeFilterFunction.class);
 
-	static final Exception CANCELLED_ERROR = new CancellationException("CANCELLED") {
-		@Override
-		public Throwable fillInStackTrace() {
-			return this; // stack trace doesn't add value here
-		}
-	};
-
 	final LazyBean<HttpTracing> httpTracing;
 
 	final Function<? super Publisher<DataBuffer>, ? extends Publisher<DataBuffer>> scopePassingTransformer;
@@ -193,13 +188,16 @@ final class TraceExchangeFilterFunction implements ExchangeFilterFunction {
 				log.debug("HttpClientHandler::handleSend: " + span);
 			}
 
+			// NOTE: We are starting the client span for the request here, but it could be
+			// canceled prior to actually being invoked. TraceWebClientSubscription will
+			// abandon this span, if cancel() happens before request().
 			this.next.exchange(wrapper.buildRequest()).subscribe(
-					new WebClientTracerSubscriber(subscriber, context, span, this));
+					new TraceWebClientSubscriber(subscriber, context, span, this));
 		}
 
 	}
 
-	private static final class WebClientTracerSubscriber
+	static final class TraceWebClientSubscriber extends AtomicReference<Span>
 			implements CoreSubscriber<ClientResponse> {
 
 		final CoreSubscriber<? super ClientResponse> actual;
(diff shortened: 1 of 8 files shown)


adriancole:
Think this sorts out the cancel code I couldn't properly explain before.

Style/improvement hints welcome. cc @simonbasle @robotmrv @violetagg

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/client/HttpClientBeanPostProcessorTest.java
```diff
@@ -0,0 +1,72 @@
...
+import reactor.core.publisher.Mono;
+import reactor.netty.Connection;
+
+import org.springframework.cloud.sleuth.instrument.web.client.HttpClientBeanPostProcessor.PendingClientSpan;
+import org.springframework.cloud.sleuth.instrument.web.client.HttpClientBeanPostProcessor.TracingMapConnect;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@RunWith(MockitoJUnitRunner.class)
+public class HttpClientBeanPostProcessorTest {
```
adriancole: no good deed goes unpunished :) @marcingrzejszczak any idea why this dies in maven, but not in intellij?

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/client/HttpClientBeanPostProcessorTest.java
```diff
@@ -0,0 +1,72 @@
...
+		TracingMapConnect tracingMapConnect = new TracingMapConnect(() -> traceContext);
+
+		Mono<Connection> original = Mono.just(connection).handle((t, ctx) -> {
+			assertThat(ctx.currentContext().get(TraceContext.class))
+					.isSameAs(traceContext);
+			assertThat(ctx.currentContext().get(PendingClientSpan.class)).isNotNull();
+		});
+
+		// Wrap and run the assertions
+		tracingMapConnect.apply(original, bootstrap).log().subscribe();
```
adriancole: ps prior to merge we can scrub the `.log()` here

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/web/client/HttpClientBeanPostProcessorTest.java
marcingrzejszczak: Ordering maybe? Delay? The question is why it's failing.

adriancole: both assertions failed and only thing I can think of is if somehow reactor subscriber context has static or thread local state which can taint other things.. I pulled this file out for now.

adriancole:
ok done messing around. hopefully all good now.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java
```diff
@@ -309,10 +288,57 @@ public Context currentContext() {
...
+
+		TraceWebClientSubscription(Subscription delegate,
+				AtomicReference<Span> pendingSpan) {
+			this.delegate = delegate;
+			this.pendingSpan = pendingSpan;
+		}
+
+		@Override
+		public void request(long n) {
+			if (compareAndSet(false, true)) {
```
robotmrv: @adriancole looks strange, you apply only first request. Why to skip others?

adriancole: because only concern was if this signal happened.. if distracting, I can do it on the others.

adriancole: actually I need to do it here, as the most important thing is if it happened. doing elsewhere would invalidate that. probably if this is confusing I can put a comment.

robotmrv: `compareAndSet(false, true)` part is ok - this value is used in `cancel()`, but I do not understand `if ()` part, why it would be invalidated?

robotmrv: I mean this `Subscriber` should not impact the way signals are propagated but only trigger appropriate `brave`  API

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java
```diff
@@ -193,13 +188,16 @@ public void subscribe(CoreSubscriber<? super ClientResponse> subscriber) {
...
+			// abandon this span, if cancel() happens before request().
 			this.next.exchange(wrapper.buildRequest()).subscribe(
-					new WebClientTracerSubscriber(subscriber, context, span, this));
+					new TraceWebClientSubscriber(subscriber, context, span, this));
 		}
 
 	}
 
-	private static final class WebClientTracerSubscriber
+	static final class TraceWebClientSubscriber extends AtomicReference<Span>
```
robotmrv: as I notice most of `core` operators implement `Scannable` the same way as `org.springframework.cloud.sleuth.instrument.reactor.ScopePassingSpanSubscriber` maybe this one should also do it?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java
adriancole: good idea. I will switch this in next change

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java
adriancole: ok sure

adriancole: this implies a good bit of thinking/work so will do it in another change

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java
adriancole: https://github.com/spring-cloud/spring-cloud-sleuth/pull/1581

PR #1579, opened 2020-03-05
Switches to a less flakey span reporter

adriancole:
The mentioned test is super flakey.. this should help

PS I raised this on master, not 2.2.x, as 2.2.x is dead due to a missing publish of s-c-openfeign which I have no idea how to rectify. Let's switch the base branch when no longer the case.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/util/BlockingQueueSpanReporter.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/util/BlockingQueueSpanReporter.java
@@ -29,7 +29,10 @@ public class BlockingQueueSpanReporter implements Reporter<Span> {
 
 	private final LinkedBlockingQueue<Span> spans = new LinkedBlockingQueue<>();
 
-	/** Blocks until a span is reported or throws an {@link AssertionError}. */
+	/**
+	 * Blocks until a span is reported or throws an {@link AssertionError}.
+	 * @return the first span not yet taken.
+	 */
 	public Span takeSpan() {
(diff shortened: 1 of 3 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/util/BlockingQueueSpanReporter.java
```diff
@@ -0,0 +1,71 @@
...
+import java.util.concurrent.LinkedBlockingQueue;
+import java.util.concurrent.TimeUnit;
+
+import zipkin2.Span;
+import zipkin2.reporter.Reporter;
+
+/**
+ * Like {@link ArrayListSpanReporter}, except appropriate for async instrumentation.
+ */
+public class BlockingQueueSpanReporter implements Reporter<Span> {
```
adriancole: I put this in main, not test, as the other one is also in main..

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/util/BlockingQueueSpanReporter.java
```diff
@@ -0,0 +1,71 @@
...
+
+	private final LinkedBlockingQueue<Span> spans = new LinkedBlockingQueue<>();
+
+	/** Blocks until a span is reported or throws an {@link AssertionError}. */
+	public Span takeSpan() {
+		Span result = takeSpan(3_000);
```
adriancole: the highly scientific 3 seconds works well in practice in brave. We've tuned the magic number over the years :)

PR #1582, opened 2020-03-06
Ignore legacy keys in MDC.

bjartek:
Added a new property disableLegacyKeys to SLF4jProperties. If that property is set ignore the legacy keys in MDC


bjartek:
Not really sure why that test failed. I did not touch any codepath there I think.

marcingrzejszczak:
Actually, for master we can remove those keys whatsover. Are you willing to change this PR to remove instead of ignoring legacy keys?

bjartek:
Sure I can fix that!

bjartek:
Was this what you had in mind or did i misunderstand the task?

marcingrzejszczak:
Yes! Thanks a lot

adriancole:

PR #1586, opened 2020-03-19
Finalizes reactor-netty HttpClient implementations

adriancole:
Uses changes in latest reactor-netty to close most test gaps

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -132,6 +132,13 @@
 
 	<dependencyManagement>
 		<dependencies>
+			<!-- force latest reactor-netty -->
+			<dependency>
+				<groupId>io.projectreactor.netty</groupId>
+				<artifactId>reactor-netty</artifactId>
+				<version>0.9.6.RELEASE</version>
+				<optional>true</optional>
+			</dependency>
 			<dependency>
 				<groupId>org.springframework.cloud</groupId>
 				<artifactId>spring-cloud-sleuth-dependencies</artifactId>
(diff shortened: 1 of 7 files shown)


pom.xml
```diff
@@ -132,6 +132,13 @@
 
 	<dependencyManagement>
 		<dependencies>
+			<!-- force latest reactor-netty -->
+			<dependency>
+				<groupId>io.projectreactor.netty</groupId>
+				<artifactId>reactor-netty</artifactId>
+				<version>0.9.6.BUILD-SNAPSHOT</version>
```
adriancole: @simonbasle @violettagg 0.9.6 has everything needed now. I raised a nice to have herehttps://github.com/reactor/reactor-netty/issues/1036

As we don't have any reactor-netty server-side instrumentation (only [WebFlux](https://github.com/spring-cloud/spring-cloud-sleuth/pull/1587) if I'm not mistaken), then a new release of reactor-netty should allow us to button this up.

one thing to get some input on, which is if we shouldn't disable webflux client instrumentation when reactor-netty is underneath.. because the latter is more accurate and it would also answer an accidentally dual instrumentation concern. we'd still need the webclient BPP when it isnt' reactor.. maybe @marcingrzejszczak has an idea on if this could be done neatly..

adriancole: I changed this to the final version. the other point about dual instrumentation will cover in #1571

pom.xml
```diff
@@ -132,6 +132,13 @@
 
 	<dependencyManagement>
 		<dependencies>
+			<!-- force latest reactor-netty -->
```
adriancole: @marcingrzejszczak is it ok to merge pinning this for now as we can't control the upstream versions and break unless very latest?

marcingrzejszczak: Of course

PR #1594, opened 2020-04-02
Updates to latest brave and removes all static thread local use

adriancole:
This moves to the pending brave release. This passes the same tests
locally as before this commit.

feign has some problems on 2.2.x so I skipped like so:

```
./mvnw clean install -pl -:spring-cloud-sleuth-instrumentation-feign-tests
```


adriancole:
2.2.x is broken so merging without CI

PR #1595, opened 2020-04-03
Rewrites Slf4jScopeDecorator internally to re-use Brave's

adriancole:
This uses fancy new [Baggage](https://github.com/openzipkin/brave/tree/master/brave#baggage) apis, eventhough configuration is the same for now.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/log/Slf4jScopeDecorator.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/log/Slf4jScopeDecorator.java
@@ -16,22 +16,20 @@
 
 package org.springframework.cloud.sleuth.log;
 
-import java.util.AbstractMap;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.List;
+import java.util.LinkedHashSet;
+import java.util.Set;
+import java.util.TreeSet;
 
-import brave.internal.HexCodec;
-import brave.internal.Nullable;
-import brave.propagation.CurrentTraceContext;
-import brave.propagation.ExtraFieldPropagation;
+import brave.baggage.BaggageField;
+import brave.baggage.BaggageFields;
+import brave.baggage.CorrelationScopeDecorator;
+import brave.context.slf4j.MDCScopeDecorator;
+import brave.propagation.CurrentTraceContext.Scope;
+import brave.propagation.CurrentTraceContext.ScopeDecorator;
 import brave.propagation.TraceContext;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.slf4j.MDC;
 
 import org.springframework.cloud.sleuth.autoconfig.SleuthProperties;
-import org.springframework.util.StringUtils;
 
 /**
  * Adds {@linkplain MDC} properties "traceId", "parentId", "spanId" and "spanExportable"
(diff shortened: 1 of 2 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/log/Slf4jScopeDecorator.java
```diff
@@ -43,196 +41,48 @@
...
-			MDC.put(LEGACY_EXPORTABLE_NAME, sampled);
-			log("Starting scope for span: {}", currentSpan);
-			if (currentSpan.parentId() != null) {
-				if (log.isTraceEnabled()) {
-					log.trace("With parent: {}", currentSpan.parentId());
-				}
+		CorrelationScopeDecorator.Builder builder = MDCScopeDecorator.newBuilder().clear()
+				.addField(BaggageFields.TRACE_ID).addField(BaggageFields.PARENT_ID)
+				.addField(BaggageFields.SPAN_ID)
+				.addField(BaggageFields.SAMPLED, "spanExportable");
```
adriancole, line 61: this is the only non-default name currently in use by sleuth for built-in fields

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/log/Slf4jScopeDecorator.java
```diff
@@ -43,196 +41,48 @@
...
+		retained.addAll(sleuthProperties.getBaggageKeys());
+		retained.addAll(sleuthProperties.getLocalKeys());
+		retained.addAll(sleuthProperties.getPropagationKeys());
+
+		for (String name : retained) {
+			if (whitelist.contains(name)) {
+				// Until we move off ExtraFieldPropagation onto BaggagePropagation,
+				// manually create the fields...
+				builder.addField(BaggageField.create(name));
+				builder.addDirtyName(name);
```
adriancole, line 76: there are some issue history in sleuth about people wanting it to intentionally clobber out-of-band changes. Later configuration can be added to make this far cheaper from an overhead perspective for those not fighting with `MDC.put`

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/log/Slf4jScopeDecorator.java
```diff
@@ -43,196 +41,48 @@
...
-	private void log(String text, TraceContext span) {
-		if (span == null) {
-			return;
-		}
-		if (log.isTraceEnabled()) {
-			log.trace(text, span);
-		}
+	@Override
+	public Scope decorateScope(TraceContext context, Scope scope) {
+		return LEGACY_IDS.decorateScope(context, delegate.decorateScope(context, scope));
```
adriancole, line 85: this is effectively the same as mapping dual names to the span identifiers. From a design POV, brave chose to only allow one mapped name per field-decorator. this reduces complexity and improves speed.  Most users will not need to sync multiple names to the same trace context field, but those who can, may use multiple decorators. Typically, we'd wire each decorator as a separate `@Bean` this trick is made just because this change is to rewrite the internals, not break up the type itself.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/log/Slf4JSpanLoggerTest.java
```diff
@@ -102,19 +97,17 @@ public void should_remove_entries_from_mdc_for_null_span() throws Exception {
 		ExtraFieldPropagation.set(this.span.context(), "my-baggage", "my-value");
 		ExtraFieldPropagation.set(this.span.context(), "my-propagation",
 				"my-propagation-value");
-		this.slf4jScopeDecorator.decorateScope(this.span.context(), () -> {
```
adriancole, line 105: this test case leaked data.. I rewrote it to not do that, but the assertions make the same intent true

adriancole:
test failures in feign are unrelated due to broken 2.2.x branch

PR #1597, opened 2020-04-03
Moves Slf4jScopeDecorator to use CorrelationField internally

adriancole:


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/log/Slf4jScopeDecorator.java
```diff
@@ -45,36 +46,46 @@
 
 	// Backward compatibility for all logging patterns
 	private static final ScopeDecorator LEGACY_IDS = MDCScopeDecorator.newBuilder()
-			.clear().addField(BaggageFields.TRACE_ID, "X-B3-TraceId")
-			.addField(BaggageFields.PARENT_ID, "X-B3-ParentSpanId")
-			.addField(BaggageFields.SPAN_ID, "X-B3-SpanId")
-			.addField(BaggageFields.SAMPLED, "X-Span-Export").build();
+			.clear()
+			.addField(CorrelationField.newBuilder(BaggageFields.TRACE_ID)
```
adriancole, line 50: the plan for 3.x is to wire these up as List bindings: one for `List<BaggageField>` and another for `List<CorrelationField>`. We build these based on properties and then they are injected to construct BaggageFieldPropagation and CorrelationScopeDecorator respectively.

adriancole: not these specific fields I mean baggage in general

adriancole:
feign again

PR #1598, opened 2020-04-04
Fixes bugs in skip pattern code and Feign tests

adriancole:
Before, the client sampler would only skip the path named ""! This was
the path used in one of the failing Feign tests, which made
troubleshooting like a murder mystery.

The "" is a regression in spring-cloud-openfeign noted below:
https://github.com/spring-cloud/spring-cloud-openfeign/pull/245/files#r403424379


adriancole:
oh my god, it is finally green. I can move on with my life !!

PR #1602, opened 2020-04-04
Stops configuring ExtraFieldPropagation. Starts configuring BaggagePropagation

adriancole:
This makes 3.x only configure `BaggagePropagation`

`ExtraFieldPropagation` was never intended for end-user code (neither is baggage field), but the `ExtraFieldPropagation` apis internally forward to the new stuff.

There will be a bunch of doc work to do later, but here is the [intro in Brave](https://github.com/openzipkin/brave/tree/master/brave#baggage)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/PropertyBasedBaggageFields.java
```diff
@@ -0,0 +1,58 @@
...
+import static java.util.Arrays.asList;
+
+/**
+ * Wire up property-based {@linkplain BaggageField baggage fields}, so that they appear as
+ * if they were defined one-by-one. This allows users to contribute fields and also for us
+ * to use the list of them above.
+ */
+@Configuration(proxyBeanMethods = false)
+@ConditionalOnProperty(value = "spring.sleuth.enabled", matchIfMissing = true)
+public class PropertyBasedBaggageFields implements BeanFactoryPostProcessor {
```
adriancole, line 39: @dsyer @wilkinsona either of you know how to do what I'm trying to do here? MultipleHopsIntegrationTests fails because this can't read any of the properties. Most tests pass though.

Basically, I want to programmatically register these constants so that other things can base on them. They need no information except a property to do that, and I spent several hours trying things and failed I guess.

any other way to accomplish same would be great!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
```diff
@@ -143,50 +146,53 @@ SpanNamer sleuthSpanNamer() {
...
-			for (String key : sleuthProperties.getPropagationKeys()) {
-				factoryBuilder = factoryBuilder.addField(key);
+	BaggagePropagation.FactoryBuilder baggagePropagationFactoryBuilder() {
+		return BaggagePropagation.newFactoryBuilder(B3Propagation.FACTORY);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	Propagation.Factory sleuthPropagation(
+			BaggagePropagation.FactoryBuilder factoryBuilder, List<BaggageField> baggage,
```
adriancole, line 163: ex I need to use them here..

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/log/SleuthLogAutoConfiguration.java
```diff
@@ -51,13 +64,41 @@
 	@EnableConfigurationProperties(SleuthSlf4jProperties.class)
 	protected static class Slf4jConfiguration {
 
+		@Bean
+		List<CorrelationField> correlationFields(List<BaggageField> baggage,
```
adriancole, line 68: and here..

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/PropertyBasedBaggageFields.java
adriancole: sorry.. this was a test config problem. the application name was wrong.

anyway feedback welcome if you think there's a cleaner way to accomplish this.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/PropertyBasedBaggageFields.java
```diff
@@ -0,0 +1,58 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.env.Environment;
+
+import static java.util.Arrays.asList;
+
+/**
+ * Wire up property-based {@linkplain BaggageField baggage fields}, so that they appear as
+ * if they were defined one-by-one. This allows users to contribute fields and also for us
+ * to use the list of them above.
```
adriancole, line 35:
```suggestion
 * to use {@code List<BaggageField>} to reference all fields without needing a trace context.
```

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
```diff
@@ -143,50 +146,53 @@ SpanNamer sleuthSpanNamer() {
...
+			if (local.contains(field.name())) {
+				factoryBuilder.addField(field);
 			}
-		}
-		if (!sleuthProperties.getLocalKeys().isEmpty()) {
-			for (String key : sleuthProperties.getLocalKeys()) {
-				factoryBuilder = factoryBuilder.addRedactedField(key);
+			else if (toPrefix.contains(field.name())) {
+				List<String> prefixedNames = new ArrayList<>();
+				prefixedNames.add("baggage-" + field.name()); // for HTTP
```
adriancole, line 178: ps this is super chatty, to encode both ways. we should stop doing this at some point. ideally we can move onto the baggage header format being discussed in w3c (used to be named correlation-context)

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/multiple/MultipleHopsIntegrationTests.java
```diff
@@ -98,68 +103,64 @@ public void should_prepare_spans_for_export() throws Exception {
...
+		foo.updateValue(initialSpan.context(), "1");
+		bar.updateValue(initialSpan.context(), "2");
 		// end::baggage[]
 
-		try (Tracer.SpanInScope ws = this.tracer.withSpanInScope(initialSpan)) {
+		try (SpanInScope ws = this.tracer.withSpanInScope(initialSpan)) {
 			// tag::baggage_tag[]
-			initialSpan.tag("foo",
-					ExtraFieldPropagation.get(initialSpan.context(), "foo"));
-			initialSpan.tag("UPPER_CASE",
```
adriancole, line 126: mixing case is a road to perdition. let's not encourage this, especially in docs!!

adriancole:
sometimes before 3.0 final we owe a new doc, which is how *not* to use the baggage apis :)

basically users should always shield themselves from interfaces like this in business code. something like..


```java
interface MyBaggage {
  String username();
  void setUsername(String username);
}

class MyBaggage {
  @Autowired
  MyBaggage(TraceContext currentTraceContext, List<BaggageField> allConfigured) {
  // do the initialization etc.
  }

  public String username() {
    if (usernameField == null) return null;
    return usernameField.getValue(currentTraceContext.get());
  }

  public void setUsername(String username) {
    if (usernameField == null) return null;
    usernameField.updateValue(currentTraceContext.get(), username);
  }
}
```

At some point, we can consider making user-safe interfaces like feign for baggage.
that way they don't get stuck in our code just to access scope properties.

adriancole:
cc'ing a zipkin java party to have a look as I know baggage is exciting! @jeqo @anuraaga @jorgheymans @reta @shakuzen @llinder @devinsba @michaelsembwever  @zeagord @kojilin

adriancole:
PS the apis underneath here were epic thanks to @anuraaga who spent many hours on this design and banging out kinks in code with me. he's awesome and patient

adriancole:
FYI the value in looking up baggage fields as beans is somewhat limited as sometime by the time sleuth 3 is done, we'll support dynamic baggage. this proves the concept https://github.com/openzipkin/brave/pull/1142

It will still be useful initial wiring, to know which fields are remote. dynamic ones will be virtual fields embedded in one header, such as "baggage"

adriancole:
blocked on https://github.com/openzipkin/brave/pull/1143

adriancole:

PR #1603, opened 2020-04-06
Adds more robust introduction

adriancole:


docs/src/main/asciidoc/spring-cloud-sleuth.adoc
```diff
@@ -12,20 +12,39 @@ include::intro.adoc[]
...
+Zipkin is an Open Source tracing system, which includes a UI and various
+collectors, such as HTTP and messaging.
+
+Many Open Source and commercial products accept https://zipkin.io/zipkin-api/#/default/post_spans[Zipkin format].
+Some options are documented https://zipkin.io/pages/extensions_choices.html[here],
+but many are not. If you cannot use Zipkin and your product isn't listed, clarify
+with your support representative and have them update that page. In many cases,
+products already support Zipkin format, they just don't document it.
+
+Traces connect from host to host using header propagation. The default format is
```
shakuzen, line 33: I wonder if "service to service" is better than host to host as it uses the vocab of tracing and doesn't imply a service has to be a different host.

PR #1604, opened 2020-04-06
Renames spring.sleuth.propagation-keys to remote-keys

adriancole:
spring.sleuth.local-keys and spring.sleuth.remote-keys make sense
together, especially as baggage-keys will be removed.


adriancole:
The old property will still be read, just with a warning.

adriancole:

PR #1605, opened 2020-04-06
Deprecates spring.sleuth.baggage-keys property

adriancole:
The `spring.sleuth.baggage-keys` property assigned two headers for each
baggage. This causes unnecessary overhead and can be accomplished in another
way now:

```java
public class MyCode {
  static final BaggageField USER_ID = BaggageField.create("userId");

  @Autowired CurrentTraceContext currentTraceContext;

  @Nullable
  public String currentUserId() {
    return USER_ID.getValue(currentTraceContext.get());
  }

}

// In your @Configuration type, if you want to map multiple prefixes...
@Bean
BaggagePropagationConfig userIdBaggageConfig() {
  return SingleBaggageField.newBuilder(MyBaggage.USER_ID)
      .addKeyName("baggage-userId")
      .addKeyName("baggage_userId")
      .build();
}
```

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -380,9 +380,6 @@ The following listing shows integration tests that use baggage:
 [source,yml]
 ----
 spring.sleuth:
-  baggage-keys:
-    - baz
-    - bizarrecase
   local-keys:
     - bp
   remote-keys:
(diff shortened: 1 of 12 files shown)


marcingrzejszczak:
We need to document it with a gigantic `WARNING` message that this requires migration. Sth like if you don't set this explicitly then you won't see baggage being propagated.

adriancole:
yeah hopefully people aren't using those keys.. another way is we could temporarily support it, but remove docs for it and add a big warning in the logs.

adriancole:
changed to warn at startup

adriancole:

Issue #1416, opened 2019-07-31
Manually set baggage is not automatically cleaned up from MDC


aasaru:
I have configured:
```
spring.sleuth:
  propagation-keys:
    - x-session-id
  log:
    slf4j:
      whitelisted-mdc-keys: x-session-id 
```

If I pass this propagation key with a value to my app as HTTP header then after the request is completed it is automatically removed from MDC by Slf4jScopeDecorator.

But if I don't pass it as header but manually set the value somewhere during processing of the request then I have to manually set it to two places:
```
ExtraFieldPropagation.set(continuedSpan.context(), "x-session-id", "mySession123");
MDC.put("x-session-id", "mySession123");
```
and I also have to create an Interceptor and manually wipe it from MDC after the request processing has completed by calling: 
`MDC.remove("my-key")`

This is because Slf4jScopeDecorator only wipes any baggage that arrived with the request.

I propose to create a mechanism to register the value with single call (so it gets set in both to ExtraFieldPropagation and to MDC) and also after a value has been registered it would be automatically wiped from MDC after the request has completed.

Related [StackOverflow post](https://stackoverflow.com/questions/57206594/how-to-avoid-manual-cleanup-when-using-baggage-with-spring-cloud-sleuth-and-mdc/57260808?noredirect=1#comment101065798_57260808).

dlsrb6342:
I got same issue. I solved this by giving more higher order to my `ExtraFieldPropagationFilter` than `TraceWebFilter`. If you are using MVC, give more higher order than `TracingFilter`. 
With this solution, Slf4jScopeDecorator can get proper `previousMdc` value. 

But above is just workaround. If changing Slf4jScopeDecorator like below, can make a problem? 
```java
// current
final List<AbstractMap.SimpleEntry<String, String>> previousMdc = Stream
		.concat(whitelistedBaggageKeysWithValue(currentSpan),
			whitelistedPropagationKeysWithValue(currentSpan))
		.map((s) -> new AbstractMap.SimpleEntry<>(s, MDC.get(s)))
		.collect(Collectors.toList());

// to be
final List<AbstractMap.SimpleEntry<String, String>> previousMdc = Stream
		.concat(whitelistedBaggageKeys(), whitelistedPropagationKeys())
		.map((s) -> new AbstractMap.SimpleEntry<>(s, MDC.get(s)))
		.collect(Collectors.toList());
```
[...]

PR #1606, opened 2020-04-06
Reorganizes code under o.s.c.sleuth.baggage

adriancole:
This moves code and properties under org.springframework.cloud.sleuth.baggage
Properties are now under "spring.sleuth.baggage"

Those coming from 2.x should migrate with the following:
 * spring.sleuth.baggage-keys -> `BaggagePropagationCustomizer` described in #1605
 * spring.sleuth.local-keys -> spring.sleuth.baggage.local-fields
 * spring.sleuth.propagation-keys -> spring.sleuth.baggage.remote-fields
 * spring.sleuth.propagation.tag.whitelisted-keys -> spring.sleuth.baggage.tag-fields
 * spring.sleuth.log.slf4j.whitelisted-mdc-keys -> spring.sleuth.baggage.correlation-fields
 * spring.sleuth.log.slf4j.enabled -> spring.sleuth.baggage.correlation-enabled

Those using MDC should know that for performance reasons, we no longer
set the following fields: `parentId` `spanExportable`.

We also do not set fields to "dirty" by default. Doing this by default
raised the overhead substantially, especially in reactive applications.

Those who want sleuth to override manual `MDC.put` operations as
described in #1416, or with to add back `parentId` `spanExportable` need
to define their own bean:

Ex. this is the former setup:

```java
@Bean CorrelationScopeDecorator oldConfig(SleuthBaggageProperties baggageProperties) {
	CorrelationScopeDecorator.Builder builder = MDCScopeDecorator.newBuilder().clear()
			.add(SingleCorrelationField.create(BaggageFields.TRACE_ID))
			.add(SingleCorrelationField.create(BaggageFields.PARENT_ID))
			.add(SingleCorrelationField.create(BaggageFields.SPAN_ID))
			.add(SingleCorrelationField.newBuilder(BaggageFields.SAMPLED)
					.name("spanExportable").build());

        // Set all fields dirty, so that any changes made by MDC directly are reverted.
	for (String name : baggageProperties.getCorrelationFields()) {
		builder.add(SingleCorrelationField.newBuilder(BaggageField.create(name))
				.dirty().build());
	}
	return builder.build();
}

```


adriancole:
I ignored one test `TraceFilterWebIntegrationTests.should_not_create_a_span_for_error_controller()` otherwise should be good to go.

I'd appreciate if you can merge and follow-up @marcingrzejszczak I will help with migration guide later, but this PR has basically what I would write.

adriancole:
@making heads up on some changes. Feedback welcome!

adriancole:
fixed the one drift thing all tests should be ok now

adriancole:
Marked as breaking change for attention eventhough we still parse the old old properties.

Ones with most impact will be those parsing traces from log lines. They will need add a `CorrelationScopeDecorator` to re-add the `parentId` and `spanExportable` 

```java
@Bean CorrelationScopeCustomizer addParentAndSpanExportable() {
  return b -> b.add(SingleCorrelationField.create(BaggageFields.PARENT_ID))
                       .add(SingleCorrelationField.newBuilder(BaggageFields.SAMPLED)
                                                                   .name("spanExportable").build());
}
```

adriancole:

making:
@adriancole 
I'm still reading this huge changes...

From the perspective of the user using the default configuration, changing the log pattern seems to have more impact than changing the baggage itself. I'm aware that some users are relying on the current pattern to parse apps logs with Logstash.  It should be also mentioned in the migration guide. They maybe need to accept both log formats until migration finishes.

adriancole:
@making so logstash will coerce parentId and spanExportable to '-', correct? this means it won't fail, just won't have a value for these which are non critical properties anyway.

Unless you are talking about the `X-B3-TraceId` names which sadly our README left there for way too long ex 
https://github.com/spring-cloud/spring-cloud-sleuth/pull/1606/files#diff-7bd4a925c695c2eb0eced3872b9b965fL307

Those using them can add them back like this:
https://github.com/spring-cloud/spring-cloud-sleuth/wiki/Spring-Cloud-Sleuth-3.0-Migration-Guide#x-b3--mdc-fields-names-are-no-longer-set

These extra properties are a massive source of overhead and if we cannot make a change on a major version, even with instructions on how to do something deprecated, when can we fix this?

adriancole:
@making if this is about `X-B3-TraceId` these were deprecated for `traceId` etc over 2 years ago. Anyone using 2.0 or 2.1 should update now to `traceId` etc so that they have a smooth transition to 3.0, instead of making 3.0 also support them.

making:
@adriancole I was talking about the change from
 `%5p [${spring.zipkin.service.name:${spring.application.name:-}},%X{traceId:-},%X{spanId:-},%X{spanExportable:-}]` to `%5p [${spring.zipkin.service.name:${spring.application.name:-}},%X{traceId:-},%X{spanId:-}]` .

adriancole:
@making so you mean sleuth's default `logging.pattern.level` property when the user doesnt' set one (ex in `TraceEnvironmentPostProcessor`)?

It effectively sets this now:
```properties
logging.pattern.level=%5p [${spring.zipkin.service.name:${spring.application.name:}},%X{traceId:-},%X{spanId:-}]
```

We can revise the instructions like this?

If you have a log parsing (Grok) config that expects the sampled (aka spanExportable) field, you can minimally set the position you were using to ',-', which retains the old format even if not populating it:
```properties
logging.pattern.level=%5p [${spring.zipkin.service.name:${spring.application.name:}},%X{traceId:-},%X{spanId:-},-]
```

If you actually read the sampled field in your log service, you can populate it like so:

```properties
logging.pattern.level=%5p [${spring.zipkin.service.name:${spring.application.name:}},%X{traceId:-},%X{spanId:-},%X{sampled:-}]
```
And add this to your java config
```
@Bean CorrelationScopeCustomizer addSampled() {
  return b -> b.add(SingleCorrelationField.create(BaggageFields.SAMPLED));
}
```

wdyt?

adriancole:
It wouldn't be end of the world to re-add the sampled flag, just want to make sure it is pulling its weight.. can you mention if people are actually filtering based on this? If so, we can add it eventhough it is too late for Greenwich.M1 which went out yesterday.

I think in general it is bad practice to rely on implicit log format properties, anyway don't you agree? users should set `logging.pattern.level` log parsing pipeline relies on. We can clarify this in the README for next time.

adriancole:
@making for example, if they set `logging.pattern.level` directly, lack of that field would coerse to ',-' anyway which would mean no break at all..

making:
@adriancole 
The instruction you gave in https://github.com/spring-cloud/spring-cloud-sleuth/pull/1606#issuecomment-610712477 looks nice. That should be enough as a workaround for users who don't want to change Grok pattern.
I just wanted to avoid a surprise that logstash fails to parse after updating sleuth.

adriancole:
@making I'm very grateful for your feedback. the migration instructions are so much better now!

adriancole:
@making PTAL? https://github.com/spring-cloud/spring-cloud-sleuth/wiki/Spring-Cloud-Sleuth-3.0-Migration-Guide#parentid-and-sampled-spanexportable-mdc-fields-names-are-no-longer-set

making:
@adriancole Looks perfect, except that the following code is not highlighted ;)
image

adriancole:
good catch.. done!

PR #1607, opened 2020-04-07
Stops propagating Sleuth 1.x spring-messaging headers

adriancole:
To reduce confusion and overhead, the following custom spring-messaging headers added in Sleuth 1.0 are no longer sent, and a log warning is issued once if they are by outside code.

* spanId
* spanSampled
* spanParentSpanId
* spanTraceId
* spanFlags

Sending the above headers actually increases the headers by up to 10 because they are duplicated in the "native" part of messages. This overhead is extreme especially if messages never leave the process.

The solution is to only send [b3 single format](https://github.com/openzipkin/b3-propagation#single-header), which has been in sleuth since 2.0 and is compatible with JMS. The B3 single format is always parsed and takes precedence, even if multiple headers are sent, so this is a safe change.

Note: Unlike RPC, messaging spans never join with their parent. Better performance is achieved by not propagating the producer's parentId downstream.

Note: Deprecated spring-messaging headers such "spanTraceId" as are still read in Sleuth 3.x. However, they will not be at some point in the future. Please pay attention to the log messages and update your code if you are accidentally using them.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -132,13 +132,6 @@
 
 	<dependencyManagement>
 		<dependencies>
-			<!-- force latest reactor-netty -->
-			<dependency>
-				<groupId>io.projectreactor.netty</groupId>
-				<artifactId>reactor-netty</artifactId>
-				<version>0.9.6.RELEASE</version>
-				<optional>true</optional>
-			</dependency>
 			<dependency>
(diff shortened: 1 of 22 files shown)


adriancole:
Marked as breaking change for attention eventhough we still parse the old headers.

adriancole:

Issue #1610, opened 2020-04-07
For LoadBalancerFeignClient extensions, the calls fails due to "double load-balancing"


OlgaMaciaszek:
The issue is in the following code: 

https://github.com/spring-cloud/spring-cloud-sleuth/blob/37645fbf9d3e50dc341f333cf8bfd35189594798/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/TraceFeignObjectWrapper.java#L85-L100

Instead of passing the delegate of the load-balanced client as the first argument of `TraceLoadBalancerFeignClient` (which itself implements `LoadBalancerFeignClient`), the actual load-balanced client is passed.

This results in the call being load-balanced first by the `TraceLoadBalancerFeignClient` `super.execute(request, options)` call, which already resolves the `clientId` to a correct host IP and then again by that load-balanced client passed as the delegate, which fails because it cannot resolve a host IP from what is already a host IP.

PR #1612, opened 2020-04-07
Fixes LoadBalancerFeignClient double loadbalancing

OlgaMaciaszek:


OlgaMaciaszek:
Fixes gh-1610.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/TraceLoadBalancerFeignClient.java
```diff
@@ -95,6 +101,11 @@ public Response execute(Request request, Request.Options options) throws IOExcep
 		}
 	}
 
+	private boolean delegateIsALoadBalancer() {
+		return getDelegate() instanceof LoadBalancerFeignClient
```
spencergibb, line 105: Ug. I guess there's nothing better. :man_shrugging:

OlgaMaciaszek: Yes, not sure we can do it any better way. In `master` we will only have to check against one class, at least.

adriancole:
LGTM

PR #1614, opened 2020-04-09
Overhauls documentation to focus on what sleuth does

### LINKED ISSUE #1466: Update docs to mention brave and slf4j earlier
In the first sentence of this project description https://github.com/spring-cloud/spring-cloud-sleuth#spring-cloud-sleuth is a bit of a "lie".

Spring cloud sleuth is only available for "Brave" client users so that needs to be corrected and a proper description that states exactly that should be used:
```
Spring Cloud Sleuth is a distributed tracing plugin Spring Cloud that works only for Zipkin Brave users.
```

adriancole:
While some overview is important, the direct implementation details of tracing
not only drifted since Sleuth 1.x, but also caused a complaint #1466

This deletes most documentation that applies to an abstraction lower than sleuth
and reworks existing docs to focus on features it enables. By doing so, we have
less to maintain and users will have a clearer idea about the relationship
between Sleuth, Zipkin and Brave. Also important is this uses updated screen
shots and is in general less cluttered than before.

Note: the relationship between Sleuth and Brave was already redone in #1603
Note: this also removes the pws app which is usually broken or out of date.

Fixes #1466


adriancole:
github is not perfect at rendering, but here's how to see the new README https://github.com/spring-cloud/spring-cloud-sleuth/blob/big-doc/README.adoc#14-request-scoped-properties-baggage

adriancole:
gonna merge this as with marcin out I don't think anyone will look at it. we can revise from here!

PR #1619, opened 2020-04-23
Restores special case that ignores sampling properties

adriancole:
Sleuth is unlike most tracing configuration libraries, as it has a
legacy from 1.x as being primarily log correlation. In short, when
Zipkin is not installed, it ignored the sampling properties. This
behavior was managed implicitly through an untested combination of
configuration conventions between the core and zipkin modules.

16fb8e3 broke this and this change puts it back, in a way not tightly
coupled to Zipkin and neither requires the more simple, but confusing
"import SamplerAutoConfiguration" approach. It also backfills tests
that should have broken earlier.

In practice, a site that only uses logging is likely uniform in that, so
whether or not the sampled bit is set is of no consequence. However,
there's a chance that someone might rely on this historical behavior.

We should follow-up on 3.0 and remove this as it is very unintuitive to
intentionally ignore sampling properties. For now, this restores the old
behavior based on heuristics of bean definitions.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/SamplerAutoConfiguration.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/SamplerAutoConfiguration.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.sleuth.sampler;
 
+import brave.sampler.CountingSampler;
 import brave.sampler.Sampler;
 
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
@@ -23,12 +24,14 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.cloud.context.config.annotation.RefreshScope;
 import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Conditional;
 import org.springframework.context.annotation.Configuration;
 
 /**
  * {@linkplain Configuration configuration} for {@link Sampler}.
  *
  * @author Marcin Grzejszczak
+ * @see SamplerCondition
  * @since 2.1.0
  */
 @Configuration(proxyBeanMethods = false)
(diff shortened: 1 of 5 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/SamplerAutoConfiguration.java
```diff
@@ -40,39 +43,48 @@
 
 	@Bean
 	@ConditionalOnMissingBean
-	Sampler sleuthTraceSampler() {
+	Sampler sampler() {
```
spencergibb, line 46: We've been meeting descriptive on names so there isn't collisions with user bean names

adriancole: sure thing. I'll put a comment. We do have places where we intentionally use fixed names, and there's no use case for multiple beans. Do you know why it would matter in this case when it is ConditionalOnMissingBean?

adriancole: no worries if you don't. the comment is probably good enough. Thanks!

adriancole:
I tested this manually in wavefront by commenting out `@Import(SamplerAutoConfiguration.class)` in https://github.com/wavefrontHQ/wavefront-spring-boot/pull/22

then running https://github.com/openzipkin/sleuth-webmvc-example/compare/wavefront (except against 2.2.x snapshot)

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/sampler/SamplerAutoConfiguration.java
adriancole: I undid the comment as the bean names weren't consistent here either. wild guess is we either meant to be consistent, but accidentally weren't, or we meant not to be consistent in order to figure out possibly which autoconfiguration branch led to that bean.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfigurationTests.java
```diff
@@ -68,8 +71,48 @@ public void should_apply_in_memory_metrics_when_meter_registry_class_missing() {
...
 	}
 
+	/**
+	 * Duplicates
+	 * {@link org.springframework.cloud.sleuth.sampler.SamplerAutoConfigurationTests}
+	 * intentionally, to ensure configuration condition bugs do not exist.
+	 */
 	@Test
-	public void should_use_B3Propagation_factory_if_no_have_any_config() {
+	public void should_use_NEVER_SAMPLER_when_only_logging() {
```
marcingrzejszczak, line 80: If we have never sampler, only logging but we want the e.g. baggage to be propagated, won't we have a problem?

PR #1622, opened 2020-04-30
replace method for deprecation and keep reference of requestTemplate

HashZhang:
there are some integration codes in out project making use of the requestTemplate, but after enable sleuth the reference is null. We found the reason here and the method of building the request is depracated


adriancole:
can you check which version of feign this sets minimum to? we should add that to the release notes.

also is there a unit test that can show the template saved is usable?

OlgaMaciaszek:
@adriancole - the OpenFeign version would be `10.7.2`. It comes as a transitive dependency here from Spring Cloud OpenFeign, anyway.

@HashZhang Thanks for the PR. Could you please add a test as @adriancole has suggested.

HashZhang:
Added
@adriancole 
@OlgaMaciaszek

Issue #966, opened 2018-05-01
Strange "Uncaught exception thrown" when including Zipkin starter


nickcodefresh:
If I include this dependency in my Spring Boot 1.5.9.RELEASE / Spring Cloud Edgware.SR3 project:
```
    <dependency>
      <groupId>org.springframework.cloud</groupId>
      <artifactId>spring-cloud-starter-zipkin</artifactId>
    </dependency>
```
I get lots of Zipkin-related errors when calling my controllers, despite not having Zipkin enabled:
```
flex-fastobject-service_1      | 2018-05-01 14:24:52 ERROR [flex-fastobject-service,7f747335b9f97a1c,ad666e2b3698bc70,false] 1 ---[o-18128-exec-89]o.s.c.sleuth.instrument.web.TraceFilter  : Uncaught exception thrown
flex-fastobject-service_1      | org.apache.catalina.connector.ClientAbortException: java.io.IOException: Broken pipe
flex-fastobject-service_1      | 	at org.apache.catalina.connector.OutputBuffer.doFlush(OutputBuffer.java:321)
flex-fastobject-service_1      | 	at org.apache.catalina.connector.OutputBuffer.flush(OutputBuffer.java:284)
flex-fastobject-service_1      | 	at org.apache.catalina.connector.CoyoteOutputStream.flush(CoyoteOutputStream.java:118)
flex-fastobject-service_1      | 	at org.springframework.cloud.sleuth.instrument.web.TraceServletOutputStream.flush(TraceServletOutputStream.java:128)
flex-fastobject-service_1      | 	at com.fasterxml.jackson.core.json.UTF8JsonGenerator.flush(UTF8JsonGenerator.java:1054)
flex-fastobject-service_1      | 	at com.fasterxml.jackson.databind.ObjectMapper.writeValue(ObjectMapper.java:2511)
flex-fastobject-service_1      | 	at com.fasterxml.jackson.core.base.GeneratorBase.writeObject(GeneratorBase.java:378)
flex-fastobject-service_1      | 	at com.fasterxml.jackson.databind.node.POJONode.serialize(POJONode.java:112)
flex-fastobject-service_1      | 	at com.fasterxml.jackson.databind.node.ObjectNode.serialize(ObjectNode.java:304)
flex-fastobject-service_1      | 	at com.fasterxml.jackson.databind.node.ObjectNode.serialize(ObjectNode.java:304)
[...]

Issue #902, opened 2018-03-15
Log and Throw in TraceFilter


oburgosm:
Affected Versions: 1.3.X

Component: TraceFilter

In TraceFilter, there is a Log and throw:

```java
...
try {
			spanFromRequest = createSpan(request, skip, spanFromRequest, name);
			filterChain.doFilter(request, new TraceHttpServletResponse(response, spanFromRequest));
} catch (Throwable e) {
			exception = e;
			errorParser().parseErrorTags(tracer().getCurrentSpan(), e);
			if (log.isErrorEnabled()) {
				log.error("Uncaught exception thrown", e);
			}
			throw e;
} 
...
```

I think it would be better remove it, or, at least, to put it at DEBUG level.

Thanks

marcingrzejszczak:
I disagree. It's put there deliberately to log uncaught exceptions. If you don't like it, you can register your own version of trace filter and not log it. You can also disable logging of trace filter.

kosciej:
@marcingrzejszczak Could we rethink it again? Current approach pollutes logs with ERROR messages, and proposed solution either adds maintenance burden (custom version of TraceFilter) or forces weird configuration setting (turning OFF all logs for given class).

Other well behaving Filters in the same situation don't log exceptions coming from downstream filters as ERROR. I think no Filter in `org.springframework.web.filter` does that.

This filter re-throws incoming error, so it's eventually caught in the call stack of the request, either by the container, or other filter which can do something meaningful to the exception.

marcingrzejszczak:
Hmm I think you're using some ancient version of Sleuth cause in the latest ones you can just set `spring.sleuth.web.exception-logging-filter-enabled` to `false`.

adriancole:
https://github.com/spring-cloud/spring-cloud-sleuth/pull/1633

Issue #859, opened 2018-02-14
New SPAN created for handling on `/error` when request does not have span header


brenuart:
Uncaught exceptions are catched by Tomcat and forwarded to `/error` for handling by Spring's default `BasicErrorController`.

It appears the span logged from within the `BasicErrorHandler` (and anywhere in the handling chain after the TraceFilter) *may* not refer to the same span as from within the handler that threw the uncaught exception. It actually depends on whether the original request had span headers on it or not:
- if span headers are present and the request is parent of an ongoing trace, span stays the same in `/error`
- if no span header is present, everything appears as if two spans were created: one first the original attempt at invoking the handler, a second for the forward to `/error`.

This behaviour is observed with SpringCloud `Edgware.SR2`.
Things are different however with `Finchley.M6`: the same span is reported in both cases.

I have attached two sample applications to illustrate the case. They are built from Spring Initializr with Web and Sleuth features only. Both are the same except one is built on `Edgware.SR2` and the other on `Finchley.SR6`.
The application contains a simple RestController that logs a small message and throws a RuntimeException whenever it receives a request (mapped on `/**`).
The context also contains a custom DefaultErrorAttributes with the same behaviour as the original except it logs a message when the `errorAttributes()` method is invoked. This method is invoked after Tomcat has forwarded the request to the ErrorDispatcher after reception of an _uncaught_ exception.
The application starts on port 8080 by default.

**Scenario 1**
Invoke the service - an uncaught RuntimeException is thrown
```
curl -v http://localhost:8080
```

Application logs:

```
 INFO [-,e7f50e4c43a76f7c,e7f50e4c43a76f7c,false] 41701 --- [nio-8080-exec-2] com.example.demo.RestService             : Request received - about to throw an exception
ERROR [-,e7f50e4c43a76f7c,e7f50e4c43a76f7c,false] 41701 --- [nio-8080-exec-2] o.s.c.sleuth.instrument.web.TraceFilter  : Uncaught exception thrown

org.springframework.web.util.NestedServletException: Request processing failed; nested exception is java.lang.RuntimeException: boom
	at org.springframework.web.servlet.FrameworkServlet.processRequest(FrameworkServlet.java:982) ~[spring-webmvc-4.3.14.RELEASE.jar:4.3.14.RELEASE]
[...]

Issue #714, opened 2017-09-26
Exceptions that escape SpringMVC DispatcherServlet should be logged with trace & span IDs


jfuerth:
Hi,

We are re-raising the issue reported in #579 and #542 because we've tried the suggested solution of handling and logging all the errors ourselves, but we have yet to land on a workable solution.

We tried adding a method like this to our global ControllerAdvice bean:

```
    @ExceptionHandler(Exception.class)
    ResponseEntity<ErrorPayload> otherwiseUnhandledException(Exception e) {
        log.error("Unexpected exception occurred. Will return HTTP 500.", e);
        return new ResponseEntity<>(new ErrorPayload(e.getMessage(), ""), INTERNAL_SERVER_ERROR);
    }
```

The error is logged with the correct trace and span IDs (great!) but the killer side effect is that all the internal Spring MVC exceptions that do nice things like return 400 on missing mandatory by our ExceptionHandler method and thus return HTTP 500 to our clients.

We also experimented with registering our own `BasicErrorController` bean based on the callout box in the docs linked from #579 (https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#boot-features-error-handling). We see that it is getting instantiated, but its `onError` method is not getting called for our failing JSON API requests:

```
@Slf4j
@Controller
public class MyErrorController extends BasicErrorController {

    public MyErrorController(ErrorAttributes errorAttributes) {
        super(errorAttributes, new ErrorProperties());
        log.info("Created error controller with errorAttributes {}", errorAttributes);
    }

    @RequestMapping(produces = "application/json")
    public ErrorPayload onError() {
        log.error("KABOOM!");
        return new ErrorPayload("Something broke (TODO: resolve actual exception)", "????");
    }
}
```

We'd really appreciate any advice you can offer on how to preserve Spring MVC's default response codes while also including the Sleuth trace and span IDs in the MDC.

Thanks!

marcingrzejszczak:
Like presented in the commit message:

> without this change any exception that is not caught by a filter, get caught and logged by server's internals (e.g. Tomcat's Valve). To make the Valve log the exception together with the tracing information we would have to allow the tracing context remain in the thread after TraceFilter gets executed. That is problematic cause we're polluting the ThreadLocal and would have to assume that some component will eventually clear the context.
>
[...]

PR #1633, opened 2020-05-15
Deprecates ExceptionLoggingFilter and disables it by default

### LINKED ISSUE #1347: ExceptionLoggingFilter should be disabled by default
I think `ExceptionLoggingFilter`should be disabled by default. You usually have your own ExceptionHandler and then need to manually disable this one. Why is it enabled by default?

adriancole:
`ExceptionLoggingFilter` logs "Uncaught exception thrown" to error level
when there is a synchronous exception not otherwise swallowed. This is a
cure worse than the disease. This issue disables by default and the 3.x
should remove this, ending the years of problems it caused.

Fixes #1347

This was done IIUC to help @jfuerth who had log messages with trace IDs,
except the uncaught one, in #714 (Sept 2017). They were formerly told to
use `ExceptionHandler` and noted that didn't work in practice as it
subverted their status code logic.

Concretely, 4203566d87434e2cb65fbbb9b90fd92bac447f2e "solved" the issue
by having sleuth log all uncaught exceptions with the message
"Uncaught exception thrown". This code was initially embedded in
sleuth's and later pulled out to `ExceptionLoggingFilter` in 2.0.

A few months later @brenuart noticed this "solution" had problems. For
example, it caused the same exception to be logged twice. Bertrand also
mentioned some faults in the implementation including edge cases like
`ClientAbortException` not addressed by this. (In fact, there are even
more problems: the implementation assumes async isn't in use!)
Bertands concerns about this having surprising logging effects were
dismissed as a non-issue. Bertrand raised #859 about glitches this
caused, the status_code related ones being if statemented around, and
it went quiet a while.

A month later @oburgosm opened #902 (currently 5 thumbs up) asking to
remove the "Uncaught exception thrown" or at least set it to debug.
The response was if they don't like it, control their own tracing
filter and the issue was closed.

A month later, @nickcodefresh opened #966 confused about "Uncaught
exception thrown" messages raised by sleuth, incidentally the
`ClientAbortException` mentioned by @brenuart before. Because the
logging came from Sleuth, it appeared they were zipkin errors, and was
explained they weren't.

A year later, @T3rm1 opened #1347 to remove the filter, or at least
disable it by default. The response was first to set the Logger to OFF.
A few days later the issue was closed as the submitter was told they
were the first person to complain about this. 5 months later, another
user rallied for support.

In April 2020, #902 was re-touched by @kosciej who asked to please
reconsider, mentioning it was not typical even in Spring to do this,
for example 'org.springframework.web.filter' package. The impact of
needing to specifically change apps to OFF the sleuth logger seemed
harsh. The suggestion was instead of setting log config, to set a
property instead `spring.sleuth.web.exception-logging-filter-enabled`,
to `false` as that could also accomplish the goal of stopping
"Uncaught exception thrown".

A month later (yesterday) @MrHurniak commented again on #1347 with the
usual complaint that it is redundant and confusing.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceAutoConfiguration.java
@@ -54,6 +54,7 @@ import org.springframework.cloud.sleuth.DefaultSpanNamer;
 import org.springframework.cloud.sleuth.LocalServiceName;
 import org.springframework.cloud.sleuth.SpanAdjuster;
 import org.springframework.cloud.sleuth.SpanNamer;
+import org.springframework.cloud.sleuth.log.SleuthLogAutoConfiguration;
 import org.springframework.cloud.sleuth.sampler.SamplerAutoConfiguration;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
@@ -73,7 +74,7 @@ import org.springframework.util.StringUtils;
 @Configuration(proxyBeanMethods = false)
 @ConditionalOnProperty(value = "spring.sleuth.enabled", matchIfMissing = true)
 @EnableConfigurationProperties(SleuthProperties.class)
-@Import(SamplerAutoConfiguration.class)
+@Import({ SleuthLogAutoConfiguration.class, SamplerAutoConfiguration.class })
 public class TraceAutoConfiguration {
 
 	/**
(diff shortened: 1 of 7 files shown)


adriancole:
It may be curious what's the alternative? for those impeded by Spring Boot's recommendation about error handling here.. https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#boot-features-error-handling

they can do exactly what this does, which is add a servlet filter which logs uncaught exceptions.
```java
	@Bean
	@ConditionalOnProperty("spring.sleuth.web.exception-logging-filter-enabled")
	public FilterRegistrationBean exceptionThrowingFilter(
			SleuthWebProperties webProperties) {
		FilterRegistrationBean filterRegistrationBean = new FilterRegistrationBean(
				new ExceptionLoggingFilter());
		filterRegistrationBean.setDispatcherTypes(DispatcherType.ASYNC,
				DispatcherType.ERROR, DispatcherType.FORWARD, DispatcherType.INCLUDE,
				DispatcherType.REQUEST);
		filterRegistrationBean.setOrder(webProperties.getFilterOrder());
		return filterRegistrationBean;
	}
```

Those that don't wish to do that can know that the span will include an error regardless, and unlike the filter here, it will also be there in async cases. A future different solution could be considered, but I think we've proven over the last few years without reasonable doubt.. the current strategy of having sleuth log synchronous errors is not the right answer to the problem of uncaught exception management

adriancole:
PS I added a commit showing that you can log the exception from any framework using a `FinishedSpanHandler` (soon to be `SpanHandler` in Brave 5.12 which is next release here)

This works even with buggy stuff like webflux that frequently loses context, provided the exception hook was eventually called cc @olssonoskar

PR #1638, opened 2020-05-16
Deprecates unnecessarily public types to ease burden

adriancole:
Types like Properties and AutoConfiguration and internal utilities are
routinely marked public when they needn't be. This causes toil as we
have to preserve signatures even if they were made public by accident.

This deprecates the mass of types marked public to give some hope of
less undifferentiated toil in the future. Ideally, future change will
consider greatly if a type should be public or not as doing so haunts
maintainers.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/SleuthAsyncProperties.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/SleuthAsyncProperties.java
@@ -26,8 +26,10 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
  *
  * @author Jesus Alonso
  * @since 2.1.0
+ * @deprecated This type should have never been public and will be hidden or removed in
+ * 3.0
  */
-
+@Deprecated
 @ConfigurationProperties(prefix = "spring.sleuth.async")
 public class SleuthAsyncProperties {
 
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/circuitbreaker/SleuthCircuitBreakerAutoConfiguration.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/circuitbreaker/SleuthCircuitBreakerAutoConfiguration.java
@@ -42,7 +42,10 @@ import org.springframework.context.annotation.Configuration;
  *
  * @author Marcin Grzejszczak
  * @since 2.2.1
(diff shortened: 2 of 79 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/SleuthAsyncProperties.java
```diff
@@ -26,8 +26,10 @@
  *
  * @author Jesus Alonso
  * @since 2.1.0
+ * @deprecated This type should have never been public and will be hidden or removed in
+ * 3.0
  */
-
+@Deprecated
```
spencergibb, line 32: I don't think this should apply to configuration properties.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/circuitbreaker/SleuthCircuitBreakerAutoConfiguration.java
```diff
@@ -42,7 +42,10 @@
  *
  * @author Marcin Grzejszczak
  * @since 2.2.1
+ * @deprecated This type should have never been public and will be hidden or removed in
+ * 3.0
  */
+@Deprecated
```
spencergibb, line 48: I don't think this should apply to auto configuration classes

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/circuitbreaker/SleuthCircuitBreakerProperties.java
```diff
@@ -23,7 +23,10 @@
  *
  * @author Marcin Grzejszczak
  * @since 2.2.1
+ * @deprecated This type should have never been public and will be hidden or removed in
+ * 3.0
  */
+@Deprecated
```
spencergibb, line 29: I don't think this should apply to configuration properties.

adriancole:
@spencergibb then the solution is to just change them without deprecation? Some of these need to disappear in 3.x

adriancole:
anyway I will revert the deprecated on properties thing, 3.x breaks many classes anyway. thanks

adriancole:
thanks for the look @spencergibb I sometimes wonder if anyone ever looks at PRs here. Nice you have eyes out https://github.com/spring-cloud/spring-cloud-sleuth/pull/1639

spencergibb:
Thanks!

PR #1640, opened 2020-05-16
Pares back some deprecation after analyzing each file

adriancole:
This puts specific comments in as to why certain files that seem like they
shouldn't be public are. Notably, this includes entrypoint autoconfiguration,
which are sometimes order sensitive. In other cases there are types that were
documented (notably the async package).

This also untangles a few configuration.

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/log/SleuthSlf4jProperties.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/log/SleuthSlf4jProperties.java
@@ -25,7 +25,7 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
  * Configuration properties for slf4j.
  *
  * @author Arthur Gavlyukovskiy
- * @since 1.0.12 3.0
+ * @since 1.0.12
  */
 @ConfigurationProperties("spring.sleuth.log.slf4j")
 // TODO: Hide in 3.x, if it isn't already deleted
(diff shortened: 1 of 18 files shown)


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/log/SleuthSlf4jProperties.java
```diff
@@ -25,8 +25,7 @@
  * Configuration properties for slf4j.
  *
  * @author Arthur Gavlyukovskiy
- * @since 1.0.12
- * 3.0
+ * @since 1.0.12 3.0
```
adriancole, line 28:
```suggestion
 * @since 1.0.12
```

spring-cloud-sleuth-core/src/main/resources/META-INF/spring.factories
```diff
@@ -4,7 +4,6 @@ org.springframework.cloud.sleuth.annotation.SleuthAnnotationAutoConfiguration,\
 org.springframework.cloud.sleuth.autoconfig.TraceAutoConfiguration,\
 org.springframework.cloud.sleuth.propagation.SleuthTagPropagationAutoConfiguration,\
 org.springframework.cloud.sleuth.instrument.web.TraceHttpAutoConfiguration,\
-org.springframework.cloud.sleuth.instrument.web.TraceWebAutoConfiguration,\
```
adriancole, line 7: imported now. actually the type only does one thing which is setup skip patterns

PR #1647, opened 2020-05-18
Backports BaggagePropagation to 2.2.x

adriancole:
This allows integrations of Sleuth to use the same approach for 2.2.x as 3.x:

If you have a custom base propagation format, override the `BaggagePropagation.Factory`
bean instead of `ExtraFieldsPropagation.Factory`

Note: one subtle difference 2.2.x to 3.x is the change in the primary inject format.

2.2.x is
```java
return BaggagePropagation.newFactoryBuilder(B3Propagation.newFactoryBuilder()
                         .injectFormat(B3Propagation.Format.MULTI).build());
```

3.0 is
```java
return BaggagePropagation.newFactoryBuilder(B3Propagation.newFactoryBuilder()
                         .injectFormat(B3Propagation.Format.SINGLE_NO_PARENT).build());
```
per #1607

See https://github.com/spring-cloud/spring-cloud-gcp/issues/2268


adriancole:
cc @meltsufin  @saturnism

PR #1655, opened 2020-05-26
Uses B3 single for non-remote spans (2.2.x)

### LINKED ISSUE #1643: B3 headers warning with JMS
**Describe the bug**
When using Spring Cloud Sleuth + Spring Integration + JMS + IBM MQ, the B3 headers are propagated to the JMS headers. However, however MQ's `JMSMessage` doesn't allow these headers and throws exception, which generates the warning:

```
WARN [,5ec0b3c28f0ecf2501557752f69a8f32,0e2e4e1c88a37d95,true] 11695 --- [sub-subscriber1] o.s.i.jms.DefaultJmsHeaderMapper         : failed to map Message header 'X-B3-TraceId' to JMS property

com.ibm.msg.client.jms.DetailedMessageFormatException: JMSCC0049: The property name 'X-B3-TraceId' is not a valid Java(tm) identifier.
...
```

I read that the warning can be ignored, but need to be explicitly suppressed from the logs. If suppressing the warning level log from this class, then I might miss other warning messages.

I'm using `IntegrationFlow` with `Jms.outboundGateway`.  In the flow, I tried to use `f.headerFilter` to remove the headers, but that didn't seem to work. I suspect the headers were re-generated in the subsequent `.handle(Jms.outboundGateway()...)`

```
IntegrationFlows.from("output")
  .transform(...)
  .headerFilter(...) // remove B3 headers
  .handle(Jms.outboundGateway()...) // this still gets the header
```

To work around this, I created a new `JmsHeaderMapper` to explicitly remove the B3 header.  The logic is mostly from the `private MessageHeaders(MessageHeaders original, Set<String> keysToIgnore)` constructor, but it's private.

```
class NoB3JmsHeaderMapper extends DefaultJmsHeaderMapper {
    @Override
    public void fromHeaders(MessageHeaders headers, Message jmsMessage) {
      List<String> ignore = Arrays
          .asList("X-B3-TraceId", "X-B3-ParentSpanId", "X-B3-SpanId", "X-B3-Sampled");
      Map<String, Object> filtered = new HashMap<>(headers.size());
      headers.forEach((key, value) -> {
        if (!ignore.contains(key)) {
          filtered.put(key, value);
        }
      });
      super.fromHeaders(new MessageHeaders(filtered), jmsMessage);
    }
  }
```

Checking in to see if there are easier ways I missed.

Thanks!

adriancole:
This makes sure there cannot be JMS problems related to use of
hyphenated headers unless someone overrides the `Propagation.Factory`.

To avoid this, we set non-remote spans to B3 single format as done on
the 3.x branch here:
https://github.com/spring-cloud/spring-cloud-sleuth/pull/1607/files#diff-db43b7e91bd69d063333c20b947c902bR83-R84

Fixes #1643

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceBaggageConfiguration.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/TraceBaggageConfiguration.java
@@ -25,6 +25,7 @@ import brave.baggage.BaggagePropagation;
 import brave.baggage.BaggagePropagationConfig.SingleBaggageField;
 import brave.baggage.BaggagePropagationCustomizer;
 import brave.propagation.B3Propagation;
+import brave.propagation.B3Propagation.Format;
 import brave.propagation.ExtraFieldCustomizer;
 import brave.propagation.ExtraFieldPropagation;
 import brave.propagation.Propagation;
@@ -57,6 +58,11 @@ class TraceBaggageConfiguration {
 	static final String BAGGAGE_KEYS = "spring.sleuth.baggage-keys";
 	static final String PROPAGATION_KEYS = "spring.sleuth.propagation-keys";
 
+	// Note: Versions <2.2.3 use injectFormat(MULTI) for non-remote (ex spring-messaging)
+	// See #1643
+	static final Propagation.Factory B3_FACTORY =
+			B3Propagation.newFactoryBuilder().injectFormat(Format.SINGLE_NO_PARENT).build();
+
 	// These List<String> beans allow us to get deprecated property values, regardless of
 	// if they were comma or yaml encoded. This keeps them out of SleuthBaggageProperties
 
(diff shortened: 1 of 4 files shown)


adriancole:
oops forgot about this stuff for messaging:

```java

	static {
		LEGACY_HEADER_MAPPING.put(TRACE_ID_NAME, TraceMessageHeaders.TRACE_ID_NAME);
		LEGACY_HEADER_MAPPING.put(SPAN_ID_NAME, TraceMessageHeaders.SPAN_ID_NAME);
		LEGACY_HEADER_MAPPING.put(PARENT_SPAN_ID_NAME,
				TraceMessageHeaders.PARENT_ID_NAME);
		LEGACY_HEADER_MAPPING.put(SAMPLED_NAME, TraceMessageHeaders.SAMPLED_NAME);
		LEGACY_HEADER_MAPPING.put(FLAGS_NAME, TraceMessageHeaders.SPAN_FLAGS_NAME);
	}
```

it was counting on the old header names existing so it would re-map them to lowercase also.. I'll sort this out somehow..

adriancole:
ok so I left the legacy header mapping there as it is useful on the read side.

PR #1656, opened 2020-05-26
Fixes bug where we didn't fall back to current span

### LINKED ISSUE #1634: For @NewSpan annotation, Flux types are generating a new trace rather than a child span.
When I add _@NewSpan_ annotation on methods returning Flux, it generates a new trace all together rather than a new child span under the current trace.

Capture

adriancole:
I introduced a bug on #1525 which resulted in restarting of a trace, instead of continuing one. This fixes the bug and backfills the missing test.

Thanks @adirepo for the tip

Fixes #1634

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/ReactorSleuthMethodInvocationProcessor.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/ReactorSleuthMethodInvocationProcessor.java
@@ -146,7 +146,9 @@ class ReactorSleuthMethodInvocationProcessor
 			Span span;
 			Tracer tracer = this.processor.tracer();
 			if (this.span == null) {
-				span = tracer.newTrace();
+				// If we aren't continuing a trace from this flow, use nextSpan so that it
+				// can consider the "current span" (typically, backed by a thread-local)
+				span = tracer.nextSpan();
 				this.processor.newSpanParser().parse(this.invocation, this.newSpan, span);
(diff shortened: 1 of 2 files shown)


adriancole:
wild guess is something in here needs to be changed so that it fails before the commit https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/annotation/SleuthSpanCreatorAspectFluxTests.java

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/ReactorSleuthMethodInvocationProcessor.java
```diff
@@ -146,7 +146,9 @@ public void subscribe(CoreSubscriber<? super Object> actual) {
 			Span span;
 			Tracer tracer = this.processor.tracer();
 			if (this.span == null) {
-				span = tracer.newTrace();
+				// If we aren't continuing a trace from this flow, use nextSpan so that it
+				// can consider the "current span" (typically, backed by a thread-local)
+				span = tracer.nextSpan();
```
adriancole, line 151: Many tests in https://github.com/spring-cloud/spring-cloud-sleuth/blob/d789454147e84ece9adfaea4d994d38834881d5d/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/annotation/ReactorSleuthMethodInvocationProcessor.java hit this line, except all result in a new root trace.

PR #1665, opened 2020-06-10
Adds tracing on every factory methods in KafkaListenerContainerFactory

fromlabs:


marcingrzejszczak:
Great work @fromlabs , thank you so much!

PR #1668, opened 2020-06-20
spring-javaformat and Git on Windows

bedla:
Hi,
when authors of spring-javaformat will clarify how to use formatter (see https://github.com/spring-io/spring-javaformat/issues/202) on windows it would be nice to have this clarified in README.
What do you think?
Thx
Ivos


spencergibb:
The readme in the root is auto generated from docs/src/main/asciidoc (see the top of the file).

This likely comes from Spring Cloud build project for that section

bedla:
Ahh, I see, I have missed it. I am sorry.

PR #1687, opened 2020-07-13
Fetch the underlying beanfactory, fixes #1657

### LINKED ISSUE #1657: A new trace is created instead of a child span when using subscribeOn(...) operator.
[bug-report-1657.zip](https://github.com/spring-cloud/spring-cloud-sleuth/files/4681423/bug-report-1657.zip)

Spring Boot: 2.2.7.RELEASE
Spring Cloud: Hoxton.SR4 (spring-cloud-sleuth-core-2.2.2.RELEASE)

When using operator `subscribeOn(boundedElastic)` a new trace is created instead of spawning a child span.

Here is a short example:

```java
@RestController
public class FooController {
	
	private final Bar bar;
	
	public FooController(Bar bar) {
		this.bar = bar;
	}

	@GetMapping("/foo")
	public String foo() {
		return Mono.fromCallable(bar::bar)
		.subscribeOn(Schedulers.boundedElastic())
		.block();
	}
}

@Component
public class Bar {
	
	Tracer tracer;
	
	public Bar(Tracer tracer) {
		this.tracer = tracer;
	}

	@NewSpan
	public String bar() {
		return tracer.currentSpan().context().toString();
	}
}

```
When running such an application on localhost:8080 then when invoking http://localhost:8080/foo the response will always be `parentId/spanId` where `parentId==spanId` which indicate new trace. But it is expected it returns different parentId from spanId as it would do if there was no `subscribeOn(Schedulers.boundedElastic())` operator in the `FooController.foo()` method.

I identified the problem being at line 55 of  [SleuthContextListener](https://github.com/spring-cloud/spring-cloud-sleuth/blob/c80e1450652d285a1dd70dbe6c03a6d246ce64ad/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/SleuthContextListener.java#L55) which sometimes creates new instance of `SleuthContextListener` even when it should return an existing instance. The new instance always reports the context hasn't been refreshed. In certain cases the method will not find the existing instance in the `CACHE` because it was put to the cache on line 75 as
```java
if (context instanceof ConfigurableApplicationContext) {
  beanFactory = ((ConfigurableApplicationContext) context).getBeanFactory();
}
``` 
The method `getBean(BeanFactory)` should also look like:

```java
static SleuthContextListener getBean(BeanFactory beanFactory) {
  if (beanFactory instanceof ConfigurableApplicationContext) {
    beanFactory = ((ConfigurableApplicationContext) beanFactory).getBeanFactory();
  }
  return CACHE.getOrDefault(beanFactory, new SleuthContextListener());
}
```

This incorrect behavior causes [ContextUtil.isContextUnusable(BeanFactory)](https://github.com/spring-cloud/spring-cloud-sleuth/blob/c80e1450652d285a1dd70dbe6c03a6d246ce64ad/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/ContextUtil.java#L38) to report the context is unusable even though the context was refreshed and is not closed which changes behavior in many places. For example the [TraceableScheduledExecutorService.schedule](https://github.com/spring-cloud/spring-cloud-sleuth/blob/c80e1450652d285a1dd70dbe6c03a6d246ce64ad/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/async/TraceableScheduledExecutorService.java#L49) method at line 49 where it instead of scheduling the `TraceRunnable` schedules the raw `Runnable` which causes a new trace is created instead of creating a child span.

[bug-report-1657.zip](https://github.com/spring-cloud/spring-cloud-sleuth/files/4681423/bug-report-1657.zip)

JBodkin:


marcingrzejszczak:
Hey! Are you able to write a test that fails without this change?

JBodkin:
I've pushed a couple of tests, one shows that the current behaviour would fail without the fix

marcingrzejszczak:
@JBodkin - thanks a lot!

Issue #1716, opened 2020-08-14
HTTP client integration docs does not mention WebClient.Builder beans


ciscoo:
[Section 13.6.3](https://docs.spring.io/spring-cloud-sleuth/docs/current-SNAPSHOT/reference/html/#webclient) for `WebClient` has a callout for `WebClient` beans only.

This is somewhat misleading as it conveys that only `WebClient` beans are supported.

Examining the source of `TraceWebClientBeanPostProcessor` reveals that a [`WebClient.Builder` beans are also supported](https://github.com/spring-cloud/spring-cloud-sleuth/blob/391059f824e200953634048aee0b55bfc8c82bab/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientBeanPostProcessor.java#L80).

marcingrzejszczak:
Indeed. Do you want to find a pull request with a fix to the docs at the 2.2.x branch?

ciscoo:
Yes, I can submit a PR sometime today.

marcingrzejszczak:
Closed via https://github.com/spring-cloud/spring-cloud-sleuth/pull/1718

PR #1718, opened 2020-08-14
Include WebClient.Builder in HTTP Client Integration docs

ciscoo:
* Added `WebClient.Builder` as a supported bean
* Updated verbiage since a `WebClient` can only be created by using one of the various factory methods and a `WebClient.Builder` can be created either by `new` or various factory methods.

Fixes https://github.com/spring-cloud/spring-cloud-sleuth/issues/1716


marcingrzejszczak:
Thank you for this contritbution @ciscoo !

Issue #1679, opened 2020-07-07
Regression - Dependency cycle between HttpTracing and actuator endpoint


sguillope:
**Describe the bug**
Spring Cloud: Hoxton.SR6
Spring Boot: 2.3.1

I've previously reported this issue for an older version in #1420 which is was fixed in 2.1.3 but seems to have come back in 2.2.3 (Hoxton.SR5). This is not reproducible when using 2.2.2 (Hoxton.SR4).

The symptoms are exactly the same as before. When a custom actuator endpoint (i.e. `@Endpoint` or `@RestControllerEndpoint`-annotated component) depends (directly or transitively) on `HttpTracing`, Spring fails to start and reports a dependency cycle.

**Sample**
[sguillope/sleuth-http-tracing-endpoint](https://github.com/sguillope/sleuth-http-tracing-endpoint) demonstrates the issue by just starting the app. It's as simple as having this piece of code:
```java
@SpringBootApplication
public class Application {
    public static void main(String[] args) {
        SpringApplication.run(Application.class, args);
    }

    @Endpoint(id = "my-endpoint")
    @Component
    public static class MyEndpoint {
        public MyEndpoint(HttpTracing httpTracing) {
            // do something with httpTracing
        }
    }
}
```

Here's the startup output:
```
***************************
APPLICATION FAILED TO START
***************************

Description:

The dependencies of some of the beans in the application context form a cycle:

   tracingFilter defined in class path resource [org/springframework/cloud/sleuth/instrument/web/TraceWebServletAutoConfiguration.class]
┌─────┐
|  httpTracing defined in class path resource [org/springframework/cloud/sleuth/instrument/web/TraceHttpAutoConfiguration.class]
↑     ↓
|  sleuthSkipPatternProvider defined in class path resource [org/springframework/cloud/sleuth/instrument/web/TraceWebAutoConfiguration.class]
↑     ↓
|  application.MyEndpoint defined in file [/Users/sguillope/Projects/sleuth-http-tracing-endpoint/target/classes/com/example/sleuthhttptracingendpoint/Application$MyEndpoint.class]
└─────┘
```
and here's the full exception when enabling debug output:
```
org.springframework.beans.factory.BeanCurrentlyInCreationException: Error creating bean with name 'httpTracing': Requested bean is currently in creation: Is there an unresolvable circular reference?
[...]

PR #1722, opened 2020-08-28
Broaden the catched exception for skipPattern calculation

juliojgd:
Relates to  #1679


marcingrzejszczak (approved):

oburgosm:
Because _BeanCreationException_ is too generic, it is dangerous to catch it and continue. I think it would be better always trace the exception (DEBUG could be an adequate level), and only proceed if the exception is a wrapper for _BeanCurrentlyInCreationException_ or directly a _BeanCurrentlyInCreationException_., throwing exception in other case.

Issue #1717, opened 2020-08-14
Provide a simple way to disable WebClient instrumentation


asibross:
I am looking to disable the `TraceWebClientBeanPostProcessor` since I have my own simple brave instrumentation for WebClient.

I cannot just exclude it because it is being created in a nested configuration in [TraceWebClientAutoConfiguration](https://github.com/spring-cloud/spring-cloud-sleuth/blob/master/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceWebClientAutoConfiguration.java#L136) and the TraceWebClientAutoConfiguration is also in charge of providing other instrumentations (RestTemplate) that I don't want to exclude.

I had to implement a `BeanDefinitionRegistryPostProcessor` to remove it, but would have preferred to set a property, or add the AutoConfiguration to the exclude list.

I think adding a @ConditionalOnProperty on the nested configuration will suffice.

marcingrzejszczak:
Are you willing to add a pull request to the master branch? We can make it `@ConditionalOnProperty` indeed.

marcingrzejszczak:
Fixed via https://github.com/spring-cloud/spring-cloud-sleuth/pull/1723

PR #1723, opened 2020-08-28
Make the WebClient instrumentation conditional on property

asibross:
#1717


marcingrzejszczak (changes requested):
Do we have a configuration property annotated class for this? If not we should add this to additional properties json. Also we could add a test for this

marcingrzejszczak:
Thanks for the pr!

PR #1734, opened 2020-09-07
 Ensure that exception is handle only if there is a cycle on sleuthSkipPatternProvider

oburgosm:
Relates to #1679


pivotal-issuemaster:
@oburgosm Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-cloud/spring-cloud-sleuth&pullRequestId=1734)!

marcingrzejszczak:
Can we have a test for that please?

oburgosm:
Not enough with existing test `EndpointWithCyclicDependenciesTests`?

marcingrzejszczak:
I mean we changed production code, no tests needed to be changed or added and the code still works. So how can we know that the new feature works fine?

PR #1774, opened 2020-11-13
WIP

marcingrzejszczak:


spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/autoconfig/OnBraveEnabled.java
```diff
@@ -0,0 +1,60 @@
...
+@Target({ ElementType.TYPE, ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@ConditionalOnProperty(value = "spring.sleuth.enabled", matchIfMissing = true)
+@Conditional(AnyTracerModePropertySetCondition.class)
+public @interface OnBraveEnabled {
+
+}
+
+final class AnyTracerModePropertySetCondition extends AnyNestedCondition {
```
snicoll, line 44: I still think this should be a condition with code rather than putting those annotations around.

spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/autoconfig/TraceBraveAutoConfiguration.java
```diff
@@ -59,13 +59,12 @@
...
+@OnBraveEnabled
+@EnableConfigurationProperties({ SleuthProperties.class, SleuthSpanFilterProperties.class,
+		SleuthBaggageProperties.class, SleuthTracerProperties.class, SleuthBaggageProperties.class })
+@Import({ BraveBridgeConfiguration.class, BraveBaggageConfiguration.class, SamplerConfiguration.class,
+		BraveHttpConfiguration.class })
 @AutoConfigureBefore(TraceAutoConfiguration.class)
-@AutoConfigureAfter(TraceBravePropagationAutoConfiguration.class)
-// public allows @AutoConfigureAfter(TraceAutoConfiguration)
-// for components needing Tracing
 public class TraceBraveAutoConfiguration {
```
snicoll, line 68: Curious why this can't be called `BraveAutoConfiguration`. Isn't Brave unique enough (with its package) to refer to tracing?

spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/autoconfig/TraceBraveAutoConfiguration.java
```diff
@@ -83,84 +82,78 @@
...
-			@Nullable List<SpanHandler> spanHandlers, @Nullable List<TracingCustomizer> tracingCustomizers) {
-		Tracing.Builder builder = Tracing.newBuilder().sampler(sampler)
-				.localServiceName(StringUtils.isEmpty(serviceName) ? DEFAULT_SERVICE_NAME : serviceName)
-				.propagationFactory(factory).currentTraceContext(currentTraceContext)
-				.traceId128Bit(sleuthProperties.isTraceId128()).supportsJoin(sleuthProperties.isSupportsJoin());
-		if (spanHandlers != null) {
-			for (SpanHandler spanHandlerFactory : spanHandlers) {
-				builder.addSpanHandler(spanHandlerFactory);
+	@Configuration(proxyBeanMethods = false)
+	static class BraveConfiguration {
```
snicoll, line 86: Why do you need a nested class here? It doesn't look necessary considering there aren't any condition on it.

spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/autoconfig/TraceBraveAutoConfiguration.java
```diff
@@ -83,84 +82,78 @@
...
+
+		@Bean(name = TRACING_BEAN_NAME)
+		@ConditionalOnMissingBean
+		// NOTE: stable bean name as might be used outside sleuth
+		Tracing tracing(@LocalServiceName String serviceName, Propagation.Factory factory,
+				CurrentTraceContext currentTraceContext, Sampler sampler, SleuthProperties sleuthProperties,
+				@Nullable List<SpanHandler> spanHandlers, @Nullable List<TracingCustomizer> tracingCustomizers) {
```
snicoll, line 93: Curious what this `@Nullable` is for. First of all, framework will never inject a `null` list but rather an empty list so the contract seems wrong to me.

I'd use `ObjectProvider` for this who gives a more precise contract of what you're after, especially since it provides a `Stream` and you're lopping over the values to do something anyway.

spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/autoconfig/TraceBraveAutoConfiguration.java
```diff
@@ -83,84 +82,78 @@
 	 */
 	public static final String DEFAULT_SERVICE_NAME = "default";
```
snicoll, line 83: Are these 3 constants brave specific? They don't seem to. If they don't, it feels odd to have public constants in an implementation auto-config.

spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/autoconfig/BraveBaggageConfiguration.java
```diff
@@ -64,10 +63,9 @@
  * @since 2.0.0
  */
 @Configuration(proxyBeanMethods = false)
-@EnableConfigurationProperties(SleuthBaggageProperties.class)
-class TraceBaggageConfiguration {
+class BraveBaggageConfiguration {
```
snicoll, line 66: I think the naming could use some polishing. Stuff start with `Trace` others with `Brave` and yet others with `TraceBrave`.

snicoll: Not related to this PR but GitHub won't let me add a comment on a unchanged line. You can't put a `ConditionalOnClass` on a method. In this particular case it works because the type is not part of the signature but's a fragile arrangement (see `correlationScopeDecoratorBuilder`).

spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/bridge/BraveBridgeConfiguration.java
```diff
@@ -0,0 +1,82 @@
...
+import org.springframework.cloud.sleuth.autoconfig.SleuthBaggageProperties;
+import org.springframework.cloud.sleuth.brave.propagation.PropagationFactorySupplier;
+import org.springframework.cloud.sleuth.brave.propagation.SleuthPropagationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.lang.Nullable;
+
+@Configuration(proxyBeanMethods = false)
+@EnableConfigurationProperties(SleuthPropagationProperties.class)
+public class BraveBridgeConfiguration {
```
snicoll, line 39: This is auto-configuration stuff. If you have an auto configuration package (you do) it should sit there.

spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/bridge/BraveHttpBridgeConfiguration.java
```diff
@@ -14,35 +14,17 @@
...
- * @author Marcin Grzejszczak
- * @since 3.0.0
- */
 @Configuration(proxyBeanMethods = false)
-@ConditionalOnProperty(value = "spring.sleuth.enabled", matchIfMissing = true)
-@ConditionalOnBean({ Tracer.class, HttpTracing.class })
-@AutoConfigureBefore(TraceHttpAutoConfiguration.class)
-@AutoConfigureAfter(TraceBraveBridgeAutoConfiguation.class)
-public class TraceBraveHttpBridgeAutoConfiguration {
+public class BraveHttpBridgeConfiguration {
```
snicoll, line 27: `autoconfigure` package

spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/instrument/web/BraveHttpConfiguration.java
```diff
@@ -0,0 +1,291 @@
...
+ *
+ * @author Marcin Grzejszczak
+ * @since 2.0.0
+ */
+@Configuration(proxyBeanMethods = false)
+@ConditionalOnProperty(name = { "spring.sleuth.http.enabled", "spring.sleuth.web.enabled" }, havingValue = "true",
+		matchIfMissing = true)
+@ConditionalOnClass(HttpTracing.class)
+@EnableConfigurationProperties({ SleuthWebProperties.class, SleuthHttpProperties.class })
+public class BraveHttpConfiguration {
```
snicoll, line 69: `autoconfigure` package. If you have some classes in that package and some classes where the actual infrastructure resides, this is inconsistent and won't allow you to split the two if you chose to in the future.

spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/instrument/web/BraveHttpConfiguration.java
```diff
@@ -0,0 +1,291 @@
...
+
+	@Configuration(proxyBeanMethods = false)
+	@ConditionalOnBean(Tracing.class)
+	@Import(BraveHttpBridgeConfiguration.class)
```
snicoll, line 73: This configuration class is imported only here. Why not merge the two beans in this configuration. This'll avoid exposing a public class that you may import several times by accident.

spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/instrument/web/BraveHttpConfiguration.java
```diff
@@ -0,0 +1,291 @@
...
+	static class HttpTracingConfiguration {
+
+		@Bean
+		@ConditionalOnMissingBean
+		// NOTE: stable bean name as might be used outside sleuth
+		HttpTracing httpTracing(Tracing tracing, @Nullable SkipPatternProvider provider,
+				@Nullable brave.http.HttpClientParser clientParser, @Nullable brave.http.HttpServerParser serverParser,
```
snicoll, line 80: `HttpClientParser` is deprecated. Feels odd to still use it in 3.0.

spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/instrument/web/BraveHttpConfiguration.java
```diff
@@ -0,0 +1,291 @@
...
+			else if (serverSampler == null) {
+				return skipPatternSampler;
+			}
+			else if (skipPatternSampler == null) {
+				return serverSampler;
+			}
+			return new CompositeHttpSampler(skipPatternSampler, serverSampler);
+		}
+
+		@Bean
```
snicoll, line 203: I'd avoid declaring a `@Bean` after a bunch of private infrastructure methods. `@Bean` methods should, IMO, be at the top of the configuration class.

spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/instrument/web/BraveHttpConfiguration.java
```diff
@@ -0,0 +1,291 @@
...
+	}
+
+}
+
+/**
+ * Composite Http Sampler.
+ *
+ * @author Adrian Cole
+ */
+final class CompositeHttpSampler implements SamplerFunction<brave.http.HttpRequest> {
```
snicoll, line 223: Did you mean to nest those in `BraveHttpConfiguration`? Because it doesn't at the moment and having 4 top-level classes within the same `.java` file seems a bit strange to me.

spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/instrument/web/client/TraceWebClientAutoConfiguration.java
```diff
@@ -45,7 +43,6 @@
 @Configuration(proxyBeanMethods = false)
 @SleuthWebClientEnabled
 @ConditionalOnBean(HttpTracing.class)
-@AutoConfigureAfter(TraceHttpAutoConfiguration.class)
 @AutoConfigureBefore(HttpClientConfiguration.class)
```
snicoll, line 46: An auto-configuration outside of the dedicated `autoconfigure` package makes no sense to me.

spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/sampler/SamplerConfiguration.java
```diff
@@ -40,7 +40,7 @@
 // This is not auto-configuration, but it was in the past. Leaving the name as
 // SamplerAutoConfiguration because those not using Zipkin formerly had to
 // import this directly. A less precise name is better than rev-locking code.
-public class SamplerAutoConfiguration {
+public class SamplerConfiguration {
```
snicoll, line 43: Right now the class header mentions some legacy stuff without any sort of deprecation warning. I would deprecate this and move all the code to `autoconfigure` with an `@Import` to the new class so that those who are using the deprecated  class are not affected.

spring-cloud-sleuth-brave/src/test/java/org/springframework/cloud/sleuth/brave/bridge/BraveTracerAccessor.java
```diff
@@ -0,0 +1,29 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.sleuth.brave.bridge;
+
+public final class BraveTracerAccessor {
+
+	private BraveTracerAccessor() {
+		throw new IllegalStateException("Can't instantiate a utility class");
```
snicoll, line 22: uh?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/autoconfig/SleuthTracerProperties.java
```diff
@@ -0,0 +1,70 @@
...
+
+		/**
+		 * Picks the OpenZipkin Brave tracer implementation.
+		 */
+		BRAVE,
+
+		/**
+		 * Picks the OpenTelemetry tracer implementation.
+		 */
+		OTEL
```
snicoll, line 66: This feels a bit odd to me. Technically you are listing all the implementations that Sleuth support while, at the same time, having a core module that (supposingly) doesn't know anything about said implementations.

I think what you have there is right though so, IMO, this calls for a `spring-cloud-sleuth-autconfigure` that sits on top of the infrastructure modules and provide auto-configuration for everything. Having this arrangement will also make things more explicit and less dependant on soft auto-configuration ordering.

snicoll (commented):
I've started a first batch and I don't think I should continue at this point. It's heading in the right direction with the following gotchas:

* If we have an auto configuration dedicated package (I think that's the right call) then it should be used consistently rather than having some configuration there and some configuration elsewhere
* Built on top of the previous point, I think a case could be made for the core not having any `@Configuration` at all (same for the implementation-specific modules) and then build a single `autoconfigure` module à la Spring Boot that sits on top. This would also simplify the starter things quite a bit as adding/swapping the implementation specific modules would be all that it takes to switch with a single starter
* I might be missing something but I still find strange that beans are defined to represent a noop (rather than just defining them at all and have the auto-configuration back off by the lack of the presence of such bean.

marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-sleuth/commit/76ad931b4298b18389d0a4ea739316c14682888b

PR #1783, opened 2020-11-20
fix: Incorrect method reference used in LazyTraceScheduledThreadPoolExecutor

atsharp:
I noticed while testing some functionality that one of the methods (`decorateTaskCallable`) mapped via Reflection in `LazyTraceScheduledThreadPoolExecutor` is incorrect. I believe this should be [decorateTask](https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/ScheduledThreadPoolExecutor.html#decorateTask(java.util.concurrent.Callable,%20java.util.concurrent.RunnableScheduledFuture)), as that is the method that exists in the `ScheduledThreadPoolExecutor` delegate.

Similarly, `newTaskForCallable` also appears to be mapped incorrectly. [The only argument to this method is a Callable](https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/AbstractExecutorService.html#newTaskFor-java.util.concurrent.Callable-), yet the Reflection mapper is expecting an Object as well


marcingrzejszczak:
Thanks! Maybe you can write a test to ensure that we don't make this mistake again?

atsharp:
> Thanks! Maybe you can write a test to ensure that we don't make this mistake again?

Will do!

atsharp:
@marcingrzejszczak I added pretty comprehensive tests for the class in question. Not sure how useful all of them are given that the majority of it is just straightforward delegation, but I figure it's not hurting anything 🤷‍♂️

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/async/LazyTraceScheduledThreadPoolExecutorTests.java
```diff
@@ -37,12 +100,532 @@ protected void finalize() {
...
-		BeanFactory beanFactory = BDDMockito.mock(BeanFactory.class);
+		BeanFactory beanFactory = mock(BeanFactory.class);
 
 		new LazyTraceScheduledThreadPoolExecutor(10, beanFactory, executor, null).finalize();
 
 		BDDAssertions.then(wasCalled).isFalse();
 		BDDAssertions.then(executor.isShutdown()).isFalse();
 	}
 
+	@Test
```
marcingrzejszczak, line 111: Can we separate `given`, `when` and `then` sections? I can't figure out which line is responsible for calling the actual code to test. I guess it's this one `assertThat((Future<?>) executor.decorateTask(runnable, value)).isEqualTo(future); ` ?

marcingrzejszczak (commented):
Thanks for adding the tests! I've added one small comment.

Also I'll wait with the merge cause I'm doing some pretty big changes in master. Once I'm done let's come back to this.

spring-cloud-sleuth-core/src/test/java/org/springframework/cloud/sleuth/instrument/async/LazyTraceScheduledThreadPoolExecutorTests.java
atsharp: Yep, I'll go ahead and do that. In the example you cite, that statement is both the when and then. It's verifying that the return value for that method matches the delegate's return value. That's not super clear, so let me rework those a bit

atsharp: Reworked the tests a bit. Lemme know what you think

marcingrzejszczak (approved):

marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-sleuth/commit/3c5b490a112c5bcb0dd5481bb0a7c46ce58d1f17 . Thanks a lot for the PR @atsharp !

PR #1810, opened 2020-12-17
Pad traceId with zeros in case of W3C propagation

jonatan-ivanov:


marcingrzejszczak:
That's interesting... I literally migrated this test https://github.com/open-telemetry/opentelemetry-java/blob/master/api/baggage/src/test/java/io/opentelemetry/api/baggage/propagation/W3CBaggagePropagatorTest.java to the Sleuth API. Most likely the difference is such that OTel always sets 32 bits and Brave not necessarilly.

marcingrzejszczak (approved):

adriancole:
I would recommend highly looking at https://github.com/openzipkin-contrib/brave-propagation-w3c as this is really inefficient and as no user needs to touch this code it could be made moreso.

even copy/pasta is probably fine

PR #1812, opened 2020-12-28
Fix typo in docs (index file)

kacperfronc:
While reading docs I found a small typo in index file. "propertie" instead of "properties", this commit fixes that.


jonatan-ivanov:
@kacperfronc Thanks for the fix

PR #1817, opened 2021-01-08
Set remote IP and port for servlet HTTP server traces.

csabakos:
Spring Cloud 2020.0.0 stopped setting the remote IP and port on traces coming from HTTP servlet requests. This change brings that functionality back.

### DIFF
--- a/spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/bridge/BraveHttpServerRequest.java
+++ b/spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/bridge/BraveHttpServerRequest.java
@@ -20,6 +20,8 @@ import java.net.InetSocketAddress;
 import java.util.Collection;
 import java.util.Collections;
 
+import javax.servlet.http.HttpServletRequest;
+
 import org.springframework.cloud.sleuth.Span;
 import org.springframework.cloud.sleuth.http.HttpServerRequest;
 import org.springframework.http.server.reactive.ServerHttpRequest;
@@ -130,6 +132,10 @@ class BraveHttpServerRequest implements HttpServerRequest {
 						return false;
 					}
 					return span.remoteIpAndPort(addr.getAddress().getHostAddress(), addr.getPort());
+				}
+				else if (delegate instanceof HttpServletRequest) {
+					HttpServletRequest servletRequest = (HttpServletRequest) delegate;
+					return span.remoteIpAndPort(servletRequest.getRemoteAddr(), servletRequest.getRemotePort());
 				}
 				return false;
 			}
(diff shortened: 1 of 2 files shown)


spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/bridge/BraveHttpServerRequest.java
```diff
@@ -131,6 +133,10 @@ private boolean resolveFromInetAddress(brave.Span span) {
 					}
 					return span.remoteIpAndPort(addr.getAddress().getHostAddress(), addr.getPort());
 				}
+				else if (delegate instanceof HttpServletRequest) {
+					HttpServletRequest servletRequest = (HttpServletRequest) delegate;
+					return span.remoteIpAndPort(servletRequest.getRemoteAddr(), servletRequest.getRemotePort());
```
jonatan-ivanov, line 138: Just like in the case of `ServerHttpRequest`, could you please do a null-check on `servletRequest.getRemoteAddr()`? As far as I remember, `getRemoteAddr` can return `null` in some cases, I guess that's the reason why there is a null-check above.

spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/bridge/BraveHttpServerRequest.java
```diff
@@ -131,6 +133,10 @@ private boolean resolveFromInetAddress(brave.Span span) {
 					}
 					return span.remoteIpAndPort(addr.getAddress().getHostAddress(), addr.getPort());
 				}
+				else if (delegate instanceof HttpServletRequest) {
```
jonatan-ivanov, line 136: I'm thinking that this might be just `ServletRequest`.

jonatan-ivanov (changes requested):

spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/bridge/BraveHttpServerRequest.java
csabakos: Done

spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/bridge/BraveHttpServerRequest.java
csabakos: Done

marcingrzejszczak (approved):

jonatan-ivanov (approved):

jonatan-ivanov:
@csabakos Thanks for the contribution! :)

PR #1837, opened 2021-01-30
Fixing infinite wrapping of LazyTracingFeignClient

### LINKED ISSUE #1824: Too much LazyTracing wrapping that produces a StackOverflowError when extending the LoadBalancerFeignClient
**Describe the bug**
A `StackOverflowError` is thrown after a massive usage of wrap/decorator of `LazyTracingFeignClient` delegate in the custom `LoadBalancerFeignClient`. This happens because the wrap method of `TraceFeignObjectWrapper` **modifies** the **custom** bean `LoadBalancerFeignClient`, changing its delegate with a wrap (`LazyTracingFeignClient`) of the previous delegate. Since the bean is a **singleton** and its runtime class is not `LoadBalancerFeignClient` (we extended it), the wrap method keep to wrap/decorate with a new `LazyTracingFeignClient` for each new _TracingContext_. It actually runs every new _TracingContext_ because of the aspect `TracingFeignAspect`.

image

Spring sleuth 2.2.6.RELEASE
Spring cloud Hoxton.SR9

**Sample screenshot**
In the screenshot the `XenonLoadBalancerFeignClient` is instance of `LoadBalancerFeignClient`, but of course the `AopUtils.getTargetClass` is not equal to `LoadBalancerFeignClient`. This behavior produces a _stack leak_ (do not know how to define it) that ends with a `StackOverflowError` during the execution.

image

jonatan-ivanov:
Fixes #1824

### DIFF
--- a/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/TraceFeignObjectWrapper.java
+++ b/spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/feign/TraceFeignObjectWrapper.java
@@ -83,7 +83,8 @@ final class TraceFeignObjectWrapper {
 	}
 
 	Object wrap(Object bean) {
-		if (bean instanceof Client && !(bean instanceof TracingFeignClient)) {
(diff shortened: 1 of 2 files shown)


jonatan-ivanov:
@albertominetti fyi

PR #1844, opened 2021-02-09
Add closing backtick for false keyword

rewolf:


jonatan-ivanov:
@rewolf Thanks for the fix

Issue #1843, opened 2021-02-09
Upgrade reactor and use QueueWrapper to support reactor-based instrumentation


jonatan-ivanov:
There was a new feature added to reactor-core that makes it possible to improve the reactor-based instrumentation by wrapping the queues used by reactor: https://github.com/reactor/reactor-core/pull/2566

tony-clarke-amdocs:
Forgive me, what is the benefit? Is it performance or it allows some additional functionality?

bsideup:
@tony-clarke-amdocs in theory, it should cover more cases (queues are at the core of everything that Reactor does).
Also (again, in theory) it should be more performant.

Once integrated (see #1858), we can run Sleuth's benchmarks and compare the performance with the `onEachOperator` approach.

PR #1858, opened 2021-02-24
New reactor queue wrapping

marcingrzejszczak:
Turns on the reactor hook instrumentation (https://github.com/reactor/reactor-core/pull/2566)

fixes https://github.com/spring-cloud/spring-cloud-sleuth/issues/1843

cc @bsideup


spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/TraceReactorAutoConfiguration.java
```diff
@@ -194,27 +213,122 @@ public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory)
...
+			@Override
+			public Object peek() {
+				return queue.peek();
+			}
+
+			@Override
+			@SuppressWarnings("unchecked")
+			public Iterator<Object> iterator() {
+				return (Iterator<Object>) queue.iterator();
+			}
```
bsideup, line 317: be careful with these two. They are not currently used by Reactor (I _think_), but there is no guarantee that they won't be used in future, and it may happen that the object offered to this queue won't be the same as `peek`ed, so I would highly recommend *at least* adding the unwrapping to these

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHandlerFunctionAdapterBeanPostProcessor.java
```diff
@@ -0,0 +1,117 @@
...
+		private TraceHandlerFunction(HandlerFunction<?> delegate,
+				BeanFactory beanFactory) {
+			this.delegate = delegate;
+			this.beanFactory = beanFactory;
+		}
+
+		@Override
+		public Mono<?> handle(ServerRequest serverRequest) {
+			AtomicReference<CurrentTraceContext.Scope> scope = new AtomicReference<>();
+			return Mono.just(scope).doFirst(() -> serverRequest
```
bsideup, line 95: nit: `Mono.just` + `doFirst` can be replaced with `Mono.fromCallable`

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/TraceReactorAutoConfiguration.java
```diff
@@ -194,27 +213,122 @@ public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory)
...
+	private static Queue<?> traceQueue(ConfigurableApplicationContext springContext,
+			Queue<?> queue) {
+		if (!springContext.isActive()) {
+			return queue;
+		}
+		CurrentTraceContext currentTraceContext = springContext
+				.getBean(CurrentTraceContext.class);
+		@SuppressWarnings("unchecked")
+		Queue<Envelope> envelopeQueue = (Queue<Envelope>) queue;
+		return new AbstractQueue<Object>() {
```
jonatan-ivanov, line 282: Would it make sense to move this into a separate, package-private class (inject the `currentTraceContext`)?
That way this would be testable in itself and maybe a little more readable.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/TraceReactorAutoConfiguration.java
jonatan-ivanov: +1

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHandlerFunctionAdapterBeanPostProcessor.java
```diff
@@ -0,0 +1,117 @@
+/*
+ * Copyright 2013-2019 the original author or authors.
```
jonatan-ivanov, line 2: 2021?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHandlerFunctionAdapterBeanPostProcessor.java
```diff
@@ -0,0 +1,117 @@
...
+					})).flatMap(r -> this.delegate.handle(serverRequest))
+					.doFinally(signalType -> {
+						CurrentTraceContext.Scope spanInScope = scope.get();
+						if (spanInScope != null) {
+							spanInScope.close();
+						}
+					});
+		}
+
+		private CurrentTraceContext currentTraceContext() {
```
jonatan-ivanov, line 107: There is no way this can be called concurrently, right?
@bsideup?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHandlerFunctionAdapterBeanPostProcessor.java
jonatan-ivanov, line 95: I need to confess something: I needed to copy this over vscode and format to be able to read it.
Is there anything we can do with the formatting here, to make it more readable (I'm looking at checkstyle)?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHandlerFunctionAdapterBeanPostProcessor.java
```diff
@@ -0,0 +1,117 @@
...
+				BeanFactory beanFactory) {
+			this.delegate = delegate;
+			this.beanFactory = beanFactory;
+		}
+
+		@Override
+		public Mono<?> handle(ServerRequest serverRequest) {
+			AtomicReference<CurrentTraceContext.Scope> scope = new AtomicReference<>();
+			return Mono.just(scope).doFirst(() -> serverRequest
+					.attribute(TraceWebFilter.TRACE_REQUEST_ATTR).ifPresent(o -> {
```
jonatan-ivanov, line 96: Can we rename `o` to `span`?
Also, the inner `{}` and `;` is not needed:
```java
.ifPresent(span -> scope.set(currentTraceContext().maybeScope(((Span) span).context())))
```

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHandlerFunctionAdapterBeanPostProcessor.java
jonatan-ivanov: That might also make it simpler to push down the `Optional` to the nullcheck on `spanInScope`.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebFilter.java
```diff
@@ -85,6 +86,8 @@
 
 	SleuthWebProperties webProperties;
 
+	CurrentTraceContext currentTraceContext;
```
jonatan-ivanov, line 89: private?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebFilter.java
```diff
@@ -116,6 +119,14 @@ SleuthWebProperties sleuthWebProperties() {
 		return this.webProperties;
 	}
 
+	CurrentTraceContext currentTraceContext() {
```
jonatan-ivanov, line 122: Can this be private?
Also, no concurrency, right?

jonatan-ivanov (changes requested):

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/SleuthReactorProperties.java
```diff
@@ -33,11 +33,21 @@
...
+	 */
+	private boolean decorateHooks = true;
+
 	/**
 	 * When true decorates on each operator, will be less performing, but logging will
 	 * always contain the tracing entries in each operator. When false decorates on last
 	 * operator, will be more performing, but logging might not always contain the tracing
 	 * entries.
+	 *
+	 * If {@link SleuthReactorProperties#decorateHooks} is used, this decoration mode will
```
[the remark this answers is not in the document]
marcingrzejszczak: They can be but only one will be used. Also, that would be a breaking change.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/TraceReactorAutoConfiguration.java
```diff
@@ -152,12 +164,19 @@ public void onApplicationEvent(RefreshScopeRefreshedEvent event) {
 		}
 		Hooks.resetOnEachOperator(SLEUTH_TRACE_REACTOR_KEY);
 		Hooks.resetOnLastOperator(SLEUTH_TRACE_REACTOR_KEY);
+		Hooks.removeQueueWrapper(SLEUTH_TRACE_REACTOR_KEY);
+		if (this.reactorProperties.isDecorateHooks()) {
```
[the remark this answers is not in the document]
marcingrzejszczak: yes, you're right!

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/TraceReactorAutoConfiguration.java
```diff
@@ -194,27 +213,122 @@ public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory)
...
 		ConfigurableEnvironment environment = springContext.getEnvironment();
-		boolean decorateOnEach = environment.getProperty(
-				"spring.sleuth.reactor.decorate-on-each", Boolean.class, true);
-		if (decorateOnEach) {
-			if (log.isTraceEnabled()) {
-				log.trace("Decorating onEach operator instrumentation");
-			}
-			Hooks.onEachOperator(SLEUTH_TRACE_REACTOR_KEY,
-					scopePassingSpanOperator(springContext));
+		Boolean decorateHooks = environment
```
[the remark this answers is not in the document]
marcingrzejszczak: AFAIR it's too early in the lifecycle to do that since it's a `BeanDefinitionRegistryPostProcessor`

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/TraceReactorAutoConfiguration.java
```diff
@@ -194,27 +213,122 @@ public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory)
...
+		boolean hasQueueWrapper = ReflectionUtils.findMethod(Hooks.class,
+				"addQueueWrapper", String.class, Function.class) != null;
+		if (!hasQueueWrapper && decorateHooks != null) {
+			log.warn(
+					"You have explicitly turned on the decorate hooks option but you're using an old version of Reactor. Please upgrade to the latest Boot version (at least 2.3.9.RELEASE). Will fall back to the previous reactor instrumentation mode");
 		}
 		else {
-			if (log.isTraceEnabled()) {
-				log.trace("Decorating onLast operator instrumentation");
+			decorateHooks = decorateHooks != null ? decorateHooks : Boolean.TRUE;
```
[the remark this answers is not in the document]
marcingrzejszczak: Yes, I explicitly removed the previous approach where i used that version of the method. I want to know whether the user has actually set the value. If they have not turned on the feature and have a previous version of reactor then I want to pass to the oneach and onlast versions. If they explicitly turned on and are using the old version of reactor then it's a problem and we'll fall back.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/reactor/TraceReactorAutoConfiguration.java
```diff
@@ -194,27 +213,122 @@ public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory)
...
+		}
+		if (Boolean.TRUE.equals(decorateHooks) && hasQueueWrapper) {
```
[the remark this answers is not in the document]
marcingrzejszczak: Same comment as above

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHandlerFunctionAdapterBeanPostProcessor.java
marcingrzejszczak: What's wrong with the formatting?

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebFilter.java
marcingrzejszczak: > That might also make it simpler to push down the Optional to the nullcheck on spanInScope.

I don't get it.

spring-cloud-sleuth-core/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceHandlerFunctionAdapterBeanPostProcessor.java
marcingrzejszczak: I don't understand what you mean. The `Mono.fromCallable` doesn't seem to work

marcingrzejszczak: I've added some formatter comments

PR #1859, opened 2021-02-25
Implement mechanism to apply custom api path

MazizEsa:
This comes in handy when another monitoring tool is being used, for instance new relic, which doesn't not require path in its tracing api.

Context as of feb 19, new relic doesn't accept request to its tracing api endpoint with path. Previous it was fine, now it will throw 404 erro.

### DIFF
--- a/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/ZipkinProperties.java
+++ b/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/ZipkinProperties.java
@@ -38,6 +38,13 @@ public class ZipkinProperties {
 	private String baseUrl = "http://localhost:9411/";
 
 	/**
+	 * Api path that will be appended to baseUrl above as suffix This is applicable if you
+	 * are using other monitoring tools. IE, new relic, the trace api doesn't need api
+	 * path, so this can be set to blank ("") in the config.
+	 */
+	private String apiPath = null;
+
+	/**
 	 * If set to {@code false}, will treat the {@link ZipkinProperties#baseUrl} as a URL
 	 * always.
 	 */
@@ -82,6 +89,14 @@ public class ZipkinProperties {
 
 	public void setBaseUrl(String baseUrl) {
 		this.baseUrl = baseUrl;
+	}
+
+	public String getApiPath() {
+		return apiPath;
+	}
+
+	public void setApiPath(String apiPath) {
+		this.apiPath = apiPath;
 	}
 
 	public boolean isEnabled() {
--- a/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/ZipkinProperties.java
+++ b/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/ZipkinProperties.java
@@ -38,9 +38,9 @@ public class ZipkinProperties {
 	private String baseUrl = "http://localhost:9411/";
 
 	/**
-	 * Api path that will be appended to baseUrl above as suffix This is applicable if you
-	 * are using other monitoring tools. IE, new relic, the trace api doesn't need api
-	 * path, so this can be set to blank ("") in the config.
+	 * The API path to append to baseUrl (above) as suffix. This applies if you use other
+	 * monitoring tools, such as New Relic. The trace API doesn't need the API path, so
+	 * you can set it to blank ("") in the configuration.
 	 */
 	private String apiPath = null;
 
@@ -92,7 +92,7 @@ public class ZipkinProperties {
 	}
 
 	public String getApiPath() {
-		return apiPath;
+		return this.apiPath;
 	}
 
 	public void setApiPath(String apiPath) {
--- a/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/sender/RestTemplateSender.java
+++ b/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/sender/RestTemplateSender.java
@@ -19,6 +19,7 @@ package org.springframework.cloud.sleuth.zipkin2.sender;
 import java.io.IOException;
 import java.net.URI;
 import java.util.List;
+import java.util.Objects;
 
 import zipkin2.Call;
 import zipkin2.Callback;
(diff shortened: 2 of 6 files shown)


spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/sender/RestTemplateSender.java
```diff
@@ -77,6 +81,16 @@ else if (this.encoding == Encoding.JSON) {
 		this.messageEncoder = BytesMessageEncoder.forEncoding(this.encoding);
 	}
 
+	private String buildUrlWithCustomPathIfNecessary(final String baseUrl,
+			final String customApiPath, final String defaultUrl) {
+		if (Objects.nonNull(customApiPath)) {
```
marcingrzejszczak, line 86: Actually we could call `StringUtils.hasText(...)`

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/ZipkinProperties.java
```diff
@@ -84,6 +91,14 @@ public void setBaseUrl(String baseUrl) {
 		this.baseUrl = baseUrl;
 	}
 
+	public String getApiPath() {
+		return apiPath;
```
marcingrzejszczak, line 95: `return this.apiPath` - let's be consistent

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/ZipkinProperties.java
```diff
@@ -37,6 +37,13 @@
 	 */
 	private String baseUrl = "http://localhost:9411/";
 
+	/**
```
marcingrzejszczak, line 40: cc @Buzzardo - can you please check this out?

marcingrzejszczak (changes requested):
LGTM apart from small changes. Could you also please update the documentation and mention the new `apiPath` somewhere around here https://docs.spring.io/spring-cloud-sleuth/docs/2.2.7.RELEASE/reference/html/#sending-spans-to-zipkin ? Once it's done then @Buzzardo would review it.

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/sender/RestTemplateSender.java
MazizEsa: @marcingrzejszczak , using nonNull was the intention. This is to cater cases where:

1. Null (when the apiPath is not configured at all, use the default
2. NonBlank (any text) as the custom api path
3. Blank (empty string) this is for cases no path needed at all (one example newrelic)

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/ZipkinProperties.java
MazizEsa: will include in the next push. Thanks @marcingrzejszczak

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/sender/RestTemplateSender.java
marcingrzejszczak: 1) `apiPath` should default to `api/v2/spans`  - that way we're simplifying the cases. 
2) Makes sense
3) Shouldn't it be then `/` ?

MazizEsa: @marcingrzejszczak 
We can't default to api/v2/spans in the apiPath because within RestTemplate the path is different between JSON2, PROTO and JSON. JSON is using v1. So I wanted to ensure that part of logic within the RestTemplateSender constructor being maintained.

For 3, at least for new relic case, there's no /. If you include /, tracing api will return 404. So for the path it must be empty.

marcingrzejszczak: OK, now I understand! You're right :+1:

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/ZipkinProperties.java
```diff
@@ -37,6 +37,13 @@
...
+	 * Api path that will be appended to baseUrl above as suffix This is applicable if you
```
Buzzardo, line 41: Change to:

The API path to append to the baseUrl (above) as a suffix. This applies if you

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/ZipkinProperties.java
```diff
@@ -37,6 +37,13 @@
...
+	 * are using other monitoring tools. IE, new relic, the trace api doesn't need api
```
Buzzardo, line 42: Change to:

use other monitoring tools, such as New Relic. The trace API doesn't need the API

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/ZipkinProperties.java
```diff
@@ -37,6 +37,13 @@
...
+	 * path, so this can be set to blank ("") in the config.
```
Buzzardo, line 43: Change to:

path, so you can set it to blank ("") in the configuration.

Buzzardo (changes requested):

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/sender/RestTemplateSender.java
```diff
@@ -77,6 +81,16 @@ else if (this.encoding == Encoding.JSON) {
...
+			return baseUrl + (baseUrl.endsWith("/") || customApiPath.startsWith("/")
+					|| customApiPath.isEmpty() ? "" : "/") + customApiPath;
```
jonatan-ivanov, line 88: Wouldn't the `/` handling logic be simpler by doing:
```suggestion
			return Path.of(baseUrl, customApiPath).toString();
```

MazizEsa: Will have a look.

Buzzardo (approved):
Thank you.

spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/sender/RestTemplateSender.java
MazizEsa: @jonatan-ivanov ,
Can we maintain the previous implementation. Using Path.of cause 2 issues:
1. It will be OS dependent. On Windows, it will be an exception if the customApiPath is blank (illegal character). We need blank in order to cater case for new relic at least.
2. Path.of will strip double forward slash to single. So http://something becomes http:/.

jonatan-ivanov (approved):

MazizEsa:
@marcingrzejszczak @Buzzardo made the changes as well in the main documentation https://github.com/spring-cloud/spring-cloud-sleuth/pull/1859/commits/de23b6ee929698a00aefb5e410b7023eef29531a . Thanks.

marcingrzejszczak:
Thanks @MazizEsa for your contribution!

Issue #1874, opened 2021-03-10
Baggage remote fields propagation is case sensitive


artemptushkin:
I noticed that `remote-fields` properties work in the case-sensitive way when Brave always converts it to lower case. In my case I use `TracingChannelInterceptor`, i.e. raw channel messaging

It causes a problem that this field can't be propagated because `MessageHeaderPropagation` tries to get the value from the message header by the lower-cased key.

```yaml
spring:
  sleuth:
    baggage:
      remote-fields: ['tracingContext']
```

Here you can see, that the key is *lower-cased* as it comes from Brave baggage but the header hey includes an upper-case letter.
![image](https://user-images.githubusercontent.com/22632802/110695376-18014980-81ea-11eb-9e97-966f5105c730.png)

Thus, the header can't be found
![image](https://user-images.githubusercontent.com/22632802/110695577-5ac32180-81ea-11eb-8bde-a5cc782cc1ca.png)

But if put in the expected way it returns the header value
![image](https://user-images.githubusercontent.com/22632802/110695665-72020f00-81ea-11eb-8b2c-ecbecc897e6e.png)

I would expect it works in a case-insensitive way.

If it is a bug, I'd join to fix it.

marcingrzejszczak:
Yeah it seems like a bug and I would assume to be case-insensitive. Please feel free to file a PR to fix this.

marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-sleuth/commit/39094e16145a482169ca97dd2a0496b2d0de7ab2

PR #1884, opened 2021-03-16
Getting header value case insensitive way

artemptushkin:
#1874 Make message header getter case insensitive for propagated headers

@marcingrzejszczak should we update native headers logic also? Seems like these are only _b3_, but but sure


marcingrzejszczak:
Yeah that's a good idea to check the native ones too.

artemptushkin:
Okay, will be updated then

marcingrzejszczak (approved):

marcingrzejszczak:
Thanks @artemptushkin ! You've managed to do the change just in time for the release :)

artemptushkin:
@marcingrzejszczak nice, thanks :)

Issue #1903, opened 2021-04-14
Add Spring Cloud Task support


marcingrzejszczak:
Root span would start and be reported at the point of the `TaskExecutorListener`. The name of the span would be `spring.application.name`.

Each of the `CommanLineRunner` or `ApplicationRunner` executions would be a span. The name of the span would be the bean name.

cc @mminella

cppwfs:
We can add this as a feature in 2.4.     Thoughts @mminella .

marcingrzejszczak:
We can add it in Sleuth that's not a problem (I think we even have to do it cause otherwise we could have a release cycle Task depending on Sleuth that depends on a lot of other things). Or we would have to push Task after Sleuth if Task is not used by any other projects.

I've created this issue after talking to @mminella and we're gathering all feedback here.

cppwfs:
That makes sense

marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-sleuth/commit/2fca60a112e360b6913ed7b2b9d12c768ce751ad

abccbaandy:
ok, so `ApplicationRunner or CommandLineRunner` will support in `spring cloud 2021.0.0`, and `spring boot 2.6.X` will support `spring cloud 2021.0.0`, right?

marcingrzejszczak:
that's the plan.

PR #1912, opened 2021-04-19
Added spring-cloud-task support

marcingrzejszczak:
fixes gh-1903

related sample https://github.com/spring-cloud-samples/spring-cloud-sleuth-samples/pull/5

### DIFF
--- a/spring-cloud-sleuth-instrumentation/pom.xml
+++ b/spring-cloud-sleuth-instrumentation/pom.xml
@@ -128,6 +128,11 @@
 			<optional>true</optional>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-starter-task</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
 			<groupId>org.aspectj</groupId>
 			<artifactId>aspectjrt</artifactId>
 		</dependency>
--- a/spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/task/TraceCommandLineRunner.java
+++ b/spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/task/TraceCommandLineRunner.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2018-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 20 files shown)


spring-cloud-sleuth-instrumentation/pom.xml
```diff
@@ -127,6 +127,11 @@
 			<artifactId>spring-cloud-starter-gateway</artifactId>
 			<optional>true</optional>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.cloud</groupId>
+			<artifactId>spring-cloud-starter-task</artifactId>
+			<optional>true</optional>
```
cppwfs, line 133: In this case it means that task must be built before Sleuth.  I think we are good shape on this, but just wanted to capture the thought.

spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/task/TraceCommandLineRunner.java
```diff
@@ -0,0 +1,64 @@
...
+import org.springframework.cloud.sleuth.Span;
+import org.springframework.cloud.sleuth.Tracer;
+
+/**
+ * Trace representation of a {@link CommandLineRunner}.
+ *
+ * @author Marcin Grzejszczak
+ * @since 3.1.0
+ */
+public class TraceCommandLineRunner implements CommandLineRunner {
```
cppwfs, line 30: With CommandLineRunner going away in Boot 2.6, should we only support ApplicationRunner?

spring-cloud-sleuth-instrumentation/pom.xml
marcingrzejszczak: We already require task to be built before Sleuth

spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/task/TraceCommandLineRunner.java
marcingrzejszczak: This will be merged for Boot 2.5 so I guess we can mark it as `@Deprecated`

tests/brave/spring-cloud-sleuth-instrumentation-task-tests/src/test/resources/application.yml
```diff
@@ -0,0 +1,5 @@
+logging.level.org.springframework.cloud: DEBUG
+logging.level.com.netflix.discovery.InstanceInfoReplicator: ERROR
+logging.level.org.springframework.cloud.sleuth.brave.instrument.web.client.feign: TRACE
+
+spring.sleuth.scheduled.skipPattern: "^org.*TestBeanWithScheduledMethodToBeIgnored$"
```
jonatan-ivanov, line 5: new line at the end

jonatan-ivanov (approved):

marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-sleuth/commit/2fca60a112e360b6913ed7b2b9d12c768ce751ad

PR #1906, opened 2021-04-15
Kafka instrumentation

flavium31:
Hi @marcingrzejszczak, as per our Zoom call we're raising a PR for Kafka instrumentation (including reactor kafka). Please have a look and let us know what you think.

PR #1920, opened 2021-04-21
Kafka instrumentation

flavium31:
Raised this PR as suggested by @marcingrzejszczak in #1906


marcingrzejszczak:
Congrats @flavium31 !

marcingrzejszczak:
BTW have you added the docs? Cause I don't see it. Please file it in another PR then

marcingrzejszczak:
Argh I haven't noticed that all of your `since` are from `3.0.3` - we need to fix it with `3.1.0`

flavium31:
@marcingrzejszczak thanks for feedback, we will raise another PR with the requested changes. If it sounds good to you, I can also include an autoconfiguration module in that PR (which I realized we're still missing here) along with some more integration tests to cover all decorated Kafka clients (Consumer, Producer, Receiver and Sender)

marcingrzejszczak:
Great!

Issue #1921, opened 2021-04-21
Kafka Instrumentation Issues


garyrussell:
Spring for Apache Kafka has a bug whereby the results of calling the Producer/Consumer post processors are discarded

https://github.com/spring-projects/spring-kafka/issues/1778

```
this.postProcessors.forEach(pp -> pp.apply(kafkaProducer));
```

This means that the `TraceProducerPostProcessor` is a no-op.

However, Sleuth still works because the factory is also wrapped in a proxy and the producer created by the factory is wrapped in a `TracingProducer`.

Final wrapping:

- `TracingProducer`
- - `CloseSafeProducer`
- - - `KafkaProducer`

When we fix the spring-kafka bug, we'll end up with 

- `TracingProducer`
- - `CloseSafeProducer`
- - - `TracingProducer`
- - - - `KafkaProducer`

More context:

https://stackoverflow.com/questions/67194702/sleuth-tracing-is-not-working-for-transactional-kafka-producers/67196367#67196367

garyrussell:
With spring-kafka bug fixed

![image](https://user-images.githubusercontent.com/483832/115573859-1c139300-a28f-11eb-906c-638ea1448296.png)
![image](https://user-images.githubusercontent.com/483832/115573889-22097400-a28f-11eb-9cf1-cdcfbb052cd2.png)

marcingrzejszczak:
@garyrussell would this fix the problem? https://github.com/spring-cloud/spring-cloud-sleuth/pull/1923

garyrussell:
It would fix the `Consumer` case but, yes, with the producer case we already wrap the raw producer.

[so67194702.zip](https://github.com/spring-cloud/spring-cloud-sleuth/files/6352841/so67194702.zip)

To get the double wrapping, you would need to change the pom to use spring-kafka 2.7.1-SNAPSHOT (commented out) and wait for [this build](https://build.spring.io/browse/SK-MAS-1384) to complete.

garyrussell:
Actually, in some cases, even with the consumer, we sometimes wrap it in a proxy.

PR #1923, opened 2021-04-21
Wraps producer / consumer once only

marcingrzejszczak:
without this change we can wrap the kafka consumer / producer multiple times
with this chang we're doing it only once

fixes gh-1921

### DIFF
--- a/spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/instrument/messaging/TraceConsumerPostProcessor.java
+++ b/spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/instrument/messaging/TraceConsumerPostProcessor.java
@@ -21,12 +21,16 @@ import org.apache.kafka.clients.consumer.Consumer;
 
 import org.springframework.beans.factory.BeanFactory;
 import org.springframework.kafka.core.ConsumerPostProcessor;
+import org.springframework.util.ClassUtils;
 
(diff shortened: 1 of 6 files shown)


spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/instrument/messaging/TraceConsumerPostProcessor.java
```diff
@@ -21,13 +21,17 @@
...
 import org.springframework.kafka.core.ConsumerPostProcessor;
+import org.springframework.util.ClassUtils;
 
 class TraceConsumerPostProcessor<K, V> implements ConsumerPostProcessor<K, V> {
 
 	private final BeanFactory beanFactory;
 
 	private KafkaTracing kafkaTracing;
 
+	// Because it's not public in Brave
```
jonatan-ivanov, line 32: Should we open an issue for this in Brave?

jonatan-ivanov (approved):

spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/instrument/messaging/TraceConsumerPostProcessor.java
marcingrzejszczak: Good idea, although as far as I know that's fine by design

PR #1934, opened 2021-05-04
Modify WebClient tracing to include the uri template in generated span names

jhinch:
Spans names are in the form `'<http method> <uri template>'` (e.g. `'GET /path/{id}'`). This naming strategy matches the span names generated by `TraceWebFilter`.

The improved naming strategy has not applied to `RestTemplate` as the uri template is not readily available to interceptors and `RestTemplate` is in maintenance mode. It is not entirely out of the realm of possibility of implementing as there is a solution which is used within Spring Boot for [metric instrumentation](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/client/MetricsClientHttpRequestInterceptor.java#L103) of RestTemplate by creating a shared thread local between the uri template handler and the interceptor

### DIFF
--- a/spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceExchangeFilterFunction.java
+++ b/spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/web/client/TraceExchangeFilterFunction.java
@@ -38,10 +38,12 @@ import org.springframework.cloud.sleuth.http.HttpClientRequest;
 import org.springframework.cloud.sleuth.http.HttpClientResponse;
 import org.springframework.cloud.sleuth.instrument.reactor.TraceContextPropagator;
 import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.http.HttpMethod;
 import org.springframework.web.reactive.function.client.ClientRequest;
 import org.springframework.web.reactive.function.client.ClientResponse;
 import org.springframework.web.reactive.function.client.ExchangeFilterFunction;
 import org.springframework.web.reactive.function.client.ExchangeFunction;
+import org.springframework.web.reactive.function.client.WebClient;
 
 /**
  * Trace representation of {@link ExchangeFilterFunction}.
@@ -52,6 +54,8 @@ import org.springframework.web.reactive.function.client.ExchangeFunction;
 public final class TraceExchangeFilterFunction implements ExchangeFilterFunction {
 
 	private static final Log log = LogFactory.getLog(TraceExchangeFilterFunction.class);
+
+	private static final String URI_TEMPLATE_ATTRIBUTE = WebClient.class.getName() + ".uriTemplate";
 
 	final ConfigurableApplicationContext springContext;
 
@@ -155,12 +159,18 @@ public final class TraceExchangeFilterFunction implements ExchangeFilterFunction
 
 		final CurrentTraceContext currentTraceContext;
 
+		final HttpMethod method;
+
+		final String httpRoute;
+
 		TraceWebClientSubscriber(CoreSubscriber<? super ClientResponse> actual, Context ctx, Span clientSpan,
 				TraceContext parent, MonoWebClientTrace mono) {
 			this.actual = actual;
 			this.parent = parent;
 			this.handler = mono.handler;
 			this.currentTraceContext = mono.currentTraceContext;
+			this.method = mono.request.method();
+			this.httpRoute = (String) mono.request.attribute(URI_TEMPLATE_ATTRIBUTE).orElse(null);
 			this.context = this.parent != null && !this.parent.equals(ctx.getOrDefault(TraceContext.class, null))
 					? ctx.put(TraceContext.class, this.parent) : ctx;
 			set(clientSpan);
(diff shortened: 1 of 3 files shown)


jhinch:
I think the build might be flaky. It seems that it is failing to download artifacts. Running the same command locally completes successfully for me locally.

marcingrzejszczak:
Hu @jhinch ! Thank you so much for this contribution. Could you please show a screenshot from Wavefront / Zipkin before and after this change?

jhinch:
It seems that when trying to use Zipkin UI locally, it doesn't display the client spans so I couldn't get a good screenshot. However I have captures the spans produced by `spring-cloud-sleuth-samples/spring-cloud-sleuth-sample-zipkin` before and after (modifying that module to make use of WebClient rather than using RestTemplate).

Before:

```
{"traceId":"27c05baf907ddbf9","parentId":"27c05baf907ddbf9","id":"f2d8b73a0b3ebcbb","kind":"SERVER","name":"GET /hi2","timestamp":1620249762102795,"duration":320323,"localEndpoint":{"serviceName":"testsleuthzipkin","ipv4":"10.1.1.84"},"remoteEndpoint":{"ipv4":"127.0.0.1","port":55319},"tags":{"http.method":"GET","http.path":"/hi2","mvc.controller.class":"SampleController","mvc.controller.method":"hi2","random-sleep-millis":"306"},"shared":true}
{"traceId":"27c05baf907ddbf9","parentId":"27c05baf907ddbf9","id":"f2d8b73a0b3ebcbb","kind":"CLIENT","name":"GET","timestamp":1620249761976433,"duration":462708,"localEndpoint":{"serviceName":"testsleuthzipkin","ipv4":"10.1.1.84"},"tags":{"http.method":"GET","http.path":"/hi2"}}
{"traceId":"27c05baf907ddbf9","id":"27c05baf907ddbf9","kind":"SERVER","name":"GET /","timestamp":1620249761637034,"duration":806335,"localEndpoint":{"serviceName":"testsleuthzipkin","ipv4":"10.1.1.84"},"remoteEndpoint":{"ipv6":"0:0:0:0:0:0:0:1","port":55318},"tags":{"http.method":"GET","http.path":"/","mvc.controller.class":"SampleController","mvc.controller.method":"hi"}}
```

After

```
{"traceId":"5d1504209f30a7f8","parentId":"5d1504209f30a7f8","id":"d0a4a9431789b6f4","kind":"SERVER","name":"GET /hi2","timestamp":1620250125350426,"duration":528620,"localEndpoint":{"serviceName":"testsleuthzipkin","ipv4":"10.1.1.84"},"remoteEndpoint":{"ipv4":"127.0.0.1","port":55357},"tags":{"http.method":"GET","http.path":"/hi2","mvc.controller.class":"SampleController","mvc.controller.method":"hi2","random-sleep-millis":"512"},"shared":true}
{"traceId":"5d1504209f30a7f8","parentId":"5d1504209f30a7f8","id":"d0a4a9431789b6f4","kind":"CLIENT","name":"GET /hi2","timestamp":1620250125231077,"duration":667700,"localEndpoint":{"serviceName":"testsleuthzipkin","ipv4":"10.1.1.84"},"tags":{"http.method":"GET","http.path":"/hi2"}}
{"traceId":"5d1504209f30a7f8","id":"5d1504209f30a7f8","kind":"SERVER","name":"GET /","timestamp":1620250124690578,"duration":1213581,"localEndpoint":{"serviceName":"testsleuthzipkin","ipv4":"10.1.1.84"},"remoteEndpoint":{"ipv6":"0:0:0:0:0:0:0:1","port":55356},"tags":{"http.method":"GET","http.path":"/","mvc.controller.class":"SampleController","mvc.controller.method":"hi"}}
```

jhinch:
It seems that I hadn't correctly captured some reformatting in my commit. I have fixed that issue now.

marcingrzejszczak:
Regardless, in the ui don't you just see one, merged span with one name? You don't have two separate logical spans for the rpc, right? Btw tog can check out spring-cloud-samples/spring-cloud-sleuth-samples where we have isolated examples specifically designed for such examples as this one

jhinch:
> Regardless, in the ui don't you just see one, merged span with one name? You don't have two separate logical spans for the rpc, right? Btw tog can check out spring-cloud-samples/spring-cloud-sleuth-samples where we have isolated examples specifically designed for such examples as this one

Yeah, I'm not suggesting that there is an issue with Zipkin merging server and client spans together. However, there are a number of situations where the client span name is still important

* Some tracing UIs don't merge spans. For example LightStep doesn't appear to do so
* You might not control the upstream service and it may not generate tracing spans, so you only have the client span available to you.

I can take another look at producing an example of a trace which doesn't have a corresponding server side span to demonstrate how it would look in the Zipkin UI.

marcingrzejszczak:
No, no it's fine. Your arguments are valid IMO. I'll try to look into this shortly

marcingrzejszczak:
I wonder if we shouldn't make this PR against 3.1.x branch. WDYT @jonatan-ivanov  ?

jhinch:
Alrightly. here is an example of a modified version of  `spring-cloud-samples/spring-cloud-sleuth-samples`. Instead of calling back to the same service, I ran two instances and one communicated to the other with the upstream service configured to not emit server spans.

Before

zipkin-before

After

zipkin-after

jhinch:
From my perspective I think I agree that this should probably go into a minor version bump. I am more than happy to rebased my changes to target 3.1.x if needed.

marcingrzejszczak:
Yeah could you rebase it against 3.1.x?

jonatan-ivanov:
@marcingrzejszczak I agree with putting this into 3.1.x.

Btw another use-case and a possible way to simulate it is when the endpoint(/network) that the client called is down so there won't be spans reported by the server since the server never received the request.

marcingrzejszczak:
Ok I'm closing this pr and please file a new one against branch 3.1.x

PR #1936, opened 2021-05-06
Kafka instrumentation enhancements

flavium31:
@marcingrzejszczak Raising this PR as agreed on #1920. This PR includes:

- Documentation for Kafka instrumentation
- Integration tests (both common and brave) for all instrumented Kafka clients (KafkaConsumer, KafkaProducer, KakaSender, KakfaReceiver)
- Redesigned instrumentation for the reactive KafkaReceiver for consistent approach with KafkaSender
- Auto-configuration module for Kafka instrumentation
- Javadocs updated (since tag bumped to 3.1.0)


spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/instrument/kafka/TracingKafkaConsumerBeanPostProcessor.java
```diff
@@ -0,0 +1,53 @@
...
+/**
+ * Bean post processor for {@link org.apache.kafka.clients.consumer.Consumer}.
+ *
+ * @author Anders Clausen
+ * @author Flaviu Muresan
+ * @since 3.1.0
+ */
+public class TracingKafkaConsumerBeanPostProcessor implements BeanPostProcessor {
+
+	private final Propagator propagator;
```
marcingrzejszczak, line 36: BeanPostProcessors should not inject beans. They should inject bean factories and then the beans should be resolved at runtime. Otherwise we can initialize beans too early.

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/instrument/kafka/TracingKafkaProducerBeanPostProcessor.java
```diff
@@ -0,0 +1,58 @@
...
+/**
+ * Bean post processor for {@link org.apache.kafka.clients.producer.Producer}.
+ *
+ * @author Anders Clausen
+ * @author Flaviu Muresan
+ * @since 3.1.0
+ */
+public class TracingKafkaProducerBeanPostProcessor implements BeanPostProcessor {
+
+	private final Tracer tracer;
```
marcingrzejszczak, line 37: Same case as with the previous bpp

spring-cloud-sleuth-autoconfigure/pom.xml
```diff
@@ -1,470 +1,475 @@
-<?xml version="1.0" encoding="UTF-8"?>
-<!--
```
spencergibb, line 2: Looks like a whitespace change? If so, please revert

spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/deployer/TraceAppDeployer.java
```diff
@@ -1,92 +1,92 @@
...
-			span.tag("deployer.app.id", id);
-			registerListener(span, id);
-			return id;
-		}
-	}
-
-	private void registerListener(Span span, String id) {
-		PreviousAndCurrentStatus previousAndCurrentStatus = new PreviousAndCurrentStatus(span);
+/*
+ * Copyright 2018-2021 the original author or authors.
```
spencergibb, line 2: More whitespace?

spencergibb (changes requested):
Looks like some whitespace changes that make large diffs. Please revert

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/instrument/kafka/TracingKafkaConsumerBeanPostProcessor.java
flavium31: @marcingrzejszczak I've changed the interfaces of the decorating classes (and the auto-configuration part as well) to accept a BeanFactory instead and lazily resolve the beans at runtime

spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/kafka/TracingKafkaPropagatorGetter.java
```diff
@@ -30,7 +30,7 @@
  *
  * @author Anders Clausen
  * @author Flaviu Muresan
- * @since 3.0.3
+ * @since 3.1.0
```
spencergibb, line 33: Why would this change?

spencergibb (changes requested):

spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/kafka/TracingKafkaPropagatorGetter.java
flavium31: Another PR for this functionality was initially raised against master (#1906), and @marcingrzejszczak suggested that we raise it against 3.1.x instead (which I did in #1920 ). However when #1920 was merged, we missed bumping up the `since` version, therefore doing it in this PR (see https://github.com/spring-cloud/spring-cloud-sleuth/pull/1920#issuecomment-824897747)

flavium31:
@marcingrzejszczak I've pushed a fix for the issue you mentioned above - both the kafka and the messaging autoconfig modules were registering `Propagator.Setter<>` and `Propagator.Getter<>` beans (with different type parameters), but the type parameter was not considered by the `@ConditionalOnMissingBean`. However the build is failing now with a different issue (looks like something in the task instrumentation module) and I can't quite see how this one is related to the code in this PR. Any  tips?

marcingrzejszczak:
The task test was my fault cause I've added transaction support and apparently it did sth to the task test. Anyways, I've merged your stuff! Thanks a lot!

PR #1959, opened 2021-05-25
Adding jdbc instrumentation 

marcingrzejszczak:


spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/instrument/jdbc/P6SpyPropertiesSetter.java
```diff
@@ -0,0 +1,188 @@
...
+		return Stream.of(spyDotProperties, new EnvironmentVariables(), new SystemProperties()).filter(Objects::nonNull)
+				.map(P6OptionsSource::getOptions).filter(Objects::nonNull)
+				.flatMap(options -> options.entrySet().stream())
+				.collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue,
+						// always using value from the first P6OptionsSource
+						(value1, value2) -> value1));
+	}
+
+	@Override
+	public void postProcessBeanDefinitionRegistry(BeanDefinitionRegistry registry) throws BeansException {
```
jonatan-ivanov, line 184: Does this need to be overridden?

spring-cloud-sleuth-api/src/main/java/org/springframework/cloud/sleuth/Span.java
```diff
@@ -99,6 +99,17 @@ default Span remoteServiceName(String remoteServiceName) {
...
 	}
 
+	/**
+	 * Sets the remote url on the span.
+	 * @param ip remote ip
+	 * @param port remote port
+	 * @return this span
+	 * @since 3.1.0
+	 */
+	default Span remoteUrl(String ip, int port) {
```
jonatan-ivanov, line 109: Should we call it `remoteUrl`?
Brave calls it `remoteIpAndPort`.

spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/jdbc/DataSourceNameResolver.java
```diff
@@ -0,0 +1,41 @@
...
+ * {@link CommonDataSource} name resolver based on bean name.
+ *
+ * @author Arthur Gavlyukovskiy
+ * @since 3.1.0
+ */
+public class DataSourceNameResolver {
+
+	private final ConcurrentHashMap<CommonDataSource, String> cachedNames = new ConcurrentHashMap<>();
+
+	public void addDataSource(String name, CommonDataSource dataSource) {
```
jonatan-ivanov, line 33: Can we keep this package-private?

spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/jdbc/SleuthJdbcSpan.java
```diff
@@ -0,0 +1,167 @@
...
+		}
+
+		@Override
+		public EventValue[] getEvents() {
+			return QueryEvents.values();
+		}
+
+		@Override
+		public String prefix() {
+			return "sql.";
```
jonatan-ivanov, line 71: Should this be jdbc (so that we can have r2dbc too)?

jonatan-ivanov (approved):

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/instrument/jdbc/P6SpyPropertiesSetter.java
marcingrzejszczak: Unfortunately it has to

spring-cloud-sleuth-api/src/main/java/org/springframework/cloud/sleuth/Span.java
marcingrzejszczak: I know. I thought that `remoteUrl` sounds better. I don't have a problem to name it in the same way as Brave does it

spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/jdbc/DataSourceNameResolver.java
marcingrzejszczak: We can't cause it's used in the `TraceDataSourceDecoratorBeanPostProcessor`

spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/jdbc/DataSourceWrapper.java
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Interface that implicitly added to the CGLIB proxy of {@link DataSource}.
+ *
+ * Returns link of both real {@link DataSource}, decorated {@link DataSource} and all
+ * decorating chain including decorator bean name, instance and result of decorating.
+ *
+ * @author Arthur Gavlyukovskiy
+ * @since 3.1.0
+ */
```
gavlyukovskiy, line 31: This doc doesn't really make sense anymore - it's no an interface and there is no chain

docs/src/main/asciidoc/_configprops.adoc
```diff
@@ -23,6 +23,27 @@
...
+|spring.sleuth.jdbc.decorator.datasource.includes |  | Which types of tracing we would like to include.
+|spring.sleuth.jdbc.decorator.datasource.p6spy.custom-appender-class |  | Class file to use (only with logging=custom). The class must implement {@link com.p6spy.engine.spy.appender.FormattedLogger}.
+|spring.sleuth.jdbc.decorator.datasource.p6spy.enable-logging | `true` | Enables logging JDBC events.
+|spring.sleuth.jdbc.decorator.datasource.p6spy.log-file | `spy.log` | Name of log file to use (only with logging=file).
+|spring.sleuth.jdbc.decorator.datasource.p6spy.log-filter.pattern |  | Use regex pattern to filter log messages. Only matched messages will be logged.
+|spring.sleuth.jdbc.decorator.datasource.p6spy.log-format |  | Custom log format.
+|spring.sleuth.jdbc.decorator.datasource.p6spy.logging |  | Logging to use for logging queries.
+|spring.sleuth.jdbc.decorator.datasource.p6spy.multiline | `true` | Enables multiline output.
+|spring.sleuth.jdbc.decorator.datasource.p6spy.tracing.include-parameter-values | `true` | Report the effective sql string (with '?' replaced with real values) to tracing systems. <p> NOTE this setting does not affect the logging message.
```
gavlyukovskiy, line 46: It doesn't make much sense to have all these properties in Sleuth, maybe some defaults need to be changed to `false`, like `spring.sleuth.jdbc.decorator.datasource.p6spy.enable-logging`/`spring.sleuth.jdbc.decorator.datasource.datasource-proxy.query.enable-logging`.  These defaults made sense when it was separate library with Sleuth as only one of the functions, but when someone enables tracing, and if I remember correctly traces are created with some probability, then starting to see logs for each and every query/commit/rollback will be a bit strange for Sleuth users.

Also maybe `spring.sleuth.jdbc.decorator.datasource.*` can be replaced with just `spring.sleuth.jdbc.*`, WDYT?

docs/src/main/asciidoc/_configprops.adoc
```diff
@@ -23,6 +23,27 @@
 |spring.sleuth.integration.enabled | `true` | Enable Spring Integration instrumentation.
 |spring.sleuth.integration.patterns | `[!hystrixStreamOutput*, *, !channel*]` | An array of patterns against which channel names will be matched. @see org.springframework.integration.config.GlobalChannelInterceptor#patterns() Defaults to any channel name not matching the Hystrix Stream and functional Stream channel names.
 |spring.sleuth.integration.websockets.enabled | `true` | Enable tracing for WebSockets.
+|spring.sleuth.jdbc.decorator.datasource.datasource-proxy.count-query | `false` | Creates listener to count queries. @see ProxyDataSourceBuilder#countQuery() @see QueryCountHolder
```
gavlyukovskiy, line 26: This one probably is not needed, I have used it to generate metrics at some point, but it's all gone now

docs/src/main/asciidoc/_configprops.adoc
marcingrzejszczak: > I remember correctly traces are created with some probability, then starting to see logs for each and every query/commit/rollback will be a bit strange for Sleuth users.

Traces are always created. With certain probability they are sent to a reporting system.

> Also maybe spring.sleuth.jdbc.decorator.datasource.* can be replaced with just spring.sleuth.jdbc.*, WDYT?

great idea

marcingrzejszczak: > maybe some defaults need to be changed to false, like spring.sleuth.jdbc.decorator.datasource.p6spy.enable-logging/spring.sleuth.jdbc.decorator.datasource.datasource-proxy.query.enable-logging.

Actually having those specific logging is pretty cool. Do you think that it should be an opt-in feature?

gavlyukovskiy: it depends :) On my current project if query logging was suddenly enabled our logging cluster would just die. It would easily add 60% of logging traffic, unlike tracing which is throttled to some extent.

marcingrzejszczak: Makes sense. I'll make it opt in.

PR #1961, opened 2021-05-31
Adds Spring Session support

marcingrzejszczak:
cc @rwinch @jzheaux


jzheaux (changes requested):
woot! Nice to see this happening.

I've shared some thoughts inline about a couple of the proposed tags.

PR #1963, opened 2021-06-01
WIP: Adds Spring Security support

marcingrzejszczak:
cc @rwinch @jgrandja @jzheaux @eleftherias


spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/instrument/web/LazyTracingFilter.java
```diff
@@ -0,0 +1,67 @@
...
+		tracingFilter().doFilter(request, response, chain);
+	}
+
+	@Override
+	public void destroy() {
+		tracingFilter().destroy();
+	}
+
+	private Filter tracingFilter() {
+		if (this.tracingFilter == null) {
```
jonatan-ivanov, line 60: This can be called concurrently in the beginning, should not we initialize this keeping that in mind, e.g.: double-checked locking?

spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/web/SleuthWebSpan.java
```diff
@@ -78,4 +78,74 @@ public String getKey() {
...
+	 */
+	enum SecurityTags implements TagKey {
+
+		/**
+		 * Authorities assigned to the user.
+		 */
+		AUTHORITIES {
+			@Override
+			public String getKey() {
+				return "security.authentication.authorities";
```
jonatan-ivanov, line 95: Would not `security.authorities` be better?
Does it make sense to talk about authorities in terms of authentication should not it belong to authorization?

spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/web/SleuthWebSpan.java
```diff
@@ -78,4 +78,74 @@ public String getKey() {
...
+			}
+		},
+
+		/**
+		 * Whether user is authenticated.
+		 */
+		AUTHENTICATED {
+			@Override
+			public String getKey() {
+				return "security.authentication.authenticated";
```
jonatan-ivanov, line 105: WDYT about `security.authenticated`?

spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/web/SleuthWebSpan.java
```diff
@@ -78,4 +78,74 @@ public String getKey() {
 
 	}
 
+	/**
+	 * Tags related to security.
+	 *
+	 * @author Marcin Grzejszczak
+	 * @since 3.1.0
+	 */
+	enum SecurityTags implements TagKey {
```
jonatan-ivanov, line 87: Is there a way that we can add a flag that shows if the principal (user) is `anonymous` (authenticated)?
Or will the list of authorities contain that information?

spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/web/SleuthWebSpan.java
jonatan-ivanov, line 95: I'm thinking if we should make this configurable (enable/disable) in case anyone is concerned that breaking into the tracing backend could give extra information to attackers.

spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/web/TracingSecurityServletFilter.java
```diff
@@ -0,0 +1,91 @@
...
+			private TracingSecurityServletFilter tracingSecurityServletFilter;
+
+			@Override
+			public void doFilter(ServletRequest servletRequest, ServletResponse servletResponse,
+					FilterChain filterChain) throws IOException, ServletException {
+				tracingSecurityFilter().doFilter(servletRequest, servletResponse, filterChain);
+			}
+
+			private TracingSecurityServletFilter tracingSecurityFilter() {
+				if (this.tracingSecurityServletFilter == null) {
```
jonatan-ivanov, line 82: Double-checked locking so that it can be safe if it is called concurrently?

spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/web/TracingSecurityTagSetter.java
```diff
@@ -0,0 +1,57 @@
...
+	static void setSecurityTags(Span span, Authentication authentication) {
+		if (log.isDebugEnabled()) {
+			log.debug("Will set security tags on span [" + span + "]");
+		}
+		AssertingSpan assertingSpan = AssertingSpan.of(SleuthWebSpan.WEB_FILTER_SPAN, span);
+		assertingSpan.tag(SecurityTags.AUTHORITIES,
+				StringUtils.collectionToCommaDelimitedString(authentication.getAuthorities()));
+		assertingSpan.tag(SecurityTags.AUTHENTICATED, String.valueOf(authentication.isAuthenticated()));
+		Object principal = authentication.getPrincipal();
+		if (principal instanceof User) {
```
jonatan-ivanov, line 46: Are there other scenarios we need to handle?

jonatan-ivanov:
I guess this can be closed because https://github.com/spring-cloud/spring-cloud-sleuth/pull/2011 and other pieces that possibly will be added in the future.

PR #1964, opened 2021-06-02
Actuator endpoint with spans

### LINKED ISSUE #1879: Custom actuator for storing trace ids
Create an actuator endpoint that would store trace ids.

cc @shakuzen @jonatan-ivanov @martinlippert

marcingrzejszczak:
fixes #1879


spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/actuate/SleuthActuatorProperties.java
```diff
@@ -0,0 +1,55 @@
...
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * Settings for Sleuth actuator.
+ *
+ * @author Marcin Grzejszczak
+ * @since 3.1.0
+ */
+@ConfigurationProperties(prefix = "spring.sleuth.actuator")
```
snicoll, line 27: That namespace looks a bit odd to me. Your endpoint is going in the `management` namespace and it already has a way to be enabled/disabled via Spring Boot...

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/actuate/TextOutputFormat.java
```diff
@@ -0,0 +1,50 @@
...
+
+import com.google.protobuf.TextFormat;
+
+import org.springframework.boot.actuate.endpoint.Producible;
+import org.springframework.http.MediaType;
+import org.springframework.util.MimeType;
+import org.springframework.util.MimeTypeUtils;
+
+/**
+ * A {@link Producible} enum for supported Prometheus {@link TextFormat}.
```
snicoll, line 27: Copy/paste?

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/actuate/TraceSleuthActuatorAutoConfiguration.java
```diff
@@ -0,0 +1,76 @@
...
+
+/**
+ * {@link org.springframework.boot.autoconfigure.EnableAutoConfiguration
+ * Auto-configuration} for Sleuth actuator endpoint.
+ *
+ * @author Marcin Grzejszczak
+ * @since 3.1.0
+ */
+@Configuration(proxyBeanMethods = false)
+@ConditionalOnProperty(value = {"spring.sleuth.enabled", "spring.sleuth.actuator.enabled"}, matchIfMissing = true)
```
snicoll, line 39: There is a `@ConditionalOnEnabledEndpoint` you should be using instead.

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/actuate/TracesScrapeEndpoint.java
```diff
@@ -0,0 +1,69 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.cloud.sleuth.autoconfig.actuate;
+
+import java.util.List;
+
+import org.jetbrains.annotations.NotNull;
```
snicoll, line 21: Wrong import?

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/actuate/TextOutputFormat.java
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @author Andy Wilkinson
+ * @since 2.5.0
+ */
+public enum TextOutputFormat implements Producible<TextOutputFormat> {
+
+	/**
+	 * OpenZipkin text.
+	 */
+	CONTENT_TYPE_OPENZIPKIN_JSON_V2(MediaType.APPLICATION_JSON_VALUE);
```
snicoll, line 37: There is only one format so why do you need this? To override the content type?

spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/internal/ContextUtil.java
```diff
@@ -36,7 +36,7 @@ private ContextUtil() {
 	private static final Log log = LogFactory.getLog(ContextUtil.class);
 
 	/**
-	 * @param beanFactory bean facotry
+	 * @param beanFactory bean factory
```
snicoll, line 39: Unrelated?

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/actuate/TracesScrapeEndpoint.java
```diff
@@ -0,0 +1,69 @@
...
+	public WebEndpointResponse<String> spansSnapshot(TextOutputFormat format) {
+		List<FinishedSpan> finishedSpans = this.bufferingSpanReporter.getFinishedSpans();
+		return response(format, finishedSpans);
+	}
+
+	@NotNull
+	private WebEndpointResponse<String> response(TextOutputFormat format, List<FinishedSpan> finishedSpans) {
+		String spans = this.finishedSpanWriter.write(format, finishedSpans);
+		if (spans == null) {
+			return new WebEndpointResponse<>("The format [" + format.getProducedMimeType().toString()  + " ] is not supported", HttpStatus.UNSUPPORTED_MEDIA_TYPE.value());
```
jonatan-ivanov, line 59: `.toString()` is not needed in `format.getProducedMimeType().toString()`

jonatan-ivanov, line 59: Isn't it `NOT_ACCEPTABLE (406)` instead of `UNSUPPORTED_MEDIA_TYPE (415)`?

To me, based on these:
https://datatracker.ietf.org/doc/html/rfc7231#section-6.5.6
https://datatracker.ietf.org/doc/html/rfc7231#section-6.5.13

`NOT_ACCEPTABLE` means that the server is not able to produce a response in the format that is acceptable to the client (`Accept` header) while `UNSUPPORTED_MEDIA_TYPE` means that the request  is in a format that is not supported by the server (`Content-Type` and its friends).

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/actuate/TracesScrapeEndpoint.java
```diff
@@ -0,0 +1,69 @@
...
+		}
+		return new WebEndpointResponse<>(spans, format);
+	}
+
+	@WriteOperation(producesFrom = TextOutputFormat.class)
```
jonatan-ivanov, line 64: Should not this be a `@DeleteOperation`? To me it is kind of what it does, it deletes all the traces.
From HTTP perspective, I guess `@WriteOperation` will be a `POST` method (creates new resources) while `@DeleteOperation` will be a `DELETE` method (deletes resources).

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/actuate/TracesScrapeEndpoint.java
```diff
@@ -0,0 +1,69 @@
...
+	private final BufferingSpanReporter bufferingSpanReporter;
+
+	private final FinishedSpanWriter finishedSpanWriter;
+
+	public TracesScrapeEndpoint(BufferingSpanReporter bufferingSpanReporter, FinishedSpanWriter finishedSpanWriter) {
+		this.bufferingSpanReporter = bufferingSpanReporter;
+		this.finishedSpanWriter = finishedSpanWriter;
+	}
+
+	@ReadOperation(producesFrom = TextOutputFormat.class)
```
jonatan-ivanov, line 49: I'm not 100% sure I get what will this do with the enum. 
Will this method be called if I pass a request with a media type that is not in the TextOutputFormat enum,  e.g.: `text/html`?

spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/bridge/CompositeSpanHandler.java
```diff
@@ -35,8 +36,11 @@
 
 	private final List<SpanFilter> exporters;
```
jonatan-ivanov, line 37: I know this is not really in scope for this PR but it is a little related and seeing `exporters` and `reporters` puzzles me a little? Can we rename `exporters` to `filters`?

jonatan-ivanov:
I have two, kind of fundamental questions for this change:
- This means we won't do SEE or WebSockets, right? 
- What would this look like from the clients' perspective (i.e.: should this work for multiple clients)?

Multiple clients can call `GET` while only one of them will call `DELETE` that clears the buffer every once in a while?
In this case the call that clears the buffer can lead to data loss for other clients and data loss for everyone if the client that suppose to clear the buffer is down.
Or is this intended by using on only a single client that polls on `DELETE` while `GET` is available for debugging purposes?

Would not it be better streaming the data as it is reported to the clients?

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/actuate/TraceSleuthActuatorAutoConfiguration.java
```diff
@@ -0,0 +1,76 @@
...
+	static class BraveActuatorConfiguration {
+
+		@Bean
+		@ConditionalOnMissingBean
+		FinishedSpanWriter braveFinishedSpanWriter() {
+			return new BraveFinishedSpanWriter();
+		}
+
+		@Bean
+		SpanHandler braveCustomSpanHandler() {
```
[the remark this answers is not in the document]
marcingrzejszczak: Otherwise all spans are noop.

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/actuate/BufferingSpanReporter.java
```diff
@@ -0,0 +1,83 @@
...
+		while (iterator.hasNext()) {
+			events.add(iterator.next());
+			iterator.remove();
+		}
+		this.estimatedSize.set(0);
+		return events;
+	}
+
+	@Override
+	public void report(FinishedSpan span) {
```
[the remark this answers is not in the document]
marcingrzejszczak: I was thinking about doing it like this but I followed what Boot does.

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/actuate/TextOutputFormat.java
marcingrzejszczak: Me? Never ;)

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/actuate/TextOutputFormat.java
marcingrzejszczak: I'll add more.

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/actuate/TracesScrapeEndpoint.java
marcingrzejszczak: We follow what Boot does.

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/actuate/BufferingSpanReporter.java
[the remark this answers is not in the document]
jonatan-ivanov: I think it might make more sense in case of Boot if you are referring to the `/actuator/startup` endpoint and `BufferingApplicationStartup`. In case of Sleuth I think old spans are not as important as new ones.

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/actuate/TracesScrapeEndpoint.java
marcingrzejszczak: Apparently in that case the first value of the enum is picked :woman_shrugging:

marcingrzejszczak:
> This means we won't do SEE or WebSockets, right?

Not until Actuator supports it.

>What would this look like from the clients' perspective (i.e.: should this work for multiple clients)?

I don't think it should.

> Would not it be better streaming the data as it is reported to the clients?

Most likely it would but it's not yet supported by actuator.

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/actuate/BufferingSpanReporter.java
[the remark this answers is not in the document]
marcingrzejszczak: I've updated the code with a pseudo ring buffer

PR #1976, opened 2021-06-17
Fix not able to override datasource-proxy log-level

### LINKED ISSUE #1973: datasource-proxy log level is always DEBUG, despite setting property to INFO
I'm using 2021.0.0-SNAPSHOT to get datasource-proxy functionality.

I'm setting the property spring.sleuth.jdbc.datasource-proxy.query.log-level to INFO and I get nothing in my log.

Added a breakpoint here DataSourceProxyBuilderCustomizer#customize(ProxyDataSourceBuilder proxyDataSourceBuilder

and it seems that the result of this.datasourceProxy.getQuery().getLogLevel() is always DEBUG , regardless of what you set it the the property to.

Tried the same datasource-proxy scenario via datasource-proxy-spring-boot-starter 1.7.1 and it works fine.

Please let me know if I can provide any further information

beatfreaker:
Fixes #1973 

@marcingrzejszczak @gavlyukovskiy can you please review this PR

- Aligned default values of `Query` & `SlowQuery` in `TraceJdbcProperties.DataSourceProxyProperties` and `DataSourceProxyProperties`
- Add code to copy properties of `Query` and `SlowQuery` in `DataSourceProxyConfiguration`
- Add test case to check if `log-level` can be overridden or not

### DIFF
--- a/spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/instrument/jdbc/DataSourceProxyConfiguration.java
+++ b/spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/instrument/jdbc/DataSourceProxyConfiguration.java
@@ -84,6 +84,8 @@ class DataSourceProxyConfiguration {
 		DataSourceProxyProperties props = new DataSourceProxyProperties();
 		BeanUtils.copyProperties(originalProxy, props);
 		props.setLogging(DataSourceProxyProperties.DataSourceProxyLogging.valueOf(originalProxy.getLogging().name()));
+		BeanUtils.copyProperties(originalProxy.getQuery(), props.getQuery());
+		BeanUtils.copyProperties(originalProxy.getSlowQuery(), props.getSlowQuery());
 		return props;
 	}
 
--- a/spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/instrument/jdbc/DataSourceProxyConfiguration.java
+++ b/spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/instrument/jdbc/DataSourceProxyConfiguration.java
@@ -50,6 +50,7 @@ import org.springframework.context.annotation.Bean;
  * {@link QueryTransformer}.
  *
  * @author Arthur Gavlyukovskiy
+ * @author Chintan Radia
  */
 @ConditionalOnClass(ProxyDataSource.class)
 @ConditionalOnProperty(name = "spring.sleuth.jdbc.datasource-proxy.enabled", havingValue = "true",
--- a/spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/instrument/jdbc/TraceJdbcProperties.java
+++ b/spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/instrument/jdbc/TraceJdbcProperties.java
@@ -190,7 +190,7 @@ public class TraceJdbcProperties {
 			/**
 			 * Enable logging all queries to the log.
 			 */
-			private boolean enableLogging = true;
+			private boolean enableLogging = false;
 
 			/**
 			 * Name of query logger.
@@ -236,7 +236,7 @@ public class TraceJdbcProperties {
 			/**
 			 * Enable logging slow queries to the log.
 			 */
-			private boolean enableLogging = true;
+			private boolean enableLogging = false;
 
 			/**
 			 * Name of slow query logger.
@@ -253,7 +253,7 @@ public class TraceJdbcProperties {
 			 */
 			private long threshold = 300;
 
-			boolean isEnableLogging() {
+			public boolean isEnableLogging() {
 				return enableLogging;
 			}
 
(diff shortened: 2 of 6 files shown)


spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/instrument/jdbc/TraceJdbcProperties.java
```diff
@@ -190,7 +190,7 @@ public void setJsonFormat(boolean jsonFormat) {
 			/**
 			 * Enable logging all queries to the log.
 			 */
-			private boolean enableLogging = false;
+			private boolean enableLogging = true;
```
marcingrzejszczak, line 193: We had a chat with @gavlyukovskiy that having this turned on by default can lead to gigantic amount of logs. I'd leave both as opt in.

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/instrument/jdbc/DataSourceProxyConfiguration.java
```diff
@@ -84,6 +84,8 @@ private DataSourceProxyProperties props(TraceJdbcProperties traceJdbcProperties)
 		DataSourceProxyProperties props = new DataSourceProxyProperties();
 		BeanUtils.copyProperties(originalProxy, props);
 		props.setLogging(DataSourceProxyProperties.DataSourceProxyLogging.valueOf(originalProxy.getLogging().name()));
+		BeanUtils.copyProperties(originalProxy.getQuery(), props.getQuery());
```
marcingrzejszczak, line 87: :grimacing: oops

marcingrzejszczak (changes requested):
Thanks for catching my mistakes :facepalm:

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/instrument/jdbc/TraceJdbcProperties.java
beatfreaker: So in that case we should set this [DataSourceProxyProperties.java#L116](https://github.com/spring-cloud/spring-cloud-sleuth/blob/3.1.x/spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/jdbc/DataSourceProxyProperties.java#L116) to `false` too, because at the end we are using this property class to build `ProxyDataSource`.

Also, if we set it to `false`, we have to modify jUnit as well, because then by default `SLF4JQueryLoggingListener.class` won't be registered.

So by default we have to enable `SLF4JSlowQueryListener` and `SLF4JQueryLoggingListener` should be opt-in feature.

Is this understanding correct ?

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/instrument/jdbc/DataSourceProxyConfiguration.java
beatfreaker: Is this correct approach ? or we can have more optimal way to do it ?

marcingrzejszczak: I said oops cause I've forgotten to add those lines :facepalm: Great that you caught it!

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/instrument/jdbc/TraceJdbcProperties.java
marcingrzejszczak: I think that we want to set all the default additional loggings to `false` and make both opt in.

beatfreaker: Got it, will make all the `enableLogging=false` means no `QueryLogListener` or `SlowQueryLogListener` will be registered by default.

marcingrzejszczak:
Lgtm - can we also mention that in the docs? Wdyt @gavlyukovskiy @jonatan-ivanov - are we not getting too defensive here with the opt in feature?

jonatan-ivanov:
@marcingrzejszczak Although, I would love to see these details, I think the right call here is disable it by default:
1. It could have a performance impact (though I only measured this when Java 6-7 was the latest version)
2. It can have security implications (logging out PII data, sensitive data, secrets)

I think we could put an `INFO` message into the logs that lets the user know that this feature is turned off (maybe in the `afterPropertiesSet`/`@PostConstruct` of the ConfigProps bean?).

jonatan-ivanov (approved):

marcingrzejszczak:
I don't think we should set ` message into the logs that lets the users know that this feature is turned off (maybe in the afterPropertiesSet/@PostConstruct of the ConfigProps bean?).` For example we don't do it with setting `RestTemplate` as a bean - we describe it in the docs. Maybe writing about this in the documentation is good enough?

marcingrzejszczak (changes requested):
I'd update the documentation too with the property how to turn on the logging. Other than that LGTM

marcingrzejszczak (approved):

marcingrzejszczak:
Great work @beatfreaker ! Thank you so much for your contribution!

PR #1977, opened 2021-06-17
Adds support for Kotlin Coroutines

### LINKED ISSUE #1737: Sleuth does not work in Coroutines context
**Describe the bug**
traceId, spanId and exportable are not printed in Coroutines context.

Versions
Spring Boot: 2.3.3.RELEASE
Spring Cloud: Hoxton.SR8

**Sample**
You can see full code by [sleuth-coroutines](https://github.com/ninja-inc/sleuth-coroutines).

```
	@GetMapping("/hello")
	suspend fun hello(): Map<String, String> {
		log.info("in main context")

		GlobalScope.launch {
			log.info("in Coroutines context (launch)")
		}

		GlobalScope.async {
			log.info("in Coroutines context (async)")
		}.await()

		return mapOf(
			"msg" to "hello"
		)
	}
```

Printed log messages
```
2020-09-12 21:03:15.306  INFO [sleuth-coroutines,42883585174b8ae0,42883585174b8ae0,true] 22621 --- [ctor-http-nio-2] c.e.sleuthcoroutines.WebController       : in main context
2020-09-12 21:03:15.312  INFO [sleuth-coroutines,,,] 22621 --- [atcher-worker-2] c.e.sleuthcoroutines.WebController       : in Coroutines context (async)
2020-09-12 21:03:15.312  INFO [sleuth-coroutines,,,] 22621 --- [atcher-worker-1] c.e.sleuthcoroutines.WebController       : in Coroutines context (launch)
```

marcingrzejszczak:
fixes #1737 

cc @sdeleuze


.editorconfig
```diff
@@ -1,28 +1,33 @@
-root = true
```
jonatan-ivanov, line 1: Why is this happening (I mean the diff shows that every line has changed). Is it because of the line endings?

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/actuate/FinishedSpanWriter.java
```diff
@@ -1,40 +1,40 @@
-/*
```
jonatan-ivanov, line 1: This file did not change, right? (line endings?)

jonatan-ivanov (changes requested):
Could you please check what happened with the 3 modified .java and the .editorconfig files?

PR #1982, opened 2021-06-23
Adds support for customize HandlerParser postHandle

wuwen5:
Spring MVC customize traced requests handler add postHandle method.

### DIFF
--- a/spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/web/mvc/HandlerParser.java
+++ b/spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/web/mvc/HandlerParser.java
@@ -23,6 +23,7 @@ import org.springframework.cloud.sleuth.SpanCustomizer;
 import org.springframework.cloud.sleuth.docs.AssertingSpanCustomizer;
 import org.springframework.web.method.HandlerMethod;
 import org.springframework.web.servlet.HandlerInterceptor;
+import org.springframework.web.servlet.ModelAndView;
 
 /**
  * Spring MVC specific type used to customize traced requests based on the handler.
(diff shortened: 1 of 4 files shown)


marcingrzejszczak:
Hi! Thanks for the pr. Can you elaborate on why we need this change and what problem are you trying to solve?

wuwen5:
It is convenient to customize tag information, such as httpstatus, payload, etc

In the case of use ```org.springframework.web.filter.AbstractRequestLoggingFilter```, the ```payload``` value can only be obtained through ```posthandler```, but not by ```prehandler```

```java

@Bean
HandlerParser handlerParser() {
    return new HandlerParser() {

        protected void postHandle(HttpServletRequest request, Object handler, ModelAndView modelAndView,
                                  SpanCustomizer customizer) {

            customizer.tag("http.status.code", modelAndView.getStatus().value());

            customizer.tag("http.payload", getMessagePayload(request));
            //etc...
        }

        protected String getMessagePayload(HttpServletRequest request) {
            ContentCachingRequestWrapper wrapper =
                    WebUtils.getNativeRequest(request, ContentCachingRequestWrapper.class);
            if (wrapper != null) {
                byte[] buf = wrapper.getContentAsByteArray();
                if (buf.length > 0) {
                    int length = Math.min(buf.length, getMaxPayloadLength());
                    try {
                        return new String(buf, 0, length, wrapper.getCharacterEncoding());
                    }
                    catch (UnsupportedEncodingException ex) {
                        return "[unknown]";
                    }
                }
            }
            return null;
        }
    };
}
```

marcingrzejszczak:
I see. Can you write a test  for this feature? Also since it's a new feature can you apply the pr to the 3.1.x branch?

marcingrzejszczak:
Thanks for changing the branch. Can you write a test for this please?

PR #1986, opened 2021-06-27
Set spanId in TracingResponderRSocketProxy

making:
also respect traceIdHigh and debug flag in TracingMetadata


marcingrzejszczak:
Thanks, great work as usual @making !

Issue #1990, opened 2021-07-06
Application failed to start with VaultPropertySource in webflux


chany714:
**Describe the bug**
* Application failed to start with VaultPropertySource + spring cloud sleuth in webflux.
* Failed to instantiate zipkin kafka sender
* (Not sure should I report this issue to spring cloud vault project.)

dependencies
- spring-cloud-starter-stream-kafka:3.1.3
- spring-cloud-sleuth-zipkin:3.0.3
- spring-cloud-starter-gateway:3.0.3
- spring-cloud-starter-sleuth:3.0.3
- spring-cloud-starter-vault-config:3.0.3

**Sample**
https://github.com/chany714/sleuth-bug-sample

```
Error starting ApplicationContext. To display the conditions report re-run your application with 'debug' enabled.
2021-07-06 10:38:28.822 ERROR [demo,,] 34990 --- [           main] o.s.boot.SpringApplication               : Application run failed

[...]

PR #74, opened 2015-12-22
Bumps to latest zipkin and hints about JAVA_OPTS in docker

adriancole:

PR #1994, opened 2021-07-13
Fix startup failure when using VaultPropertySource + reactive

beatfreaker:
Fixes gh-1990

When using `@ValuePropertySource` in reactive web app,
`KafkaSender` bean is getting created before `BeanPostProcessor` populate `@Value`
and as a result topic is passed as `null` in `KafkaSender.Builder` and it throws `NullPointerException`

To resolve this we are using `Environment` to get the property value instead on relying on `BeanPostProcessor`
to populate `@Value`

@jonatan-ivanov Added test to simulate failure scenario here [#74](https://github.com/spring-cloud-samples/sleuth-issues/pull/74)


marcingrzejszczak (approved):

marcingrzejszczak:
IMO this is an obvious fix. We're just resolving the property later on within the bean lifecycle. LGTM

jonatan-ivanov (approved):

jonatan-ivanov:
@beatfreaker Thank you for the fix! :)

Issue #1996, opened 2021-07-14
/traces actuator endpoint should return valid JSON response when Accept = */* is passed in request header


beatfreaker:
**Is your feature request related to a problem? Please describe.**
when calling `/traces` actuator endpoint without passing `Accept` in `Request Header`
`The format [application/x-protobuf ] is not supported` is returned in response

I think it would be good by default [CONTENT_TYPE_OPENZIPKIN_JSON_V2](https://github.com/spring-cloud/spring-cloud-sleuth/blob/3.1.x/spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/actuate/TextOutputFormat.java#L34) should be used and a valid `JSON` response is returned

**Describe the solution you'd like**
This is happening because as per  [Producible.java](https://github.com/spring-projects/spring-boot/blob/8540bc0d121e1af98fb1d42749fe53f51e548df5/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/Producible.java#L32-L42) documentation in `spring-boot-actuator` the enum values are matched using `Enum#ordinal()` in highest to the lowest fashion.

and as ordinal value of [CONTENT_TYPE_OTLP_PROTOBUF](https://github.com/spring-cloud/spring-cloud-sleuth/blob/3.1.x/spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/actuate/TextOutputFormat.java#L39) is highest it is matched first.

I tested in local if we change the order of `enum` declaration in `TextOutputFormat` and call endpoint withought passing `Accept` by default `CONTENT_TYPE_OPENZIPKIN_JSON_V2` will be picked and result will be returned as `JSON`

If this is a valid enhancement, I can submit a PR for the required changes.

marcingrzejszczak:
Definitely, great catch. Be my guest!

marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-sleuth/commit/f0af136af821deca1dbf960185b6a0cce1a9912e

PR #1997, opened 2021-07-15
Fix traces actuator should return json by default

beatfreaker:
Fixes gh-1996

Change order of TextOutputFormat enum and give `CONTENT_TYPE_OPENZIPKIN_JSON_V2`
highest ordinal value, so that it will be matched first by spring-boot-actuator
and in case of default Accept header JSON data will be returned.

### DIFF
--- a/spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/actuate/TextOutputFormat.java
+++ b/spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/actuate/TextOutputFormat.java
@@ -29,14 +29,14 @@ import org.springframework.util.MimeType;
 public enum TextOutputFormat implements Producible<TextOutputFormat> {
 
 	/**
-	 * OpenZipkin text.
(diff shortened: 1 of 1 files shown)


beatfreaker:
@marcingrzejszczak @jonatan-ivanov 
build is failing with below error, and I saw that these methods have been overridden in main branch [602749a](https://github.com/spring-cloud/spring-cloud-sleuth/commit/602749a8d148c281e9f519a1ef362744eeff1f44)
```log
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.8.1:compile (default-compile) on project spring-cloud-sleuth-instrumentation: Compilation failure
Error:  /home/runner/work/spring-cloud-sleuth/spring-cloud-sleuth/spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/cassandra/TraceReactiveSession.java:[45,8] org.springframework.cloud.sleuth.instrument.cassandra.TraceReactiveSession is not abstract and does not override abstract method getKeyspace() in org.springframework.data.cassandra.ReactiveSession
```

PR #2010, opened 2021-08-19
Fix typo in document example

giger85:
Using `Backend.class` getLogger() argument in `Example.class` maybe a typo.

e.g)
```java
public class Foo {
    private static final Logger log = LoggerFactory.getLogger(Foo.class);
}
```


jonatan-ivanov (approved):

jonatan-ivanov:
Thanks for the fix! :)

PR #2011, opened 2021-08-20
Instrumenting Spring Security

jonatan-ivanov:


spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/security/TracingSecurityContextChangedListener.java
```diff
@@ -0,0 +1,99 @@
...
+	private void handleNoContextEvent() {
+		attachEvent("security-context noop");
+	}
+
+	private String toString(Authentication authentication) {
+		return authentication != null ? authentication.getClass().getSimpleName() + authentication.getAuthorities()
+				: "null";
+	}
+
+	private void attachEvent(String name) {
```
marcingrzejszczak, line 91: Will you convert this to a `DocumentedSpan` and the `AssertingSpan` option?

spring-cloud-sleuth-samples/spring-cloud-sleuth-sample-security/pom.xml
```diff
@@ -0,0 +1,90 @@
...
+	<modelVersion>4.0.0</modelVersion>
+
+	<artifactId>spring-cloud-sleuth-sample-security</artifactId>
+	<packaging>jar</packaging>
+	<name>Spring Cloud Sleuth Sample Security</name>
+	<description>Spring Cloud Sleuth Sample Security</description>
+
+	<parent>
+		<groupId>org.springframework.cloud</groupId>
+		<artifactId>spring-cloud-sleuth-samples</artifactId>
```
marcingrzejszczak, line 30: Isn't it better to add this to `tests-common` so that we can run the same tests against the same sample for OTel?

spring-cloud-sleuth-samples/spring-cloud-sleuth-sample-security/src/main/java/sample/SecurityConfiguration.java
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Spring Security config
+ *
+ * @author Jonatan Ivanov
+ * @since 3.1.0
+ */
+@Configuration
+@EnableWebSecurity
+public class SecurityConfiguration extends WebSecurityConfigurerAdapter {
```
jzheaux, line 42: I'd recommend not extending `WebSecurityConfigurerAdapter`.

Instead consider exposing a `SecurityFilterChain` bean like so:

```java
@Bean 
SecurityFilterChain filterChain(HttpSecurity http) {
    http
        .authorizeRequests((requests) -> requests
            .antMatchers("/favicon.ico").permitAll()
            .anyRequest().authenticated()
        )
        // ...

    return http.build();
}
```

spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/security/TracingSecurityContextChangedListener.java
```diff
@@ -0,0 +1,99 @@
...
+	public TracingSecurityContextChangedListener(Tracer tracer) {
+		this.tracer = tracer;
+	}
+
+	@Override
+	public void securityContextChanged(SecurityContextChangedEvent securityContextChangedEvent) {
+		SecurityContext previousContext = securityContextChangedEvent.getPreviousContext();
+		SecurityContext currentContext = securityContextChangedEvent.getCurrentContext();
+
+		if (previousContext != null) {
```
jzheaux, line 50: I'd recommend reducing the logic to looking at the `Authentication`. For example:

```java
Authentication previousAuthentication = previousContext == null ? null : previousContext.getAuthentication();
// ...

if (previousAuthentication != null) {
// ...
```

The reason is that Spring Security uses the null object pattern to reduce null checks. Going from no security context to an empty one is purely infrastructural and carries no information for the end user.

spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/security/TracingSecurityContextChangedListener.java
```diff
@@ -0,0 +1,99 @@
...
+import org.springframework.security.core.context.SecurityContextChangedEvent;
+import org.springframework.security.core.context.SecurityContextChangedListener;
+
+/**
+ * {@link SecurityContextChangedListener} that adds tracing support for Spring Security.
+ *
+ * @author Jonatan Ivanov
+ * @since 3.1.0
+ */
+public class TracingSecurityContextChangedListener implements SecurityContextChangedListener {
```
jzheaux, line 35: While implementing `SecurityContextChangedListener` is fine, I wonder if you'd be better off leaving this off and then using `@EventListener` on your `securityContextChanged` method. 

The reason I'm thinking this is it would allow someone to configure the `SecurityContextHolder` without having to first know about all the listeners, e.g. `SecurityContextHolder.addListener(applicationContext::publishEvent)`.

jzheaux (changes requested):
Nice, @jonatan-ivanov! I'm happy to see this happening.

spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/security/TracingSecurityContextChangedListener.java
jonatan-ivanov: I'm thinking about this from AutoConfiguration perspective. I think we should have something like this (in Boot?):

```java
private void registerListeners(List<SecurityContextChangedListener> securityContextChangedListeners, ApplicationContext context) {
	securityContextChangedListeners.forEach(SecurityContextHolder::addListener);
	SecurityContextHolder.addListener(context::publishEvent);
}
```

So users don't need to register their listeners manually they just create a bean for their `SecurityContextChangedListener` or `@EventListener` and it will be auto-configured.

With this, please let me know if you prefer `@EventListener`, I'm ok with it, I slightly leaning towards having a `SecurityContextChangedListener` in Sleuth because:
- It's discoverable (users can search for implementations of `SecurityContextChangedListener`)
- It's a little bit nicer from testing perspective

spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/security/TracingSecurityContextChangedListener.java
jonatan-ivanov: Yupp 😃

spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/security/TracingSecurityContextChangedListener.java
jzheaux: Yes, I also like the idea of Boot providing this configuration.

Good points about discoverability and testing. I think it's good to leave things as they are.

jzheaux (approved):
Nice, @jonatan-ivanov!

PR #2021, opened 2021-09-06
fix #2004 Span is not terminated properly in Spring Cloud Gateway

### LINKED ISSUE #2004: Span is not terminated properly in Spring Cloud Gateway
This issue is also submitted in Spring Cloud Gateway. [https://github.com/spring-cloud/spring-cloud-gateway/issues/2320](https://github.com/spring-cloud/spring-cloud-gateway/issues/2320)

There are two example for this phenomenon.

First is to enable `AdaptCachedBodyGlobalFilter` for a route, then the span will be terminated after the `AdaptCachedBodyGlobalFilter`  in request chain. For example:

```
//other filter...
LogFilter1
AdaptCachedBodyGlobalFilter (cache is enabled for this route)
LogFilter2
```

You would find the log context possess spanId and traceId in LogFilter1 **but does not possess in LogFilter2**.

The other example is while use RetryGatewayFilterFactory. You can enable RetryGatewayFilterFactory trace log and you would find the log context possess spanId and traceId only before the first request sent out, and **there is no spanId or traceId in the log context for the possible subsequent retry**.

Maybe we should subscribe some specific filter in Spring Cloud Gateway to solve it. The only subscribe in `TraceWebFilter` is not enough.

HashZhang:
As stated in title, for request body, it should also be traced


marcingrzejszczak:
Can you please create a test for that?

marcingrzejszczak:
The build is broken due to 

```
 Error:  /home/runner/work/spring-cloud-sleuth/spring-cloud-sleuth/spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebFilter.java:47:1: Wrong order for 'org.springframework.core.Ordered' import. [ImportOrder]
```

Can you put the imports in order and write the missing tests please?

spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/web/TraceWebFilter.java
```diff
@@ -96,12 +100,18 @@ public TraceWebFilter(Tracer tracer, HttpServerHandler handler, CurrentTraceCont
 	@Override
 	public Mono<Void> filter(ServerWebExchange exchange, WebFilterChain chain) {
 		String uri = exchange.getRequest().getPath().pathWithinApplication().value();
-		Mono<Void> source = chain.filter(exchange);
+		ServerWebExchange finalExchange = exchange.mutate().request(new ServerHttpRequestDecorator(exchange.getRequest()) {
```
marcingrzejszczak, line 103: Can you explain what exactly you're trying to achieve here?

marcingrzejszczak:
Closing due to inactivity

PR #2053, opened 2021-11-03
Refactored TraceListenerStrategy to better track current scope in case multiple connections are used

gavlyukovskiy:
Refactored `remoteServiceName` resolution logic to be easier to follow
Fixed `TraceQueryExecutionListener#getConnection` that leaks connection by acquiring (and not closing) additional connection from a `DataSource`

The main idea is to track current connection and only put it and statements/resultSets it creates into the scope. For most applications there will be only one connection and all resources are closed properly, but if it's not the case we'll not mess up the scope. 
The only change that I'm not sure about is that I changed `SpanAndScope` to have optional `Scope`, if that's not okay we can always track spans/scopes inside `ConnectionInfo`/`StatementInfo`/`ResultSetInfo`.

I have also removed getters from inner classes, as it seems more in line with other inner classes in Sleuth.


marcingrzejszczak (approved):

gavlyukovskiy:
Looks like that failed `KafkaSenderTest.should_inject_native_headers` is flaky, it doesn't fail when I run tests locally. Is there a way to rerun he tests?

marcingrzejszczak:
I've rerun the build

PR #2055, opened 2021-11-08
Fix HookRegisteringBeanFactoryPostProcessor to work with native.

OlgaMaciaszek:


spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/instrument/reactor/TraceReactorAutoConfiguration.java
```diff
@@ -354,6 +348,14 @@ public Object next() {
 		};
 	}
 
+	@Override
+	public void setApplicationContext(ApplicationContext applicationContext) throws BeansException {
+		if (log.isTraceEnabled()) {
+			log.trace("Setting context for HookRegisteringBeanFactoryPostProcessor: [" + applicationContext + "]");
+		}
+		springContext = (ConfigurableApplicationContext) applicationContext;
```
jonatan-ivanov, line 356: Can we get into trouble because of this cast?

jonatan-ivanov (approved):

jonatan-ivanov:
Should we do this against 3.0 as well, or should this be 3.1 only?

spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/instrument/reactor/TraceReactorAutoConfiguration.java
OlgaMaciaszek: I don't think so, at least in contexts created by our libs (https://github.com/spring-projects/spring-framework/blob/5acc3228f772630b44113286b9c0f2ab3768ab4a/spring-context/src/main/java/org/springframework/context/ConfigurableApplicationContext.java#L31), but I could modify it to do nothing if can't cast.

Issue #2059, opened 2021-11-19
DefaultClientResources was not shut down properly, shutdown() was not called before it's garbage-collected. 


shyr:
Hello. 
I am using spring-cloud-sleuth:3.0.3 for distributed tracing and I defined my own ClienResources Bean like below.
```kotlin
@Configuration
class LockConfig {
    ....
    @Bean(destroyMethod = "shutdown")
    fun resources(meterRegistry: MeterRegistry): ClientResources {
        return createClientResources(8, 8, redisLatencyMonitor(meterRegistry))
    }

    private fun createClientResources(
        ioThreadPoolSize: Int,
        computationThreadPoolSize: Int,
        redisLatencyMonitor: RedisLatencyMonitor
    ): ClientResources {
        return DefaultClientResources.builder()
            ...
            .build()
    }
}
```

Every now and then, my spring boot application prints warning logs during termination as below.
```
io.lettuce.core.resource.DefaultClientResources was not shut down properly, shutdown() was not called before it's garbage-collected. Call shutdown() or shutdown(long,long,TimeUnit)
```

This warning occurs because the DefaultClientResources object is terminated by the garbage collector before the DefaultClientResources.shutdown() method is called. Please see [this code](https://github.com/lettuce-io/lettuce-core/blob/cda3be6b9477da790365ad098c6e39c8687f5002/src/main/java/io/lettuce/core/resource/DefaultClientResources.java#L680-L685) in lettuce-core.

And the [description of DefaultClientResources](https://github.com/lettuce-io/lettuce-core/blob/cda3be6b9477da790365ad098c6e39c8687f5002/src/main/java/io/lettuce/core/resource/DefaultClientResources.java#L54) also says that I need to call shutdown() when I don't need it.

> The {@link DefaultClientResources} instance is stateful, you have to shutdown the instance if you're no longer using it.

AFAIK, if the DefaultClientResources object is registered as a bean, the shutdown() method would be called during the spring bean destruction process.

[...]

PR #2060, opened 2021-11-20
Shutdown unused ClientResources

shyr:
Implementations of ClientResources are stateful and must be shutdown after they are no longer in use. 

This PR will fix https://github.com/spring-cloud/spring-cloud-sleuth/issues/2059


marcingrzejszczak:
Thanks @shyr for your contribution!

marcingrzejszczak:
I had a chat with @mp911de and we've reverted the change. We need to solve this problem in a different way.

Issue #2083, opened 2021-12-20
spring-cloud-sleuth-instrumentation sql issue for spans


mschielmann:
**Describe the bug**
Version: 3.1.0
The `TraceListenerStrategy` in `spring-cloud-sleuth-instrumentation` throws a `StringOutOfBoundsException` when the SQL consists of only one word, like: `BEGIN`. Such word, afaik, constitutes a valid SQL statement, and as such should be handled without throwing an exception.

The exception is thrown here:
https://github.com/spring-cloud/spring-cloud-sleuth/blob/main/spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/jdbc/TraceListenerStrategy.java#L432
(in `spanName` method), as `indexOf(' ')` is -1.

I'd suggest handling this with returning the whole SQL as span name in such case. Another option would be to ignore such SQL.

**Sample**
N/A

mschielmann:
@marcingrzejszczak answering your question, yes, that would be the case. I don't think any validation is possible though as to if that word can create a valid SQL statement by itself (for example 'BEGINN').

If you agree that this can be fixed by using the whole word - then I could create a MR for that.

marcingrzejszczak:
Be my guest!

mschielmann:
@marcingrzejszczak, waiting for your review.
https://github.com/spring-cloud/spring-cloud-sleuth/pull/2086

gavlyukovskiy:
In my starter I didn't try to parse query to give span a name, so all of them were `query`, there are certainly a lot of edge cases when trying to parse user input, couple that comes to mind
- Not terminated by space: `select*from ...`. Can be solved by `query.split("[^a-zA-Z]", 2)[0]`
- Has a comment in the beginning: `/*some comment*/select ...`
- Not an edge case, but in certain queries first (non-comment) word does not identify the query: `with ... [select|insert|...]`

Not sure what would be the best way to handle those, what do you think about checking that first word is `select`/`insert`/`update`/`delete` and using it as a span name, but in all other cases use something generic (like `query`) as a span name.

mschielmann:
@gavlyukovskiy sounds reasonable to me, I'm happy to make that change to PR if you think that would be a best way to go.

marcingrzejszczak:
I like @gavlyukovskiy comment. Let's do it like he suggests.

gavlyukovskiy:
[...]

PR #2086, opened 2021-12-21
fixing issue #2083

mschielmann:
No tests, as no tests exists so far for this class and it does look as a separate task to add them.

### DIFF
--- a/spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/jdbc/TraceListenerStrategy.java
+++ b/spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/jdbc/TraceListenerStrategy.java
@@ -429,7 +429,11 @@ class TraceListenerStrategy<CON, STMT, RS> {
 	}
 
 	private String spanName(String sql) {
-		return sql.substring(0, sql.indexOf(' ')).toLowerCase(Locale.ROOT);
+		String spanName = sql.toLowerCase(Locale.ROOT);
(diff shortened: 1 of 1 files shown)


spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/jdbc/TraceListenerStrategy.java
```diff
@@ -429,7 +434,15 @@ void afterConnectionClose(CON connectionKey, @Nullable Throwable t) {
...
-		return sql.substring(0, sql.indexOf(' ')).toLowerCase(Locale.ROOT);
+		String lowercaseSql = sql.toLowerCase(Locale.ROOT);
+		String spanName = DEFAULT_SPAN_NAME;
+		for (String spanNameCandidate : KNOWN_SPAN_NAMES) {
+			if (lowercaseSql.startsWith(spanNameCandidate)) {
+				spanName = spanNameCandidate;
+				break;
+			}
+		}
+		return spanName;
```
gavlyukovskiy, line 445: Maybe a bit premature optimization, but what do you think about
```
String prefix = sql.substing(0, Math.min(6, sql.length()).toLowerCase(Locale.ROOT);
return KNOWN_SPAN_NAMES.contains(prefix) ? prefix : DEFAULT_SPAN_NAME
```
so that we generate less garbage by not lower-casing whole query. `KNOWN_SPAN_NAMES` can remain a list, that's likely faster than hashing anyway.

gavlyukovskiy (approved):

PR #2091, opened 2021-12-28
#2083 improvements

mschielmann:
My previous PR #2086 has been merged before I could respond to the comments. I've created another PR with some improvements suggested in the previous PR.

@gavlyukovskiy I've used regex instead of doing substring. I've figured substring(6) would be a result of lucky coincidence that all known query types (select, update, insert, delete) have 6 chars. If someone in the future wanted to add another type - for example BEGIN or ALTER - then they would have a refactoring problem in their hands. Now all that would need to be done is to add the new type to the list of 'known' types.

I've also extracted the span 'generation' into separate class that can be now tested, so that it's also somewhat documented now.

Please let me know if that works better.


marcingrzejszczak:
There are some checkstyle issues https://github.com/spring-cloud/spring-cloud-sleuth/runs/4690019673?check_suite_focus=true

Issue #2064, opened 2021-11-24
TraceId spreading between containers thread pools 


CyberpunkPerson:
Hi Spring Team! 
We found a bug with traceId spreading in `TracingChannelInterceptor ` the issue starts in [here](https://github.com/spring-projects/spring-integration/issues/3649), the essence of the bug [here](https://github.com/spring-projects/spring-integration/issues/3649#issuecomment-977036096).

Best regards,
Andrew

maecval:
@marcingrzejszczak 
Hi Marcin,
Thanks for the fix.
Can you tell us what we can do to get this fix asap?
The bug causes major issues in our productive environment.
Is there a hotfix release available?
Cheers!
Valentin

marcingrzejszczak:
You can either disable that integration via the `spring.sleuth.messaging.enabled=false` property, or wait until 17.02 when we do the release or use tools such as jitpack.io to point to a concrete commit in this repository and use a snapshot version.

maecval:
Thanks for the quick response.
We have already disabled it for transactions (`spring.sleuth.tx.enabled=false`) and need to check what it means if we disable it for messaging.

marcingrzejszczak:
This issue is about `TracingChannelInterceptor` which is a messaging component. If you disable messaging support in Sleuth then you won't have issues with `TracingChannelInterceptor`

maecval:
As far as we understand we will loose the population of traceId / spanId if we do so we so.
BTW: it is certainly closely related to [issues 2082](https://github.com/spring-cloud/spring-cloud-sleuth/issues/2082) which is also one of the issues our apps suffer.

marcingrzejszczak:
Yes, you will lose that functionality

gigermocas:
Hi, can we expect this fix on a 3.0.x release?

marcingrzejszczak:
3.0.x is outside of OSS support. Please check this website for more information https://spring.io/projects/spring-cloud-sleuth#support

pankajchugh7:
as per the doc 3.0.x, end of support date is 2022-05-19. Can you pls check once.
We need this fix in 3.0.x.

can you pls highlight any compatibility breaks between 3.1.x and 3.0.x.

PR #2108, opened 2022-02-02
TraceId spreading between containers thread pools fix

CyberpunkPerson:
The stack of calls was out of ThreadLocal, resulting in distorted reading by other threads under load

Fixes gh-2064

### DIFF
--- a/spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TracingChannelInterceptor.java
+++ b/spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/messaging/TracingChannelInterceptor.java
@@ -16,9 +16,9 @@
 
 package org.springframework.cloud.sleuth.instrument.messaging;
 
+import java.util.ArrayDeque;
 import java.util.NoSuchElementException;
 import java.util.Set;
-import java.util.concurrent.LinkedBlockingDeque;
 import java.util.function.Function;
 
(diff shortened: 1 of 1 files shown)


marcingrzejszczak:
Hi! Thanks for the PR. Can you add some tests to this please? We have similar code in the `ThreadLocalSpan` class, do you think that's buggy too?

CyberpunkPerson:
Hi @marcingrzejszczak !
1. I checked `org.springframework.cloud.sleuth.ThreadLocalSpan` and yes, seems that it's also has this issue. 
2. I'll add some tests a little bit later. I used [Tracing](https://github.com/CyberpunkPerson/Tracing) sandbox to reproduce.

marcingrzejszczak:
Hey @CyberpunkPerson , I took the liberty of fixing this and some other things in this commit https://github.com/spring-cloud/spring-cloud-sleuth/commit/66c63a53ff902f5d9c1d65595bec778389e044bd

CyberpunkPerson:
**Note (for those who may faced with it):** 
If you want to have `traceId` per message you need to override default behavior, 
switch `singleRootSpanOnReceiveBatch` by

```
 @Bean
    KafkaTracing kafkaTracing(MessagingTracing messagingTracing, SleuthMessagingProperties properties) {
        return KafkaTracing.newBuilder(messagingTracing)
                .remoteServiceName(properties.getKafka().getRemoteServiceName())
                .singleRootSpanOnReceiveBatch(false)
                .build();
    }
```

otherwise `TracingConsumer` will reuse `traceId` for topic.

PR #2157, opened 2022-05-16
Support configuring queuedMaxSpans in AsyncReporter

karkum:
This PR provides the ability to configure the `queuedMaxSpans` property of the AsyncReporter with a config: `spring.zipkin.queuedMaxSpans`, the default is unchanged.

We would like to be able to tweak this property to support applications with a high-volume of spans, which are dropped once the queue is full. 

I understand that this bean could be overridden to achieve the same result, but this seems like a more ideal way to configure this property.

There is also `queuedMaxBytes` which provides a memory-based bound on the queue ([defaults to 1%](https://www.javadoc.io/doc/io.zipkin.reporter2/zipkin-reporter/2.6.0/zipkin2/reporter/AsyncReporter.Builder.html#queuedMaxBytes-int-)), I can also override that property in this PR (although that is not necessary for my current use-case)


karkum:
@asibross @narayaruna

marcingrzejszczak:
Thanks @karkum !

karkum:
@marcingrzejszczak thanks for the review. Do you know when this change will be released?

marcingrzejszczak:
https://github.com/spring-cloud/spring-cloud-release/milestone/121 today according to the calendar

PR #2177, opened 2022-06-09
Fixes #2156 (Bug with multiple W3C baggage headers)

### LINKED ISSUE #2156: W3C baggage extraction does not work with multiple baggage headers
**Describe the bug**

Spring Cloud Version: 2021.0.2
Spring Boot Version: 2.6.7

According to the [W3C baggage spec](https://www.w3.org/TR/baggage/#header-name), multiple baggage headers are allowed. 

Sleuth seems to pick the first `baggage` header and disregards the others. 

**Sample**

I have a sample project here: https://github.com/jamesmoessis/demo. It should be easy to recreate the issue with the `testing.http` file. 

How to reproduce: 
```
spring.sleuth.propagation.type=w3c
spring.sleuth.baggage.remote-fields=baggage-key1,baggage-key-2
```
and send a request with multiple baggage headers:
```
###
GET http://localhost:8080/path
traceparent:00-00000000000000003e7eb8c882e31157-6585b8bcb4cd1caf-01
baggage:baggage-key-1=abc
baggage:baggage-key-2=def 
```
Only `baggage-key-1=abc` is present in the baggage. 

Additionally, there are cases where sleuth doesn't respect the `remote-fields` config. For example, with this request: 
```
###
GET http://localhost:8080/howdy
traceparent:00-00000000000000003e7eb8c882e31157-6585b8bcb4cd1caf-01
baggage:baggage-key-3=ghi
```
`baggage-key-3=ghi` shows up in the baggage even though it's not in `remote-fields`.

jamesmoessis:
This fixes #2156. When retrieving the baggage header, noted as a "combinable" header as per the [HTTP 1.1 specification](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.2) and the [W3C baggage header specification](https://www.w3.org/TR/baggage/#baggage-http-header-format), multiple headers of the same key are allowed in certain circumstances.

When retrieving headers, sleuth sleuth currently calls `HttpServletRequest.getHeader(String name)` which returns the first header with that name. As a result, data in the non-first baggage header is lost. To solve this, I've changed it so (in the baggage case) `HttpServletRequestWrapper` calls `getHeaders(String name)` and joins the Enumeration via a "," into a single header, as per the HTTP spec: 

> A recipient MAY combine multiple header fields with the same field name into one "field-name: field-value" pair, without changing the semantics of the message, by appending each subsequent field value to the combined field value in order, separated by a comma.

### DIFF
--- a/spring-cloud-sleuth-brave/src/test/java/org/springframework/cloud/sleuth/brave/bridge/W3CBaggagePropagatorTest.java
+++ b/spring-cloud-sleuth-brave/src/test/java/org/springframework/cloud/sleuth/brave/bridge/W3CBaggagePropagatorTest.java
@@ -29,6 +29,9 @@ import brave.propagation.TraceContextOrSamplingFlags;
 import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Test;
 
+import org.springframework.cloud.sleuth.instrument.web.servlet.HttpServletRequestWrapper;
+import org.springframework.mock.web.MockHttpServletRequest;
+
 import static java.util.Collections.singletonMap;
 import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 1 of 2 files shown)


marcingrzejszczak:
Great work @jamesmoessis ! Thank you!

PR #2183, opened 2022-06-28
Escape special characters in config Asciidoc

RobertRad:
Disabling parsing of special characters in Asciidoc to show valid configuration examples


marcingrzejszczak:
We are computing these properties somewhere around here https://github.com/spring-cloud/spring-cloud-build/blob/3.0.x/docs/src/main/java/org/springframework/cloud/internal/Main.java#L95 so we would need to apply the change there

PR #2210, opened 2022-09-25
all: fix some typos

cuishuang:


marcingrzejszczak:
The `_spans.adoc` actually gets built from sources, so you could search for that text in the code so that we can fix it there.

PR #2264, opened 2023-02-17
Thermoweb bugfix/2246 handle invocation target exception

### LINKED ISSUE #2246: TraceSessionRepositoryAspect#callMethodOnWrappedObject does not handle InvocationTargetException
**Describe the bug**
I am using spring-cloud-sleuth-instrumentation-3.1.5.jar in combination with spring-aop-5.3.24.jar

The method `callMethodOnWrappedObject(ProceedingJoinPoint pjp, T target)` invokes a method to access a sessionRepository.
But it doesn't handle the case properly, when the invoked method throws a `RuntimeException` or another declared Exception.

The result of this bug is, that a `org.springframework.web.reactive.handler.WebFluxResponseStatusExceptionHandler` and perhaps other places also see a `java.lang.reflect.UndeclaredThrowableException` instead of the actually thrown exception.
The `UndeclaredThrowableException` is created in https://github.com/spring-projects/spring-framework/blob/46875c91ced9387a598c3ca7eba3ed23fdb4fc63/spring-aop/src/main/java/org/springframework/aop/framework/CglibAopProxy.java#L769
because the called sessionRepository method doesn't declare to throw an `InvocationTargetException` (which is correct).

Current implementation: https://github.com/spring-cloud/spring-cloud-sleuth/blob/8302481d89fc6744876ba560f95aab604c7890f6/spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/session/TraceSessionRepositoryAspect.java#L82

Suggested implementation:
```java
try {
  return method.invoke(target, pjp.getArgs());
} catch(InvocationTargetException ex) {
  throw ex.getCause();
}
```
similar to https://github.com/spring-projects/spring-framework/blob/46875c91ced9387a598c3ca7eba3ed23fdb4fc63/spring-aop/src/main/java/org/springframework/aop/aspectj/AbstractAspectJAdvice.java#L640
or you use org.springframework.util.ReflectionUtils.invokeMethod.

I found a similar issue [#1092](https://github.com/spring-cloud/spring-cloud-sleuth/issues/1092). Maybe there are more like this.

marcingrzejszczak:
fixes gh-2246

### DIFF
--- a/spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/session/TraceSessionRepositoryAspect.java
+++ b/spring-cloud-sleuth-instrumentation/src/main/java/org/springframework/cloud/sleuth/instrument/session/TraceSessionRepositoryAspect.java
@@ -16,6 +16,7 @@
 
 package org.springframework.cloud.sleuth.instrument.session;
 
+import java.lang.reflect.InvocationTargetException;
 import java.lang.reflect.Method;
 import java.lang.reflect.Modifier;
 import java.util.Arrays;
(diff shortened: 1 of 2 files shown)


marcingrzejszczak:
Congrats @thermoweb  on your first contribution! :tada:

thermoweb:
> Congrats @thermoweb on your first contribution! tada

Thanks, you did all the work at the end ^^'

marcingrzejszczak:
We did it together @thermoweb ! :)

ghost:
Thank you both. Nice and clean code.

PR #2280, opened 2023-04-13
Properly clears the scopes

marcingrzejszczak:
without this fix when calling `maybeScope(null)` or `withSpan(null)` or `newScope(null)` we ended up with creating a null scope that would allocate additional resources and not clear things.

with this fix we're clearing all the scopes and removing thread locals

### DIFF
--- a/spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/bridge/BraveCurrentTraceContext.java
+++ b/spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/bridge/BraveCurrentTraceContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2021 the original author or authors.
+ * Copyright 2013-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,16 +20,14 @@ import java.util.concurrent.Callable;
 import java.util.concurrent.Executor;
 import java.util.concurrent.ExecutorService;
 
+import brave.propagation.ThreadLocalCurrentTraceContext;
+
 import org.springframework.cloud.sleuth.CurrentTraceContext;
 import org.springframework.cloud.sleuth.TraceContext;
 
-/**
- * Brave implementation of a {@link CurrentTraceContext}.
- *
- * @author Marcin Grzejszczak
- * @since 3.0.0
- */
 public class BraveCurrentTraceContext implements CurrentTraceContext {
+
+	final ThreadLocal<Scope> scopes = new ThreadLocal<>();
 
 	final brave.propagation.CurrentTraceContext delegate;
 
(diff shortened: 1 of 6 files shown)


spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/bridge/BraveCurrentTraceContext.java
```diff
@@ -20,58 +20,70 @@
...
 
+	final ThreadLocal<Scope> scopes = new ThreadLocal<>();
+
 	final brave.propagation.CurrentTraceContext delegate;
 
 	public BraveCurrentTraceContext(brave.propagation.CurrentTraceContext delegate) {
 		this.delegate = delegate;
 	}
 
-	@Override
```
jonatan-ivanov, line 40: `@Override` isn't needed anymore?

spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/bridge/BraveCurrentTraceContext.java
```diff
@@ -20,58 +20,70 @@
...
 
+import brave.propagation.ThreadLocalCurrentTraceContext;
+
 import org.springframework.cloud.sleuth.CurrentTraceContext;
 import org.springframework.cloud.sleuth.TraceContext;
 
-/**
- * Brave implementation of a {@link CurrentTraceContext}.
- *
- * @author Marcin Grzejszczak
```
jonatan-ivanov, line 29: Should not we have this, since it is a public class?

spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/bridge/BraveCurrentTraceContext.java
```diff
@@ -20,58 +20,70 @@
...
+	public CurrentTraceContext.Scope newScope(TraceContext context) {
 		if (context == null) {
-			return null;
+			clearScopes();
+			return Scope.NOOP;
 		}
-		return new BraveTraceContext(context);
+		CurrentTraceContext.Scope previous = this.scopes.get();
+		CurrentTraceContext.Scope current = new BraveScope(this.delegate.newScope(BraveTraceContext.toBrave(context)));
+		return new RevertingScope(this, previous, current);
```
jonatan-ivanov: If `RevertingScope` deals with the `scopes` `ThreadLocal`, would not it be better to let it get the previous one on it's own instead of passing it? That way all logic that touched `scopes` would be in the same place.

marcingrzejszczak: Can you elaborate? I don't think i follow

marcingrzejszczak: Actually now i wonder if we shouldn't do it on the level of sleuth api rather than Braves. The same way we do it with micrometer tracing...

jonatan-ivanov: I meant moving this:
```java
CurrentTraceContext.Scope previous = this.scopes.get();
```
Into the ctor of `RevertingScope` but maybe we can also move current:
```java
CurrentTraceContext.Scope current = new BraveScope(this.delegate.newScope(BraveTraceContext.toBrave(context)));
```

marcingrzejszczak: We can't move all of that into reverting scpoe cause there's one reverting scope but two delegates (maybeScope, newScope). I can improve this constructor however. 

We also can't move the code to sleuth's api cause there are no implementations there - just interfaces. It has to stay here & in OTel.

PR #2335, opened 2024-01-07
deps: bumps to Brave 5.18.1 and moves off internal type

codefromthecrypt:
This updates to Brave 5.18.0 and dodges an internal type that will not be in Brave 6.0. See https://github.com/openzipkin/brave/pull/1396 about Propagation and Brave 6.0

### DIFF
--- a/spring-cloud-sleuth-autoconfigure/pom.xml
+++ b/spring-cloud-sleuth-autoconfigure/pom.xml
@@ -414,6 +414,7 @@
 			<artifactId>zipkin</artifactId>
 			<optional>true</optional>
 		</dependency>
+		<!-- Needed for ZipkinRestTemplateSenderConfiguration: BytesMessageEncoder  -->
 		<dependency>
 			<groupId>io.zipkin.reporter2</groupId>
 			<artifactId>zipkin-reporter</artifactId>
--- a/spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/bridge/W3CPropagation.java
+++ b/spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/bridge/W3CPropagation.java
@@ -31,7 +31,6 @@ import brave.baggage.BaggageField;
 import brave.baggage.BaggagePropagation;
 import brave.baggage.BaggagePropagationConfig;
 import brave.internal.baggage.BaggageFields;
-import brave.internal.propagation.StringPropagationAdapter;
 import brave.propagation.Propagation;
 import brave.propagation.TraceContext;
 import brave.propagation.TraceContextOrSamplingFlags;
@@ -129,8 +128,8 @@ class W3CPropagation extends Propagation.Factory implements Propagation<String> 
 	}
 
 	@Override
-	public <K> Propagation<K> create(KeyFactory<K> keyFactory) {
-		return StringPropagationAdapter.create(this, keyFactory);
+	public Propagation<String> get() {
+		return this;
 	}
 
 	@Override
@@ -321,7 +320,7 @@ class W3CBaggagePropagator {
 	private BaggagePropagation.FactoryBuilder factory() {
 		return BaggagePropagation.newFactoryBuilder(new Propagation.Factory() {
 			@Override
-			public <K> Propagation<K> create(Propagation.KeyFactory<K> keyFactory) {
+			public Propagation<String> get() {
 				return null;
 			}
 		});
--- a/pom.xml
+++ b/pom.xml
@@ -60,8 +60,6 @@
 		<spring-cloud-openfeign.version>3.1.10-SNAPSHOT</spring-cloud-openfeign.version>
 		<spring-cloud-task.version>2.4.6</spring-cloud-task.version>
 		<spring-cloud-deployer.version>2.5.1</spring-cloud-deployer.version>
-		<brave.version>5.13.9</brave.version>
-		<opentracing.version>0.32.0</opentracing.version>
 		<!-- Deprecated - reached EOL -->
 		<spring-security-boot-autoconfigure.version>2.3.4.RELEASE</spring-security-boot-autoconfigure.version>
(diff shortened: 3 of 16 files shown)


spring-cloud-sleuth-brave/src/main/java/org/springframework/cloud/sleuth/brave/bridge/W3CPropagation.java
```diff
@@ -128,9 +127,19 @@ class W3CPropagation extends Propagation.Factory implements Propagation<String>
 		this.braveBaggageManager = braveBaggageManager;
 	}
 
+	/**
```
codefromthecrypt: p.s I'm not sure if this was copied because brave's wasn't published to maven central. It is now, so maybe can reduce maintenance here https://central.sonatype.com/artifact/io.zipkin.contrib.brave-propagation-w3c/brave-propagation-tracecontext/versions from https://github.com/openzipkin-contrib/brave-propagation-w3c

spring-cloud-sleuth-autoconfigure/pom.xml
```diff
@@ -414,6 +414,7 @@
 			<artifactId>zipkin</artifactId>
 			<optional>true</optional>
 		</dependency>
+		<!-- Needed for ZipkinRestTemplateSenderConfiguration: BytesMessageEncoder  -->
```
codefromthecrypt, line 417: maybe this code should be rewritten to avoid classes or this dep become not optional?

```
backend         | 2024-01-08 05:03:14.185 [] [/] WARN  [main] o.s.c.s.AbstractApplicationContext - Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'zipkinSender' defined in class path resource [org/springframework/cloud/sleuth/zipkin2/sender/ZipkinRestTemplateSenderConfiguration.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [zipkin2.reporter.Sender]: Factory method 'restTemplateSender' threw exception; nested exception is java.lang.NoSuchMethodError: 'zipkin2.reporter.BytesMessageEncoder zipkin2.reporter.BytesMessageEncoder.forEncoding(zipkin2.codec.Encoding)'
```

codefromthecrypt:
@shakuzen @marcingrzejszczak @ryanjbaxter I got this as far as I can.. there's an unrelated error below, which seems likely due to the runner using too new a JDK

```
org.springframework.cloud.sleuth.instrument.messaging.BraveTraceFunctionAwareWrapperTests.test_tracing_with_function()  Time elapsed: 1 sec  <<< FAILURE!
java.lang.ClassCastException: class java.lang.String cannot be cast to class org.springframework.messaging.Message (java.lang.String is in module java.base of loader 'bootstrap'; org.springframework.messaging.Message is in unnamed module of loader 'app')
```

The summary is that this change allows an upgrade to brave 5.18 or 6.0, but while sleuth supplies and consumes AsyncReporter types, it cannot upgrade to zipkin-reporter 2. This may be fine at least for a while, as Brave 6 doesn't care about zipkin or zipkin-reporter versions anymore.

In a bit, I will update this with Brave 6 (need to release that, first)

jonatan-ivanov:
Hi Adrian, thanks for the PR! :)

Sleuth is already over its OSS support [EOL date](https://spring.io/projects/spring-cloud-sleuth/#support) so Brave 6 might not be a concern.
I think a minor version upgrade is a good question: we are not planning to release a new minor version of Sleuth only patch versions for commercial users.

On the other hand, we should do this in [Micrometer Tracing](https://github.com/micrometer-metrics/tracing) as it is the successor of Sleuth.

codefromthecrypt:
ok so what I'll do is cross-test this PR with Brave 5.18 and 6 and then close it if it won't be merged. Then, in case you need the code later, you know where to look!

codefromthecrypt:
So, the main thing is that if this PR isn't merged and released, users can't upgrade to Brave 5.18 as things that wrap Propagation.Factory don't wrap `.get()`. So, they will be stuck at 5.17.1 regardless of zipkin-reporter or deprecation.

If it is merged, likely this will work with Brave 6 as well, just the messaging tests are broken and until they aren't I'm not sure if there is another issue.

jonatan-ivanov:
I think that might be a compelling argument for merging this even though we are not planning to do another minor release. We can also keep it open to see if there is user interest.

marcingrzejszczak:
I'm looking into this and I'm getting some `NoClassDefFoundError`. I'll look into that

```
itialization - cancelling refresh attempt: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'kafkaStreamsTracing' defined in class path resource [org/springframework/cloud/sleuth/autoconfig/brave/instrument/messaging/BraveKafkaStreamsAutoConfiguration.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [brave.kafka.streams.KafkaStreamsTracing]: Factory method 'kafkaStreamsTracing' threw exception; nested exception is java.lang.NoClassDefFoundError: org/apache/kafka/streams/processor/api/FixedKeyProcessorSupplier
```

codefromthecrypt:
glad to hear you are looking at the build. I think people will be using boot 2 for many years.. if we can get a bit more life extension to sleuth to the community, I'm sure they will appreciate it!

marcingrzejszczak:
Done via https://github.com/spring-cloud/spring-cloud-sleuth/commit/8f39760f5329433ed924d5489123c4d3a7bcd92c, polished via https://github.com/spring-cloud/spring-cloud-sleuth/commit/83155920f15db457f5ee6b71a19fc76b4ba293d8

codefromthecrypt:
this works provably. While people can't upgrade zipkin-reporter to a new version, they can upgrade brave without a snag: https://github.com/openzipkin/brave-example/pull/113

For most people, since they are using sleuth's reporters anyway, there's little problem here.

